PR #7, opened 2025-06-13
Update auth.md

madhurprash:
*Issue #, if available:*

*Description of changes:*


aarora79:
The update to the file is not going to be required because the original line itself is going to be removed or written in a different form.

PR #179, opened 2025-10-13
Feature/import servers

nishadeborahphilips:
*Issue #, if available:*

*Description of changes:*
1. import servers from anthropic
2. set corresponding status (healthy (streamable-http and see server that i can access and retrieve tools from), unhealthy (streamable-http and see server that are inaccessible), unknown (stdio transport type for which the server needs to be deployed locally))
3. deployment instructions aren't being generated in this PR


aarora79:
incorporated in https://github.com/agentic-community/mcp-gateway-registry/pull/181

PR #180, opened 2025-10-14
Draft: $HOME potentially breakdown the local startup in mac

ryohang:
*Issue #, if available:*
On macOS and Linux distributions, $HOME is typically set to /Users/{username} or /home/{username} respectively. When users start the service, they frequently encounter volume mounting errors due to this system variable conflict.
*Description of changes:*
This PR introduces a new $APP_HOME variable in the .env configuration to replace the use of the system $HOME variable. While this is a relatively small code change, it has significant implications for startup scripts and could resolve mounting issues for users across different environments.

Note: This PR is marked as DRAFT as the change, though minor in scope, requires careful review due to its potential impact on the application's startup behavior and user experience. Let me know if the approach is okay for you.


aarora79 (commented):
Thank you for providing this PR. This is something that I was also considering to add as part of another (yet to be created) issue. Please add comments in `.env.example` that APP_HOME could be set to these potential paths (provide examples). We need to avoid using `/opt` because then we run into `sudo` requirements, so different options for the home directory would be useful with default as `/home/ubuntu` and then MAC setup can suggest MAC appropriate defaults.
@nishadeborahphilips , @dheerajoruganty could you please test this PR so that we can merge this in today. Thanks.

ryohang:
> different options for the home directory would be useful with default as `/home/ubuntu`

Thanks for taking into the consideration, i added a few comments as per suggestion.

PR #266, opened 2025-12-02
fix: metrics-service build

kanghengliu:
./build-and-run.sh won't work for local deployment. Metrics service is building in wrong path. This commit fixes it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,8 +60,8 @@ services:
   # Metrics Collection Service
   metrics-service:
     build:
-      context: .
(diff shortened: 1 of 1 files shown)


aarora79:
This change has already been implemented in commit 065df59 (fix: Configure local-only packages to prevent PyPI resolution).

The metrics-service build context was changed from:
```yaml
context: .
dockerfile: metrics-service/Dockerfile
```

to:
```yaml
context: metrics-service
dockerfile: Dockerfile
```

This fix is already in the main branch, so closing this PR as duplicate. Thank you for your contribution!

PR #324, opened 2025-12-30
split registry image build process and use torch cpu wheel

### LINKED ISSUE #319: Registry uses torch with CUDA but doesn't require GPUs
https://github.com/agentic-community/mcp-gateway-registry/blob/main/docker/Dockerfile.registry#L46

This pulls in a lot of CUDA libraries which creates a large image and increases the build/pull times. If we switch to a CPU only wheel, this will reduce the image size / pull / startup times

omrishiv:
*Issue #, if available:*

Fixes #319 

*Description of changes:*

Updates the registry docker to use a multistage build for the front/back end. This reduces the image from 2.6 GB to 1.6 GB and reduces the build time by leveraging cache (if available). It also uses the torch CPU wheel to further reduce the library size. 

### DIFF
--- a/docker/Dockerfile.registry
+++ b/docker/Dockerfile.registry
@@ -1,22 +1,29 @@
-# Registry Dockerfile - includes nginx, SSL, FAISS, models, and registry service
-FROM python:3.12-slim
+# Registry Dockerfile - Multi-stage build with frontend and backend stages
+
+# ===== FRONTEND BUILD STAGE =====
+FROM node:20-slim AS frontend-builder
+
+WORKDIR /app/frontend
+
+# Copy package files and install dependencies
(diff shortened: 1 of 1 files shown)


aarora79:
This PR has been superseded by #333, which includes your original multi-stage build work (commit d8fedb1) plus additional optimizations:

- Fixed selective file copying to avoid bloating the final image
- Extended CPU-only PyTorch optimization to all MCP servers
- Optimized .dockerignore to reduce build context from 1.77GB to <500MB
- Added missing dependencies and configuration files

Your original contribution is preserved with proper attribution in the new PR. Thank you for the excellent foundation!

Closing this PR in favor of #333.

PR #353, opened 2026-01-12
Registry-Auth: Make OAuth2 provider enablement configurable via environment variables

ndrsfel:
Hey, nice project! Just a small improvement to make OAuth2 provider configuration more flexible for different deployment environments.

## Motivation
By making provider enablement configurable via environment variables, operators can:

- Enable/disable specific identity providers per environment (dev, staging, production)
- Support multi-tenant deployments with different authentication requirements
- Simplify CI/CD pipelines by controlling provider configuration through environment-specific variables
- Avoid maintaining separate configuration files for each deployment target

## Summary
- Replace hardcoded `enabled: true/false` values in oauth2_providers.yml with environment variable references (KEYCLOAK_ENABLED, COGNITO_ENABLED, etc.)
- Add corresponding environment variables to .env.example with documentation
- Pass through *_ENABLED environment variables in docker-compose.yml with sensible defaults (false)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -208,7 +208,7 @@ services:
       - METRICS_API_KEY=${METRICS_API_KEY_AUTH_SERVER}
       # Keycloak configuration
       - AUTH_PROVIDER=${AUTH_PROVIDER:-cognito}  # 'cognito' or 'keycloak'
-      - KEYCLOAK_ENABLED=${KEYCLOAK_ENABLED:-false}
+      - KEYCLOAK_ENABLED=${KEYCLOAK_ENABLED:-true}  # Enable Keycloak by default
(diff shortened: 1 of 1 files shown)


aarora79:
## Follow-up Fix Applied

A follow-up fix was needed and has been pushed directly to main in commit f575995.

**Issue**: Environment variables like `KEYCLOAK_ENABLED` return string values (`"true"`/`"false"`) from `os.environ`, but the OAuth2 provider enablement check was comparing against Python booleans.

**Fix**: Added `.lower()` comparison in `substitute_env_vars()` to convert string booleans to actual Python booleans:

```python
result = template.substitute(os.environ)

# Convert string booleans to actual booleans
if result.lower() == "true":
    return True
elif result.lower() == "false":
    return False

return result
```

This ensures the `enabled` field in `oauth2_providers.yml` works correctly when set via environment variables.

PR #364, opened 2026-01-14
Changed line 365 of docker-compose.prebuilt.yml to reference the official mongo8:2 image

viviluccioli:
*Issue #, if available:* n/a

*Description of changes:* 

- *bug*: --prebuilt was failing because docker-compose.prebuilt.yml was referencing an image mcpgateway/mongo:8.2 that can't be recognized by machines deploying locally, instead of the official mongo:8.2 image
- *fix*: Changed line 365 of docker-compose.prebuilt.yml from -mcpgateway}/mongo:${MONGODB_VERSION:-8.2} to mongo:8.2

### DIFF
--- a/docker-compose.prebuilt.yml
+++ b/docker-compose.prebuilt.yml
@@ -362,7 +362,7 @@ services:
   # Vector search is implemented in application code (see search_repository.py)
   # Running without authentication for local development simplicity
   mongodb:
-    image: mongo:${MONGODB_VERSION:-latest}
+    image: ${DOCKERHUB_ORG:-mcpgateway}/mongo:${MONGODB_VERSION:-latest}
     container_name: mcp-mongodb
     command: mongod --replSet rs0 --bind_ip 127.0.0.1,mongodb
     ports:


aarora79 (changes requested):
this line should be similar to how we handle other external images, for example keycloak, please see https://github.com/agentic-community/mcp-gateway-registry/blob/main/docker-compose.prebuilt.yml#L315. So basically the 8.2 should be `latest`.

viviluccioli:
> this line should be similar to how we handle other external images, for example keycloak, please see https://github.com/agentic-community/mcp-gateway-registry/blob/main/docker-compose.prebuilt.yml#L315. So basically the 8.2 should be `latest`.

done, pushed a second commit and changed it to image: mongo:${MONGODB_VERSION:-latest}

viviluccioli:
> should be `image: ${DOCKERHUB_ORG:-mcpgateway}/mongo:${MONGODB_VERSION:-latest}`

sorry, just fixed it

PR #369, opened 2026-01-14
Fix DocumentDB scope repository lookup logic for group mappings

ndrsfel:
## Problem

The DocumentDB implementation of `get_group_mappings()` and `get_ui_scopes()` was using incorrect query logic.

It was querying `find_one({"_id": keycloak_group})`, looking for a document with `_id` equal to the Keycloak group UUID. However, scope documents are keyed by scope names (e.g., `_id: "registry-admins"`), not Keycloak group IDs.

## Solution

Changed both methods to query `collection.find({"group_mappings": keycloak_group})`:

- **`get_group_mappings`**: Finds all scope documents where the `group_mappings` array contains the keycloak group, then returns their `_id` values (scope names)
- **`get_ui_scopes`**: Same query approach, then merges `ui_permissions` from all matching scopes

This aligns the DocumentDB implementation with the file-based repository behavior.

## Testing

Tested locally against DocumentDB/MongoDB with scope documents structured as:
```yaml
_id: "scope-name"
group_mappings: ["keycloak-group-uuid-1", "keycloak-group-uuid-2"]
ui_permissions: { ... }
```

### DIFF
--- a/registry/repositories/documentdb/scope_repository.py
+++ b/registry/repositories/documentdb/scope_repository.py
@@ -76,14 +76,26 @@ class DocumentDBScopeRepository(ScopeRepositoryBase):
         collection = await self._get_collection()
 
         try:
-            group_doc = await collection.find_one({"_id": group_name})
-            if not group_doc:
-                logger.debug(f"DocumentDB READ: Group '{group_name}' not found")
-                return {}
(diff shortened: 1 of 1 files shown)


ndrsfel:
Fixed with #378 by @aarora79, thanks!

PR #398, opened 2026-01-25
Add FastEmbed as fallback embedding provider with pre-baked model support

erica-prog:
The steps I followed when reviewing the design document:
- Added the FastEmbedClient class with ONNX-based inference.
- Implemented automatic fallback in case the primary provider fails.
- Pre-baked the BAAI/bge-small-en-v1.5 model into the Docker image.
- Added a health endpoint with embedding status (registry/api/health_routes.py).
- Configured fallback model and dimensions.
- Included comprehensive unit tests for the new functionality.

*Issue #, if available:*  When the other scripts were created, the registry/api/health_routes.py was missing, as there was no file for the api folder. 

*Description of changes:* I reviewed the design document's Health Endpoint Enhancement section and implemented the required file as specified. The task was completed in two steps: first, following the document instructions for each file and providing an example snippet in "Section 2: Factory Function Enhancement" to guide the process. After reviewing the generated files, I issued a second step to ensure compliance with the instructions for "6. Health Endpoint Enhancement," specifically to create the registry/api/health_routes.py file and update related files accordingly.


aarora79:
Thanks for this contribution! We appreciate the effort to add FastEmbed as a fallback embedding provider.

We ended up implementing a different approach for handling embedding model unavailability. Instead of adding an alternative embedding provider, we added lexical fallback search that automatically kicks in when any embedding model is unavailable. This keeps the search system operational using keyword matching (regex on path, name, description, tags, and tools) without requiring additional embedding dependencies.

The implementation is in PR #415 and includes:
- Automatic lexical-only search fallback when the embedding model fails to load
- HNSW recall improvements (efSearch tuning, minimum k=50)
- Score-before-filter pattern for better result ranking
- Consistent scoring across DocumentDB and MongoDB CE backends

See https://github.com/agentic-community/mcp-gateway-registry/pull/415 for details.

PR #424, opened 2026-02-02
Feature: Add Device Code Flow CLI for End-User Authentication

ndrsfel:
Hey Amit, here is an idea/concept on how we could support the device code flow for developers. We add a new CLI tool (`mcp-gateway`) that provides AWS SSO-like authentication for end users, enabling seamless programmatic access to MCP Gateway Registry without requiring Azure App Registration Client Secrets.

Especially point 3) below caused a lot of friction for us. Let me know what you think 🚀 

----

### The Problem Today

Currently, developers who want to interact with the registry programmatically face significant friction:

1. **No unified CLI exists**: The registry has a web UI and REST API, but no official command-line tool for end users.

2. **Token acquisition is fragmented**: Multiple scripts exist for different purposes, but none provide a "login once, use everywhere" experience:
   - `api/registry_management.py` - Powerful but requires externally-obtained tokens via `--token-file`
   - `cli/get_user_token.py` - Obtains MS Graph tokens that don't map to registry permissions
   - `scripts/refresh_m2m_token.sh` - Service accounts only (M2M), not for interactive users

3. **Scope mismatch**: The existing token tools obtain tokens with MS Graph scopes, but the registry requires session cookies with group-to-scope mappings. These are fundamentally incompatible - you can't just pass a raw Entra token to the registry API ❗ 

4. **No session persistence**: Every API call requires manual token management, unlike tools like `aws sso login` that handle this automatically.

### The Developer Journey We Want

```bash
# First time setup (once)
export ENTRA_TENANT_ID="..."
export ENTRA_CLIENT_ID="..."

# Login (once per session, ~8 hours)
$ mcp-gateway login
To sign in:
  1. Open: https://microsoft.com/devicelogin
  2. Enter code: ABCD-EFGH

Logged in as: developer@company.com

# Now just use it - no more thinking about auth
$ mcp-gateway mcp list
+------------------+------------+-------+---------+
| Name             | Path       | Tools | Status  |
+------------------+------------+-------+---------+
| CurrentTime      | /time      | 2     | healthy |
| Weather Service  | /weather   | 5     | healthy |
+------------------+------------+-------+---------+

$ mcp-gateway agent list --enabled
+------------------+------------+--------+--------+---------+
| Name             | Path       | Skills | Trust  | Enabled |
+------------------+------------+--------+--------+---------+
| Travel Assistant | /travel    | 3      | medium | Yes     |
+------------------+------------+--------+--------+---------+

# Session expired? Clear message.
$ mcp-gateway mcp list
Session expired. Run: mcp-gateway login
```

This is the same experience developers expect from `aws sso login`, `gh auth login`, or `gcloud auth login`.

---

### 1. New Auth Server Endpoint: `/oauth2/exchange-token`

A minimal, secure endpoint that exchanges a verified Entra ID token for a registry session cookie.

**Request:**
```json
POST /oauth2/exchange-token
{"id_token": "eyJ..."}
```

**Response:**
```json
{
  "session_cookie": "...",
  "username": "user@company.com",
  "groups": ["group-uuid-1", "group-uuid-2"],
  "scopes": ["mcp-servers-unrestricted/read", "..."],
  "expires_in": 28800
}
```

**Security:**
- Validates ID token signature using Entra's JWKS (public keys)
- Verifies audience matches the configured client_id
- Checks token expiration
- Only accepts tokens from the configured Entra tenant

### 2. New CLI Tool: `mcp-gateway`

A Typer-based CLI that mimics the AWS SSO login experience:

```bash
# Authenticate (device code flow)
mcp-gateway login

# Check current session
mcp-gateway whoami

# List MCP servers
mcp-gateway mcp list
mcp-gateway mcp list --query "weather" --format json

# List A2A agents
mcp-gateway agent list
mcp-gateway agent list --enabled --format json

# Clear session
mcp-gateway logout

# View/set configuration
mcp-gateway config
mcp-gateway config --set registry_url=https://mcp-gateway.example.com
```

**Session Management:**
- Session stored in `cli/.mcp_session` with `chmod 600`
- Automatic session expiry detection with helpful re-login prompts
- Configuration via environment variables or `cli/.mcp_config.json`

---

## Comparison with Existing Approaches

| Approach | User Type | Session Mgmt | Multi-Provider | Dependencies |
|----------|-----------|--------------|----------------|--------------|
| `api/registry_management.py` | Admin | None (manual token) | N/A | requests |
| `cli/get_user_token.py` | Developer | None (file only) | Entra only | requests |
| `scripts/refresh_m2m_token.sh` | Service | None | Keycloak only | curl, jq |
| **`mcp-gateway` CLI (new)** | **End User** | **Automatic** | **Via auth server** | msal, typer, rich |

---

## Limitations

1. **Entra ID Only**: The current implementation focuses on Entra ID. Other providers (Keycloak, Cognito) would require similar `exchange-token` logic with their respective JWKS validation.

2. **No Token Refresh**: Sessions last 8 hours. Users must re-run `login` after expiry. Future enhancement could add automatic refresh using MSAL's token cache.

3. **MSAL Dependency**: The CLI uses MSAL for device code flow. A lighter implementation using raw HTTP requests (like `get_user_token.py`) could be added as an alternative.

4. **Local Session Storage**: Session is stored in a file. For shared environments, consider environment variable support or system keyring integration.

---

## Entra ID App Registration Requirements

The app registration must have:
- **Public client flows enabled** (for device code flow)
- **ID tokens enabled** (for the exchange endpoint)
- Optional: Group claims configured in token configuration

---

## Future Enhancements

1. **Multi-provider support**: Add Keycloak and Cognito token exchange
2. **Token refresh**: Leverage MSAL token cache for automatic refresh
3. **System keyring**: Store sessions in OS keyring instead of file
4. **Shell completion**: Add bash/zsh completion scripts
5. **MCP server management commands**: `mcp-gateway mcp enable/disable/register`

----

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -2404,3 +2404,115 @@ async def oauth2_logout(provider: str, request: Request, redirect_uri: str = Non
             "success_redirect", "/login"
         )
         return RedirectResponse(url=redirect_url, status_code=302)
+
+
+@app.post("/oauth2/exchange-token", tags=["OAuth2"])
+async def exchange_token_for_session(request: Request):
+    """Exchange a verified Entra ID token for a registry session cookie.
+
+    This enables CLI/device code flow authentication:
+    1. Client authenticates with Entra directly (device code flow)
+    2. Client sends the id_token here
+    3. Server validates the token signature and claims
+    4. Returns a session cookie for direct registry API access
+
+    Request body: {"id_token": "eyJ..."}
+    Returns: {"session_cookie": "...", "username": "...", "expires_in": 28800}
+    """
+    try:
+        body = await request.json()
+        id_token = body.get("id_token")
+
+        if not id_token:
+            raise HTTPException(status_code=400, detail="id_token is required")
+
+        # Get Entra provider for token validation
+        if "entra" not in OAUTH2_CONFIG.get("providers", {}):
+            raise HTTPException(status_code=400, detail="Entra provider not configured")
+
+        provider_config = OAUTH2_CONFIG["providers"]["entra"]
+        if not provider_config.get("enabled", False):
+            raise HTTPException(status_code=400, detail="Entra provider is disabled")
+
+        auth_provider = get_auth_provider("entra")
+
+        # Validate ID token signature and claims using Entra's JWKS
+        try:
+            # Get JWKS for signature verification
+            jwks = auth_provider.get_jwks()
+
+            # Get the key ID from token header
+            unverified_header = jwt.get_unverified_header(id_token)
+            kid = unverified_header.get("kid")
+
+            if not kid:
+                raise HTTPException(status_code=400, detail="Token missing key ID")
+
+            # Find the signing key
+            signing_key = None
+            for key in jwks.get("keys", []):
+                if key.get("kid") == kid:
+                    signing_key = PyJWK(key).key
+                    break
+
+            if not signing_key:
+                raise HTTPException(status_code=400, detail="Token signing key not found")
+
+            # Validate and decode token with full verification
+            claims = jwt.decode(
+                id_token,
+                signing_key,
+                algorithms=["RS256"],
+                audience=auth_provider.client_id,  # Must match our app's client_id
+                options={"verify_exp": True, "verify_aud": True}
+            )
+
+        except jwt.ExpiredSignatureError:
+            raise HTTPException(status_code=401, detail="Token has expired")
+        except jwt.InvalidAudienceError:
+            raise HTTPException(status_code=401, detail="Token audience mismatch")
+        except jwt.InvalidTokenError as e:
+            logger.error(f"Token validation failed: {e}")
+            raise HTTPException(status_code=401, detail="Invalid token")
+
+        # Extract user info from validated claims
+        username = (
+            claims.get("preferred_username") or
+            claims.get("email") or
+            claims.get("upn") or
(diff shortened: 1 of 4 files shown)


aarora79:
I like the idea of a CLI, but as an abstraction on top of the `api/registry_management.py`. The api/registry_management.py is not limited to admin use, in fact it could be whatever JWT token admin/regular user/M2M that you have. When we click the `Get JWT Token` from the top-left corner of the UI, we get a self-signed JWT token with an 8-hour validity that has the exact same scope access as the user who generated that token, and for the M2M use-case you could anyway generate JWT tokens via the IdP. So this abstracts out the IdP part and works for any IdP simply because the IdP based checks have been done when the user logged in. See https://github.com/agentic-community/mcp-gateway-registry/blob/main/docs/design/authentication-design.md#part-2-programmatic-access-self-signed-jwt-tokens.

Also, thank you for your contributions and thoughtful issues, would like to understand your use-case better, please feel free to write to me at aroraai@amazon.com.

ndrsfel:
Hey Amit, thanks for sharing the link. This really helped to understand the /validate and token-exchange given the registry management script and nginx redirect. Great to see these hidden 💎 already built into the library 🙏

aarora79:
Glad this was useful!

aarora79:
I still like your idea about the cli though, something that we can work on together.

PR #429, opened 2026-02-04
feat: Add OAUTH_STORE_TOKENS_IN_SESSION env var

### LINKED ISSUE #399: feat: Implement server-side token storage to resolve session cookie size limit
## Problem

OAuth session cookies exceed the browser's 4096-byte limit when storing JWT access and refresh tokens from identity providers. In our setup the response from the `http://localhost:8888/oauth2/callback/entra` is too large.

Image

**Current cookie contents:**
- User metadata (username, email, groups): ~500 bytes
- Access token (JWT): ~2-3 KB
- Refresh token (JWT): ~2-3 KB
- Serialization overhead (~30%)
- **Total: ~7 KB** (exceeds 4096 byte browser limit)

This causes authentication failures when using OAuth providers that issue larger JWTs (e.g., Microsoft Entra ID with group claims, Keycloak with custom attributes).

**Current workaround:** OAuth tokens are commented out from the session cookie, which means the "Get JWT Token" functionality is unavailable for OAuth users -> https://github.com/agentic-community/mcp-gateway-registry/blob/13990fb6febeb54ed11223976af7d53ebb38740a/auth_server/server.py#L1945

## Proposed Solution: Token Reference Pattern

Store OAuth tokens in MongoDB separately, while the session cookie holds only user metadata and a token reference ID.

### Implementation

1. **Create `TokenStore` class** for encrypted token storage in MongoDB
2. **Modify OAuth callback** to store tokens in MongoDB instead of the cookie
3. **Add token reference** (`token_ref`) to session cookie (~16 bytes)
4. **New endpoint** `GET /api/tokens/current` to retrieve tokens on-demand
5. **TTL index** for automatic cleanup of expired tokens

### Benefits

- Session cookie size reduced to ~600 bytes (well under 4096 limit)
- No performance impact on normal requests (cookie-only validation)
- Tokens loaded on-demand only when needed
- Backward compatible with existing authentication flow
- Tokens encrypted at rest in MongoDB
- Automatic cleanup via TTL index

### MongoDB Schema

```javascript
// Collection: oauth_tokens
{
  token_ref: "secure-random-id",
  username: "user@example.com",
  access_token_encrypted: "base64...",
  refresh_token_encrypted: "base64...",
  expires_in: 3600,
  created_at: ISODate,
  expires_at: ISODate  // TTL index
}
```

### Files Changed

- `auth_server/token_store.py` (new)
- `auth_server/server.py` (OAuth callback + new endpoint)
- `auth_server/pyproject.toml` (add `cryptography` if missing)

ndrsfel:
OAuth session cookies can exceed the browser's 4096-byte limit when storing JWT access and refresh tokens from identity providers like Microsoft Entra ID.

This adds`OAUTH_STORE_TOKENS_IN_SESSION` environment variable to optionally disable storing OAuth tokens in the session cookie:
- Default: true (backward compatible - tokens stored)
- For Entra ID: Set to false to avoid cookie size errors

The tokens stored in the session were never actually used by any code, so disabling storage has no functional impact.

*Closes #399*


aarora79:
Thanks @ndrsfel for this contribution! Your PR correctly identified the cookie size issue with large IdP tokens and the solution approach is sound.

We've incorporated your changes into a new PR from branch `feat/oauth-token-storage-terraform` which includes your commits plus the following additions:

- Changed default to `false` (tokens are not used functionally, safer default)
- Added Terraform ECS support (variables, module plumbing, task definitions)
- Surfaced all missing auth server config params in the system config UI panel
- Added an integration test that exercises the actual `oauth2_callback` endpoint (vs the unit tests that re-implement parsing logic locally)
- Logging changed to INFO on startup (always logs the setting value)

Closing this PR in favor of the consolidated one. Thanks again for the contribution!

PR #431, opened 2026-02-04
fix: Use nginx_proxied_auth for /api/auth/me to support Bearer tokens

ndrsfel:
The `/api/auth/me` endpoint was using `enhanced_auth` dependency which only supports session cookies. This made it inconsistent with other API endpoints like `/api/servers` that use `nginx_proxied_auth`.

Changed to `nginx_proxied_auth` which supports both:
- Bearer tokens (via nginx-proxied X-User headers after JWT validation)
- Session cookies (fallback)

This allows API clients using Bearer token authentication to access the `/api/auth/me` endpoint for user info.

### DIFF
--- a/registry/main.py
+++ b/registry/main.py
@@ -32,8 +32,8 @@ from registry.health.routes import router as health_router
 
 # Import auth dependencies
 from registry.auth.dependencies import (
-    enhanced_auth,
     get_ui_permissions_for_user,
+    nginx_proxied_auth,
 )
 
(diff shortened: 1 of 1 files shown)


aarora79:
Note: This PR is missing the two nginx config files (nginx_rev_proxy_http_and_https.conf and nginx_rev_proxy_http_only.conf) needed to complete this fix. @aarora79 will add these in a separate PR.

PR #511, opened 2026-02-20
Feat/DCR

ryohang:
*Issue #, if available:*

*Description of changes:*
support DCR


aarora79:
For traceability: the decision to **not** implement server-side DCR in the registry is now recorded as an ADR — [docs/adr/0001-no-server-side-dcr.md](https://github.com/agentic-community/mcp-gateway-registry/blob/main/docs/adr/0001-no-server-side-dcr.md) (tracked by #995).

The ADR bans only the registry acting as a DCR *authorization server* (a registry-hosted `/register` endpoint, which is what this PR attempted). It does **not** forbid delegated DCR against the upstream IdP, nor the gateway acting as a DCR *client* on egress (e.g. Atlassian Rovo `authv2` in #1519). See the ADR for the rationale and the revisit trigger.

PR #513, opened 2026-02-20
bug: Handle duplicate inserts into the audit events DB

snorlaX-sleeps:
### **Description of changes:**
When the auth-server receives a request it inserts the request ID into the DB as an audit event.
Many of the registry APIs also insert the same data into the DB with the same ID, which emits an error each time, since the nginx request ID is identical in both cases.

This change returns true on audit events that already exist in the DB, rather than emitting an error on each API call, which can make debugging harder than it should be.

### **Alternatives with more code changes**
The code could be changed to `upsert` or update the audit DB entry, since it needs to go through validation and onto the registry API anyway.
Or we could remove the audit logic from one or more parts of the code

### DIFF
--- a/registry/repositories/audit_repository.py
+++ b/registry/repositories/audit_repository.py
@@ -235,7 +235,8 @@ class DocumentDBAuditRepository(AuditRepositoryBase):
             record: The audit record to insert (RegistryApiAccessRecord or MCPServerAccessRecord)
 
         Returns:
-            True if inserted successfully, False otherwise
+            True if inserted successfully or if the record already exists (duplicate request_id),
+            False if an unexpected error occurs
         """
         logger.debug(
             f"DocumentDB WRITE: Inserting audit event with request_id={record.request_id}"
@@ -255,7 +256,7 @@ class DocumentDBAuditRepository(AuditRepositoryBase):
                 f"DocumentDB WRITE: Inserted audit event request_id={record.request_id}"
             )
             return True
-        except DuplicateKeyError as e:
+        except DuplicateKeyError:
             logger.debug(
                 f"DocumentDB WRITE: Skipped duplicate audit event for request_id={record.request_id}. "
                 f"This occurs when the same request_id is processed twice (auth validation + endpoint execution). "
--- a/tests/unit/audit/test_audit_repository.py
+++ b/tests/unit/audit/test_audit_repository.py
@@ -5,6 +5,7 @@ Validates: Requirements 6.1, 6.2
 """
 
 from datetime import datetime, timezone
+from pymongo.errors import DuplicateKeyError
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


aarora79:
Thanks for this PR, this is definitely a problem to be fixed. Let me think through about the best way to fix the root cause. Will update here.

aarora79:
## Review Feedback

Good catch. We will merge this PR after the changes listed below. Your PR correctly identified a real issue in the audit logging architecture:

**The auth server and the registry middleware both insert audit events for the same MCP proxy request using the same `request_id`, but with different record types (`MCPServerAccessRecord` vs `RegistryApiAccessRecord`). These are complementary records with different data -- the MCP record captures protocol-level details (tool name, server, transport) while the registry record captures HTTP-level details (actual response code, end-to-end duration, authorization decision). The current unique index on `request_id` causes the second record to be rejected, silently losing data.**

The correct fix for this is a composite unique index on `(request_id, log_type)` which is tracked in #527. After that change, your `DuplicateKeyError` handler remains valuable as a safety net for true duplicates (retries, race conditions).

## Required Changes Before Merge

### 1. Remove unused exception variable

`registry/repositories/audit_repository.py:258`

```python
# Before
except DuplicateKeyError as e:

# After
except DuplicateKeyError:
```

The variable `e` is caught but never referenced in the log message.

### 2. Add unit test for DuplicateKeyError path

Add the following test to `tests/unit/audit/test_audit_repository.py` inside the `TestInsert` class:

```python
async def test_returns_true_on_duplicate_key(self):
    """insert() returns True when a duplicate audit event already exists."""
    mock_collection = AsyncMock()
    mock_collection.insert_one.side_effect = DuplicateKeyError("duplicate key error")

    with patch.object(DocumentDBAuditRepository, '_get_collection', return_value=mock_collection):
        repo = DocumentDBAuditRepository()
        repo._collection = mock_collection

        result = await repo.insert(make_test_record())

        assert result is True
```

Add `DuplicateKeyError` to the imports at the top of the test file:

```python
from pymongo.errors import DuplicateKeyError
```

### 3. Update the insert() docstring

`registry/repositories/audit_repository.py:227-239`

Update the Returns section to document the duplicate handling:

```python
"""
Insert an audit event record.

Args:
    record: The audit record to insert (RegistryApiAccessRecord or MCPServerAccessRecord)

Returns:
    True if inserted successfully or if the record already exists (duplicate request_id),
    False if an unexpected error occurs
"""
```

aarora79:
Hi @snorlaX-sleeps please make the changes requested here https://github.com/agentic-community/mcp-gateway-registry/pull/513#issuecomment-3941698422 and then I can merge this.

snorlaX-sleeps:
@aarora79 requested changes have been made

aarora79:
All three requested changes addressed -- unused exception variable removed, unit test added, docstring updated. Tests pass (1512 passed, 46.51% coverage). Merging.

PR #514, opened 2026-02-20
Feature: Allow setting the nodeSelector for applications deployed from the `mcp-gateway-registry-stack` Helm Chart

snorlaX-sleeps:
*Description of changes:*
Many scaling / provisioning tools for larger clusters / workloads allow the use of `nodeSelector` to decide the compute to allocate to all of the pods. This will set for all charts in the repo as well as `KeyCloak`, `Postgres` and `MongoDB`
Does not set for the Job pods, as they are small and temporary (can be added)

### DIFF
--- a/charts/mcp-gateway-registry-stack/templates/mongodb-cluster.yaml
+++ b/charts/mcp-gateway-registry-stack/templates/mongodb-cluster.yaml
@@ -26,12 +26,12 @@ spec:
       scramCredentialsSecretName: my-scram
   additionalMongodConfig:
     storage.wiredTiger.engineConfig.journalCompressor: zlib
+  {{- with .Values.global.nodeSelector }}
   statefulSet:
     spec:
       template:
         spec:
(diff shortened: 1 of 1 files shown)


omrishiv:
Hi @snorlaX-sleeps, thank you for the PR. Is there a reason to use a global nodeselector only for Mongo? Does it make sense to have one global that would apply to all apps, but allow overriding with app specific values?

snorlaX-sleeps:
> Hi @snorlaX-sleeps, thank you for the PR. Is there a reason to use a global nodeselector only for Mongo? Does it make sense to have one global that would apply to all apps, but allow overriding with app specific values?

This can be done - the MongoDB cluster is the only thing directly deployed by this Helm Chart, hence the global.
I can look into how this would be done, as I would need to figure out setting global variables for child charts we do not control (could use the anchor from https://github.com/agentic-community/mcp-gateway-registry/pull/517

PR #515, opened 2026-02-20
Bug: Auth-Server admin username / password was missing from the Helm Chart

snorlaX-sleeps:
*Description of changes:*
This is required for at least one endpoint. This PR allows adding a username and password, which is currently only used for one internal API. Does not set a default, however if `adminUser` is set then `adminPassword` must also be set.

Unsure if this was previously part of the `.env` config docs and just missing from the Helm Chart

### DIFF
--- a/charts/auth-server/templates/secret.yaml
+++ b/charts/auth-server/templates/secret.yaml
@@ -66,6 +66,10 @@ data:
   REGISTRY_ROOT_PATH: {{ $registryRootPath | b64enc | quote }}
   SESSION_COOKIE_SECURE: {{ .Values.app.sessionCookieSecure | toString | b64enc | quote }}
   SESSION_COOKIE_DOMAIN: {{ printf ".%s" $domain | b64enc | quote }}
+  {{- if .Values.app.adminUser}}
+  ADMIN_USER: {{ .Values.app.adminUser | b64enc | quote }}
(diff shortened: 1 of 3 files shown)


omrishiv:
@aarora79 do we want to allow this endpoint to work with the admin scope so we don't need local account? I was under the assumption at this point that keycloak or otherwise is required

snorlaX-sleeps:
@omrishiv I was debugging adding Anthropic MCP servers and noticed that these credentials are also used in the registry [scope-service.py](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/services/scope_service.py#L48-L49)  to reload the scopes when adding a new MCP server but not provided in the Helm Chart

omrishiv:
@snorlaX-sleeps thank you for the clarifications on these PRs. I am about to open one that will use the SECRET_KEY as a JWT rather than requiring the admin username/password for internal communication. What do you think?

aarora79:
Closing this in favor of #533, which implements the approach discussed here -- replacing HTTP Basic Auth with self-signed JWT bearer tokens (using the shared SECRET_KEY) for all internal service-to-service communication.

Thank you @tofaramususa for raising this issue and the thoughtful analysis of the credential architecture. Your work identifying the problem and proposing the direction was instrumental in getting this resolved.

PR #516, opened 2026-02-21
Feature: Allow Helm Chart to not deploy the keycloak-ingress-patch

snorlaX-sleeps:
*Description of changes:*
All other ingresses that can be deployed can be disabled, adding variables to allow this for the patch also.
Useful when cluster has a service-mesh controlling this functionality (e.g istio as mentioned in the documentation)

### DIFF
--- a/charts/mcp-gateway-registry-stack/values.yaml
+++ b/charts/mcp-gateway-registry-stack/values.yaml
@@ -34,7 +34,6 @@ global:
 
   # Common ingress settings
   ingress:
-    enabled: true
     inboundCidrs: # optional comma separated list of allowed inbound CIDR ranges
     className: alb
     tls: true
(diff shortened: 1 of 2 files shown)


charts/mcp-gateway-registry-stack/values.yaml
```diff
@@ -34,6 +34,7 @@ global:
 
   # Common ingress settings
   ingress:
+    enabled: true
```
omrishiv: Can we scope this to just this chart? This naming makes it seem like it will disable all ingresses

snorlaX-sleeps: There wasn't a section in the values file for this, so it was included where the other variables are.
I will update the PR

snorlaX-sleeps: updated

snorlaX-sleeps:
@omrishiv - review addressed

PR #517, opened 2026-02-21
Bug: Keycloak auth required for Registry API endpoints missing from Helm Chart

snorlaX-sleeps:
*Description of changes:*
When using the [registry_management](https://github.com/agentic-community/mcp-gateway-registry/blob/main/api/registry_management.py) script, the `user-list` command was failing with a `http 502` error without any more information, while other endpoints were successful .
After adding debug logging it became clear that the authn / authz was correct (auth-server confirming permissions), however the variables required to query KeyCloak were not available and the code was silently failing.

Updated the Helm Chart:
- added global variables for username and realm
- used "anchors" to avoid repeating username and realm across the various services (can be changed or can add more comments / documentation)
- Set everything to the current defaults
- Added the KeyCloak password secret as a new env var instead of trying to pass it through the Helm Chart
- Set REALM and ADMIN in the registry secret

Added the logging in the `try / catch` as there was no information on why the `502` was occurring. Can add more logging if it would be informative

### DIFF
--- a/registry/api/management_routes.py
+++ b/registry/api/management_routes.py
@@ -136,7 +136,9 @@ async def management_list_users(
 
     try:
         raw_users = await iam.list_users(search=search, max_results=limit)
+        logger.debug(f"[LIST_USERS] Retrieved {len(raw_users)} users from IAM")
     except Exception as exc:
+        logger.error(f"[LIST_USERS] Exception calling list_users: {type(exc).__name__}: {exc}")
         raise _translate_iam_error(exc) from exc
 
     summaries = [
--- a/registry/auth/dependencies.py
+++ b/registry/auth/dependencies.py
@@ -630,7 +630,7 @@ async def nginx_proxied_auth(
         # Set user context on request state for audit logging middleware
         request.state.user_context = user_context
 
-        logger.debug(f"nginx-proxied auth context for {username}: {user_context}")
+        logger.debug(f"nginx-proxied auth context for {username} (is_admin={is_admin}): {user_context}")
         return user_context
 
     # Fallback to session cookie authentication


omrishiv:
@snorlaX-sleeps sorry for the confusion, is this related to #515?

snorlaX-sleeps:
@omrishiv - they are similar, but no.
Auth Server is looking for its own admin credentials, while this registry endpoint needs the **keyCloak** admin credentials. Unsure if it's all user / group registry interactions 
The auth-server admin creds are only for that one endpoint as far as I can tell

omrishiv:
Sorry, for all the questions on my part, but how would this work if we're using an external OIDC provider? We currently support Entra, but between this and #515 , I think we may have a gap. What do you think @aarora79 ?

snorlaX-sleeps:
@omrishiv - unfortunately I've only deployed the gateway-registry with KeyCloak so far, I have not yet looked into other authn providers so can't comment

omrishiv:
@snorlaX-sleeps I think we have a issue here where we rely on either the local account or the fact that we're using a local KeyCloak (rather than Entra or an external keycloak or other OIDC provider). We could take the PRs you've opened, which, thank you for. But we still need to address this issue at a more fundamental level. 

For both this PR #517 and #515, how blocked are you? We intend to remove the local user altogether and will need to properly address this issue of assuming a local KeyCloak

snorlaX-sleeps:
@omrishiv I'm not blocked in the sense of I've built and uploaded my own version of the container to use in Kubernetes with all the PR changes I've opened while we are investigating if we can use `mcp-gateway-registry`.

> We intend to remove the local user altogether and will need to properly address this issue of assuming a local KeyCloak

I am unsure what you mean by this, will there not need to be some sort of user required for API authentication?

omrishiv:
@aarora79 to comment more on this, but I believe the plan is to either require a local Keycloak instance or use an external provider rather than a single local account for admin

omrishiv:
Ok, starting to make sense of these PRs, thank you for sticking with me. Do you think it would make sense to pull the keycloak/entra related items into a separate secret that can be shared between the registry/auth-server? The way this is written, it only supports keycloak, which makes sense as that's the default. Otherwise, happy to merge this and refactor it after to address the missing Entra variables

aarora79:
Thanks @omrishiv and @snorlaX-sleeps for the discussion. So will merge this one for now so that Keycloak and open up a separate PR for Entra.

snorlaX-sleeps:
> Do you think it would make sense to pull the keycloak/entra related items into a separate secret that can be shared between the registry/auth-server?

It could be worth it - there could be a single secret supporting both or even external authentication, with interpolation based on the settings. This single secret could be mounted on all the pods

PR #532, opened 2026-02-23
Bug: Registry - missing ROOT_PATH in generated nginx config

snorlaX-sleeps:
*Description of changes:*
When the registry generates or regenerates the nginx config it creates API configurations for MCP Servers and Virtual Gateways, as well as a further configuration block for Registry-only mode.

However none of these paths include the `{{ROOT_PATH}}` which is present for the other routes in the nginx config.
This PR updates these routes to also template in this variable. The string replacement has also been moved further down the file to ensure all of the additional paths are present in the file before replacing this variable.

Unsure if this was intentional and has just been misunderstood, as with path based routing via ISTIO these URL paths are not reachable externally without `{{ROOT_PATH}}` being present

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,7 +180,7 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
+.idea/
 
 # Ruff stuff:
 .ruff_cache/
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -19,7 +19,7 @@ from contextlib import asynccontextmanager
 from datetime import datetime
 from pathlib import Path
 from string import Template
-from typing import Any, Optional
+from typing import Any
 from urllib.parse import urlparse
 
 import boto3
@@ -42,14 +42,13 @@ from providers.factory import get_auth_provider
 from pydantic import BaseModel
 
 sys.path.insert(0, "/app")
-from registry.common.scopes_loader import reload_scopes_config
-from registry.repositories.factory import get_scope_repository
-
 # Import MCP audit logging components
 from registry.audit.mcp_logger import MCPLogger
-from registry.audit.service import AuditLogger
 from registry.audit.models import Identity, MCPServer
+from registry.audit.service import AuditLogger
+from registry.common.scopes_loader import reload_scopes_config
 from registry.core.config import settings
+from registry.repositories.factory import get_scope_repository
 from registry.utils.request_utils import get_client_ip
 
 # Configure logging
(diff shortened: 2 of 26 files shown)


aarora79:
Good catch -- the missing `{{ROOT_PATH}}` in dynamically generated nginx location blocks is a real issue for Istio path-based routing deployments. The approach of moving the replacement after all blocks are assembled is the right architectural fix.

However, there are 3 issues that need to be addressed before we can merge.

## Required Changes Before Merge

### 1. Fix 9 failing unit tests

The PR changes the output of `_create_location_block()` and related functions to include `{{ROOT_PATH}}`, but the existing tests still expect the old output without it.

**Failing tests in `tests/unit/core/test_nginx_service.py`:**
- `test_create_location_block_sse`
- `test_create_location_block_streamable_http`
- `test_create_location_block_external_service`
- `test_create_location_block_internal_service`
- `test_create_location_block_direct_transport`
- `test_generate_transport_location_blocks_streamable_http`
- `test_generate_transport_location_blocks_sse`
- `test_generate_transport_location_blocks_both_transports`
- `test_generate_transport_location_blocks_no_transports`

For example, `test_create_location_block_sse` asserts `"location /test" in block` but the code now produces `location {{ROOT_PATH}}/test/`. Update the assertions to expect `{{ROOT_PATH}}` in the output.

### 2. Remove trailing slash addition from `_create_location_block()`

`registry/core/nginx_service.py:1176`

```python
# Before (your change)
location {{{{ROOT_PATH}}}}{location_path}/ {

# Should be
location {{{{ROOT_PATH}}}}{location_path} {
```

The trailing `/` is a separate behavioral change from the ROOT_PATH fix. In nginx, `location /foo` and `location /foo/` have different matching semantics. The comment on line 1171 explicitly says "Users have full control over the location path format (with or without trailing slash)" -- adding a forced `/` contradicts this. Please keep `{location_path}` without the trailing `/`.

### 3. Remove ROOT_PATH from internal nginx locations

`registry/core/nginx_service.py:850` (`_generate_virtual_backend_locations`)

```python
# Before (your change)
location {{{{ROOT_PATH}}}}{location_path} {{

# Should be (revert to original)
location {location_path} {{
```

The `/_vs_backend*` locations have the `internal;` directive, meaning they are only accessible via nginx subrequests (e.g., from `proxy_pass` within other location blocks), not from external URLs. These should NOT have `ROOT_PATH` prepended because:

- They are never hit by external requests
- The subrequest paths are hardcoded elsewhere without ROOT_PATH
- Adding ROOT_PATH here would break the virtual server backend routing

## Summary

The core fix (adding ROOT_PATH to external-facing generated blocks and moving the replacement after assembly) is correct. Just needs the 3 items above addressed and all tests passing before merge.

aarora79:
@snorlaX-sleeps please resubmit after addressing the issues pointed out here https://github.com/agentic-community/mcp-gateway-registry/pull/532#issuecomment-3946928932.

snorlaX-sleeps:
Will do @aarora79 - thank you for the swift response
I'd not used `/_vs_backend*` yet, thank you for the clarification

snorlaX-sleeps:
@aarora79 - requested changes made, tests pass

PR #533, opened 2026-02-23
Sign SECRET_KEY as JWT for internal communication

omrishiv:
*Issue #, if available:*

*Description of changes:*

Switches from using admin username/password basic auth to JWT bearer auth for internal communication.

Should address #515 

### DIFF
--- a/charts/mcp-gateway-registry-stack/values.yaml
+++ b/charts/mcp-gateway-registry-stack/values.yaml
@@ -29,6 +29,10 @@ global:
   authProvider:
     # Provider type: "keycloak" or "entra"
     type: keycloak
+    keycloak:
+      adminUsername: &keycloakAdmin "user"
+      realm: &keycloakRealm "mcp-gateway"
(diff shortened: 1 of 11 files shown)


auth_server/server.py
```diff
@@ -2002,52 +1999,96 @@ async def generate_user_token(request: GenerateTokenRequest):
...
-    admin_password = os.environ.get("ADMIN_PASSWORD")
+    caller_identity = "unknown"
+
+    if authorization.startswith("Bearer "):
+        # Validate self-signed JWT using shared SECRET_KEY
+        token = authorization.split(" ", 1)[1]
+        try:
+            claims = jwt.decode(
+                token,
+                SECRET_KEY,
```
snorlaX-sleeps, line 2026: We need to look this up from the ENV VARs as we haven't used it at this point?

omrishiv: We assign this when the module is first loaded on L2225

aarora79:
Verified on EC2 and EKS, merging.

Issue #520, opened 2026-02-21
Fix Bandit B104 findings: Replace hardcoded bind to all interfaces (11 occurrences)


aarora79:
## Parent Issue

Tracking issue: #519

## Summary

Fix Bandit **B104 (hardcoded_bind_all_interfaces)** -- 11 occurrences where services bind to `0.0.0.0`, exposing them on all network interfaces.

**Phase:** 2 - Medium Complexity
**Severity:** MEDIUM

## Problem

Binding to `0.0.0.0` exposes services on all network interfaces, which can be a security risk if the service is not intended to be publicly accessible. Services should bind to `127.0.0.1` for local-only access or to a specific private IP when external access is needed.

Bandit rule: [B104 - hardcoded_bind_all_interfaces](https://bandit.readthedocs.io/en/latest/plugins/b104_hardcoded_bind_all_interfaces.html)

### agents/ (2 occurrences)

| File | Line | Code |
|------|------|------|
| `agents/a2a/src/flight-booking-agent/env_settings.py` | 33 | `self.host: str = "0.0.0.0"` |
| `agents/a2a/src/travel-assistant-agent/env_settings.py` | 33 | `self.host: str = "0.0.0.0"` |

### auth_server/ (1 occurrence)

| File | Line | Code |
|------|------|------|
| `auth_server/server.py` | 2060 | `default="0.0.0.0"` |

### metrics-service/ (2 occurrences)

| File | Line | Code |
|------|------|------|
| `metrics-service/app/config.py` | 15 | `METRICS_SERVICE_HOST: str = os.getenv("METRICS_SERVICE_HOST", "0.0.0.0")` |
| `metrics-service/app/otel/exporters.py` | 27 | `start_http_server(port=..., addr="0.0.0.0")` |

### registry/ (1 occurrence)

| File | Line | Code |
|------|------|------|
| `registry/main.py` | 665 | `host="0.0.0.0"` |

### servers/ (5 occurrences)

| File | Line | Code |
|------|------|------|
| `servers/currenttime/server.py` | 49 | `FastMCP("CurrentTimeAPI", host="0.0.0.0", ...)` |
| `servers/example-server/server.py` | 51 | `FastMCP("ExampleMCPServer", host="0.0.0.0", ...)` |
| `servers/fininfo/server.py` | 575 | `mcp.run(..., host="0.0.0.0", ...)` |
[...]

Issue #519, opened 2026-02-21
Address Bandit static analysis findings (143 total)


aarora79:
## Summary

This is the tracking issue for addressing all findings from the Bandit static security analysis tool. A full scan identified **143 findings** across the codebase, grouped into 4 remediation phases by complexity and risk.

**Scan date:** 2026-02-21

## Findings Overview

| Rule ID | Count | Description | Severity |
|---------|-------|-------------|----------|
| B101 | 43 | assert_used | LOW |
| B603 | 19 | subprocess_without_shell_equals_true | LOW |
| B113 | 16 | request_without_timeout | MEDIUM |
| B105 | 16 | hardcoded_password_string | LOW |
| B404 | 16 | import_subprocess | LOW |
| B104 | 11 | hardcoded_bind_all_interfaces | MEDIUM |
| B110 | 9 | try_except_pass | LOW |
| B608 | 6 | hardcoded_sql_expressions | MEDIUM |
| B607 | 5 | start_process_with_partial_path | LOW |
| B307 | 1 | eval | MEDIUM |
| B310 | 1 | urllib_urlopen | MEDIUM |

### Phase 1: Easy Wins (16 findings)
- [ ] #518 - B113: Add missing request timeouts (16 findings)

### Phase 2: Medium Complexity (20 findings)
- [ ] #520 - B104: Fix binding to all interfaces (11 findings)
- [ ] #521 - B110: Improve error handling / remove try-except-pass (9 findings)

### Phase 3: Security-Critical (30 findings)
- [ ] #522 - B608: Fix SQL injection risks (6 findings)
- [ ] #523 - B603/B607: Secure subprocess calls (24 findings)

### Phase 4: Review and Document (77 findings)
- [ ] #524 - B101: Review assert usage in test vs production code (43 findings)
- [ ] #525 - B105: Review hardcoded password false positives (16 findings)
- [ ] #526 - B404/B307/B310: Review low-priority imports and patterns (18 findings)

## Acceptance Criteria

- [ ] All sub-issues resolved
- [ ] `uv run bandit -r .` returns 0 unaddressed findings (all remaining are documented with `# nosec` comments and justification)
- [ ] Existing tests still pass
- [ ] No functional regressions

aarora79:
All sub-issues have been successfully resolved! 🎉

## Completed Work

All 143 Bandit static analysis findings have been addressed across 4 remediation phases:

[...]

PR #534, opened 2026-02-23
fix: Bandit B104 findings across 11 files (#520)

abkrsinh:
*Issue: #520 (sub-issue of #519)*

## Summary

Running `bandit -r . -t B104` reported 11 unaddressed B104 (hardcoded_bind_all_interfaces) findings across Python files that bind to `0.0.0.0`. This PR resolves all 11 findings using a three-category approach while preserving all container runtime behavior.

## Changes

**Category 1 — Already configurable (1 file):**
- `metrics-service/app/config.py`: Add `# nosec B104` to the existing `os.getenv` line

**Category 2 — Make configurable via env var (4 files):**
- `auth_server/server.py`: `--host` default → `os.getenv("AUTH_SERVER_HOST", "127.0.0.1")`
- `registry/main.py`: host → `os.getenv("REGISTRY_HOST", "127.0.0.1")`
- `agents/a2a/src/flight-booking-agent/env_settings.py`: host → `os.getenv("AGENT_HOST", "0.0.0.0")`
- `agents/a2a/src/travel-assistant-agent/env_settings.py`: same as above

**Category 3 — Container-only, add suppression comment (6 files):**
- `metrics-service/app/otel/exporters.py`: Replace hardcoded `addr="0.0.0.0"` with `settings.METRICS_SERVICE_HOST`
- `servers/currenttime/server.py`, `servers/example-server/server.py`, `servers/fininfo/server.py`, `servers/mcpgw/server.py`, `servers/realserverfaketools/server.py`: Add `# nosec B104`

## Verification

- `bandit -r agents auth_server metrics-service registry servers -t B104` reports **0 findings**
- Docker entrypoints (`auth-entrypoint.sh`, `registry-entrypoint.sh`) pass `--host 0.0.0.0` explicitly, bypassing Python defaults — container behavior unchanged
- `docker-compose.yml` still sets `METRICS_SERVICE_HOST=0.0.0.0` — metrics service unchanged
- Agent default stays `0.0.0.0` since agents are always containerized

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -2112,8 +2112,8 @@ def parse_arguments():
     parser.add_argument(
         "--host",
         type=str,
-        default="0.0.0.0",
-        help="Host for the server to listen on (default: 0.0.0.0)",
+        default=os.getenv("AUTH_SERVER_HOST", "127.0.0.1"),  # nosec B104
+        help="Host for the server to listen on (default: 127.0.0.1, override with AUTH_SERVER_HOST env var)",
     )
 
     parser.add_argument(
--- a/agents/a2a/src/flight-booking-agent/env_settings.py
+++ b/agents/a2a/src/flight-booking-agent/env_settings.py
@@ -30,7 +30,7 @@ class EnvSettings:
         self.agent_url: str = os.getenv("AGENTCORE_RUNTIME_URL", "http://127.0.0.1:9000/")
 
         # Server configuration (fixed for A2A protocol)
-        self.host: str = "0.0.0.0"
+        self.host: str = os.getenv("AGENT_HOST", "0.0.0.0")  # nosec B104
         self.port: int = 9000
 
         logger.info(f"EnvSettings initialized: agent_name={self.agent_name}, version={self.agent_version}")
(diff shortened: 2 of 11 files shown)


auth_server/server.py
```diff
@@ -2069,8 +2069,8 @@ def parse_arguments():
     parser.add_argument(
         "--host",
         type=str,
-        default="0.0.0.0",
-        help="Host for the server to listen on (default: 0.0.0.0)",
+        default=os.getenv("AUTH_SERVER_HOST", "127.0.0.1"),  # nosec B104
```
omrishiv, line 2115: is this an issue if we use both the external auth server address `auth-server.domain.com` on the user end and `authserver.kubernetes.svc` internal domains for service to service requests?

abkrsinh: Hi Omri,
Thanks for the question. I don't think that should be an issue since the auth server addresses are dns names resolved separately from the bind address. 
Also, 127.0.0.1 default only applies when running the server entirely locally (no Docker) during development.

omrishiv: I was confusing this with AUTH_SERVER_URL, my mistake.

PR #620, opened 2026-03-08
fix: AgentCard streaming AttributeError and agent visibility consistency

### LINKED ISSUE #616: AgentCard.streaming AttributeError in discover_agents_by_skills (partial fix from #244 missed this route)
## Bug Description

`POST /api/agents/discover` crashes with a 500 Internal Server Error when it finds matching agents. The route tries to access `agent.streaming` on an `AgentCard` object, but `AgentCard` has no `streaming` attribute -- per the A2A spec, streaming is declared inside the `capabilities` dict.

```
File "/app/registry/api/agent_routes.py", line 1128, in discover_agents_by_skills
    streaming=agent.streaming,
AttributeError: 'AgentCard' object has no attribute 'streaming'
```

## Root Cause

The `AgentCard` model correctly follows the A2A spec by storing streaming capability in the `capabilities` dict (e.g., `{"streaming": true}`). The `AgentInfo` response model has a convenience `streaming: bool` field. The bug is in the route code that converts between them.

## History of Partial Fixes

This bug was partially fixed in commit `4812e21` (PR #244, Nov 23 2025) for the `list_agents` route:

```python
# list_agents (line 519-520) -- FIXED in 4812e21
streaming = agent.capabilities.get("streaming", False) if agent.capabilities else False
```

However, `discover_agents_by_skills` was missed in that fix and still has:

```python
# discover_agents_by_skills (line 1121) -- BROKEN
streaming=agent.streaming,
```

A later commit `6bc3100` (Dec 25 2025, "Return full AgentCard in semantic search results") incidentally fixed the same bug in `discover_agents_semantic` by refactoring it to return `agent_card.model_dump()` instead of converting to `AgentInfo`. But `discover_agents_by_skills` was again not addressed.

## Impact

- **Every** `POST /api/agents/discover` call that finds matching agents returns 500
- Skill-based agent discovery is completely non-functional
- Observed in production as a steady ~50% Tool Execution Error Rate in Grafana dashboards (when combined with load generator traffic hitting this endpoint)

## Three Locations That Need Attention

| Location | Line (on main) | Status | Issue |
|---|---|---|---|
| `list_agents` | 519-520 | Fixed | Correctly uses `agent.capabilities.get("streaming", False)` |
| `discover_agents_by_skills` | 1121 | **Broken** | Uses `agent.streaming` -- crashes with `AttributeError` |
| `register_agent` | 381 | Silent bug | Passes `streaming=request.streaming` to `AgentCard()` constructor, but `AgentCard` has no `streaming` field. Pydantic silently drops it. The value is never persisted. |
| `update_agent` | 899 | Silent bug | Same as `register_agent` -- streaming value silently dropped. |

### 1. Fix `discover_agents_by_skills` (crash fix)

Apply the same pattern already used in `list_agents`:

```python
# agent_routes.py, in discover_agents_by_skills (around line 1117)
# Before:
streaming=agent.streaming,

# After:
streaming=agent.capabilities.get("streaming", False) if agent.capabilities else False,
```

### 2. Fix registration and update routes (data loss fix)

Map `request.streaming` into `capabilities` when constructing the `AgentCard`:

```python
# In register_agent and update_agent, when constructing AgentCard:
capabilities = {"streaming": request.streaming} if request.streaming else {}

agent_card = AgentCard(
    ...
    capabilities=capabilities,  # instead of streaming=request.streaming
    ...
)
```

### 3. Also fix `provider` extraction in `discover_agents_by_skills`

The `list_agents` route correctly extracts the provider org name:
```python
provider_name = agent.provider.organization if agent.provider else None
```

But `discover_agents_by_skills` passes the raw `AgentProvider` object:
```python
provider=agent.provider,  # AgentInfo.provider expects str | None, not AgentProvider
```

## Reproduction Steps

1. Register an A2A agent with at least one skill
2. Call `POST /api/agents/discover` with a matching skill name
3. Observe 500 Internal Server Error

```bash
# Register agent
curl -X POST "$REGISTRY_URL/api/agents" \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"name": "test-agent", "url": "http://example.com", "skills": [{"id": "test_skill", "name": "test_skill"}]}'

# Trigger the crash
curl -X POST "$REGISTRY_URL/api/agents/discover" \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"skills": ["test_skill"]}'
```

## Environment

- Branch: `main` (HEAD at `f2d104c`)
- Python 3.12, Pydantic v2, FastAPI
- Deployed on AWS ECS/Fargate via CloudFormation workshop stack

WPrintz:
Fixes #616

## Summary

- **Streaming AttributeError:** `discover_agents_by_skills` crashed with 500 on every skill-based discovery that found matching agents. `AgentCard` stores streaming in the `capabilities` dict per A2A spec, not as a top-level attribute. The partial fix in `4812e21` covered `list_agents` but missed `discover_agents_by_skills`.
- **Visibility consistency:** Agent visibility used `"internal"` in some places and `"private"` in others. Aligned to `"private"` across Pydantic models, validators, route filters, and federation export.

## Changes

**Streaming fix** (`registry/api/agent_routes.py`):
- Read streaming from `agent.capabilities.get("streaming", False)` instead of `agent.streaming`
- Extract `agent.provider.organization` instead of passing `AgentProvider` object
- Map `request.streaming` into `capabilities` dict during registration and update (was silently dropped by Pydantic)

**Visibility consistency** (4 files):
- `agent_routes.py`: `_filter_agents_by_access` checks `"private"` instead of `"internal"`
- `search_routes.py`: `_user_can_access_agent` checks `"private"` instead of `"internal"`
- `agent_models.py`: Default visibility, description, and validator updated to `"private"`
- `federation_export_routes.py`: Filter both `"internal"` (servers) and `"private"` (agents)

**Tests updated** (5 files):
- All `AgentCard` factory calls updated from `visibility="internal"` to `"private"`
- Federation export tests extended to cover `"private"` filtering

## Test plan

- [x] All affected unit tests pass (278 passed)
- [x] `py_compile` passes on all changed source files
- [x] Verified in production deployment (Grafana error rate at 0%)

### DIFF
--- a/registry/api/agent_routes.py
+++ b/registry/api/agent_routes.py
@@ -290,7 +290,7 @@ def _filter_agents_by_access(
             accessible.append(agent)
             continue
 
-        if agent.visibility == "internal":
+        if agent.visibility == "private":
             if agent.registered_by == username:
                 accessible.append(agent)
             continue
@@ -384,7 +384,7 @@ async def register_agent(
             provider=provider_obj,
             security_schemes=request.security_schemes or {},
             skills=request.skills or [],
-            streaming=request.streaming,
+            capabilities={"streaming": request.streaming} if request.streaming else {},
             tags=tag_list,
             license=request.license,
             visibility=request.visibility,
@@ -909,7 +909,7 @@ async def update_agent(
             provider=request.provider,
             security_schemes=request.security_schemes or {},
             skills=request.skills or [],
-            streaming=request.streaming,
+            capabilities={"streaming": request.streaming} if request.streaming else {},
             tags=tag_list,
             license=request.license,
             visibility=request.visibility,
@@ -1114,6 +1114,12 @@ async def discover_agents_by_skills(
 
         relevance_score = 0.6 * skill_match_score + 0.2 * tag_match_score + 0.2 * trust_boost
 
+        # Extract streaming capability from agent capabilities dict
+        streaming = agent.capabilities.get("streaming", False) if agent.capabilities else False
+
+        # Extract provider organization name (provider is AgentProvider object)
+        provider_name = agent.provider.organization if agent.provider else None
+
         agent_info = AgentInfo(
             name=agent.name,
             description=agent.description,
@@ -1124,8 +1130,8 @@ async def discover_agents_by_skills(
             num_skills=len(agent.skills),
             num_stars=agent.num_stars,
             is_enabled=True,
-            provider=agent.provider,
-            streaming=agent.streaming,
+            provider=provider_name,
+            streaming=streaming,
             trust_level=agent.trust_level,
         )
 
--- a/registry/api/federation_export_routes.py
+++ b/registry/api/federation_export_routes.py
@@ -195,8 +195,8 @@ def _filter_by_visibility(
         # Default to "public" if visibility not specified (backwards compatibility)
         visibility = _get_item_attr(item, "visibility", "public")
 
-        # Never export internal items
-        if visibility == "internal":
+        # Never export internal/private items
+        if visibility in ("internal", "private"):
             continue
 
         # Always export public items
@@ -655,7 +655,7 @@ async def export_security_scans(
         # Check visibility
         visibility = server_data.get("visibility", "public")
 
-        if visibility == "internal":
+        if visibility in ("internal", "private"):
             continue
 
         if visibility == "public":
--- a/registry/api/search_routes.py
+++ b/registry/api/search_routes.py
@@ -303,7 +303,7 @@ async def _user_can_access_agent(agent_path: str, user_context: dict) -> bool:
     if agent_card.visibility == "public":
         return True
 
-    if agent_card.visibility == "internal":
+    if agent_card.visibility == "private":
         return agent_card.registered_by == user_context.get("username")
 
     if agent_card.visibility == "group-restricted":
--- a/registry/schemas/agent_models.py
+++ b/registry/schemas/agent_models.py
@@ -487,8 +487,8 @@ class AgentCard(BaseModel):
 
     # Access control
     visibility: str = Field(
-        "internal",
-        description="public, group-restricted, or internal (default for security)",
+        "private",
+        description="public, group-restricted, or private (default for security)",
     )
     allowed_groups: list[str] = Field(
         default_factory=list,
@@ -554,7 +554,7 @@ class AgentCard(BaseModel):
         v: str,
     ) -> str:
         """Validate visibility value."""
-        valid_values = ["public", "group-restricted", "internal"]
+        valid_values = ["public", "group-restricted", "private"]
         if v not in valid_values:
             raise ValueError(f"Visibility must be one of: {', '.join(valid_values)}")
         return v
@@ -760,8 +760,8 @@ class AgentRegistrationRequest(BaseModel):
         description="License information",
     )
     visibility: str = Field(
-        default="internal",
-        description="Visibility: public, group-restricted, or internal (default)",
+        default="private",
+        description="Visibility: public, group-restricted, or private (default)",
     )
 
     model_config = ConfigDict(populate_by_name=True)
--- a/tests/integration/test_search_integration.py
+++ b/tests/integration/test_search_integration.py
@@ -359,7 +359,7 @@ def search_test_agents() -> list[dict[str, Any]]:
         path="/agents/internal-agent",
         tags=["internal"],
         skills=[],
-        visibility="internal",
+        visibility="private",
         registered_by="testuser",
     )
 
--- a/tests/unit/api/test_agent_routes.py
+++ b/tests/unit/api/test_agent_routes.py
@@ -190,12 +190,12 @@ def sample_agent_card() -> AgentCard:
 
 @pytest.fixture
 def sample_internal_agent_card() -> AgentCard:
-    """Create an internal agent card for testing."""
+    """Create a private agent card for testing."""
     return AgentCardFactory(
         name="internal-agent",
         path="/agents/internal-agent",
         url="http://localhost:9000/internal-agent",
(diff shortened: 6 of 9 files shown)


aarora79:
Thanks for the fix @WPrintz.

Good catch on the streaming `AttributeError` crash in `discover_agents_by_skills` -- this is a real bug that affects every skill-based discovery call.

## Plan: Split This PR Into Two Parts

We are going to fix the visibility/trust_level inconsistency comprehensively in issue #736, which covers the same ground as the visibility portion of this PR but with a full-stack fix (backend defaults, frontend wiring, data backfill script). So we'd like to split this PR:

**This PR (#620):** Keep only the streaming/capabilities fix.
**Issue #736:** Will handle the `"internal"` -> `"public"` visibility change, trust_level defaults, and data migration.

## What We Need From You

### 1. Rebase on main

The branch is 79 commits behind. Please rebase on current main and resolve any conflicts.

### 2. Remove the visibility changes

Please revert all the `"internal"` -> `"private"` changes from this PR. Specifically remove:
- The visibility default changes in `AgentCard` and `AgentRegistrationRequest` models
- The `_filter_agents_by_access` change (`"internal"` -> `"private"`)
- The `_user_can_access_agent` change in `search_routes.py`
- The federation export changes
- The validator change in `_validate_visibility_field`
- All test updates related to visibility renaming

These will be handled properly in #736 with a data migration to avoid breaking existing deployments.

### 3. Keep the streaming/capabilities fixes

These are the core bug fix and should stay:
- `capabilities={"streaming": request.streaming} if request.streaming else {}` in both `register_agent` and `update_agent`
- `streaming = agent.capabilities.get("streaming", False) if agent.capabilities else False` in `discover_agents_by_skills`
- `provider_name = agent.provider.organization if agent.provider else None` in `discover_agents_by_skills`

### 4. Add a backward-compatibility fallback for the streaming read

Agents registered before this fix have `streaming` as a top-level field on the MongoDB document, not inside the `capabilities` dict. After your change, `discover_agents_by_skills` only reads from `capabilities`, so old agents with `streaming=True` would silently lose that flag.

Please update the streaming extraction to fall back to the top-level field for old documents:

```python
# Extract streaming capability -- check capabilities dict first, fall back to
# top-level field for agents registered before the capabilities dict change
streaming = (
    agent.capabilities.get("streaming", False) if agent.capabilities
    else getattr(agent, "streaming", False)
)
```

This one-liner ensures backward compatibility for existing agents while new agents use the `capabilities` dict correctly.

## Summary of Changes to Keep

| File | Change | Keep/Remove |
|------|--------|-------------|
| `registry/api/agent_routes.py` | `capabilities={"streaming": ...}` in register_agent | Keep |
| `registry/api/agent_routes.py` | `capabilities={"streaming": ...}` in update_agent | Keep |
| `registry/api/agent_routes.py` | streaming/provider extraction in discover_agents_by_skills | Keep (with fallback fix) |
| `registry/api/agent_routes.py` | `"internal"` -> `"private"` in _filter_agents_by_access | Remove |
| `registry/api/federation_export_routes.py` | All visibility changes | Remove |
| `registry/api/search_routes.py` | `"internal"` -> `"private"` | Remove |
| `registry/schemas/agent_models.py` | Default/validator visibility changes | Remove |
| All test files | Visibility-related test updates | Remove |

Once you've made these changes, we'll re-review and merge. Thanks again for catching the streaming bug.

aarora79:
@WPrintz https://github.com/agentic-community/mcp-gateway-registry/pull/737 is now merged FYI.

aarora79:
Thanks @WPrintz for the careful root-cause work on this and on #616. Closing as superseded: the bug is fully fixed on `main` and in some places the fix landed on `main` is broader than what this PR proposed.

Quick mapping of where each piece landed:

| This PR's change | On `main` today |
|---|---|
| `discover_agents_by_skills` reads `streaming` from `capabilities` dict | [agent_routes.py:1581-1585](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/api/agent_routes.py#L1581-L1585) (with an added `getattr` fallback for legacy agents) |
| `discover_agents_by_skills` extracts `provider.organization` | [agent_routes.py:1588](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/api/agent_routes.py#L1588) |
| `register_agent` / `update_agent` map `streaming` into `capabilities` | [agent_routes.py:454-459](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/api/agent_routes.py#L454-L459) (merges `request.capabilities` dict with `request.streaming` instead of replacing) |
| `_filter_agents_by_access` checks `"private"` | [agent_routes.py:333](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/api/agent_routes.py#L333) |
| `_user_can_access_agent` checks `"private"` | [search_routes.py:352](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/api/search_routes.py#L352) |
| Federation export drops both `"private"` and `"internal"` | [federation_export_routes.py:212, :673](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/api/federation_export_routes.py#L212) |
| Schemas default to `"private"` and validate accordingly | [agent_models.py:499](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/schemas/agent_models.py#L499) |

One additional piece worth flagging: rather than hardcoding `"private"` everywhere and breaking any caller still sending `"internal"`, `main` added [registry/utils/visibility.py](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/utils/visibility.py) which normalizes the legacy `"internal"` alias to `"private"` at validation time, so existing federation peers and older agents keep working.

Going to close this, but the issue you filed (#616) is what got the full fix landed, so thank you.

PR #716, opened 2026-03-31
feat(charts): add existing secret support to Helm charts

davidgibbons:
## Summary

- Adds `existingSecret` support to all 6 Helm charts — when set, the chart skips creating its managed Secret and the deployment references the pre-existing secret instead
- Adds stack-level existing secret values (`global.existingSharedSecret`, `global.existingOauthProviderSecret`, `global.existingMongoCredentialsSecret`, `mongodb.existingPasswordSecret`) so users with external secret management (ESO, Vault, AWS Secrets Manager) can opt out of Helm-managed secrets entirely
- Adds per-key `{field}ExistingSecret` / `{field}ExistingSecretKey` references for individual sensitive credentials, so a single API key or client secret can be sourced from an existing secret without replacing the whole chart secret

All new values default to `""` — existing behavior is 100% unchanged.

## Covered fields

**auth-server + registry:** `entra.clientSecret`, `okta.clientSecret`, `okta.m2mClientSecret`, `okta.apiToken`, `auth0.clientSecret`, `auth0.m2mClientSecret`, `auth0.managementApiToken`

**registry only:** `ans.apiKey`, `ans.apiSecret`

**mcpgw only:** `app.embeddingsApiKey`

## How per-key references work

Set `okta.clientSecretExistingSecret=my-secret` and the chart:
1. Skips writing `OKTA_CLIENT_SECRET` into its managed secret
2. Adds a `secretKeyRef` env entry in the deployment pointing at `my-secret`

The key within the existing secret defaults to the env var name (e.g. `OKTA_CLIENT_SECRET`) but can be overridden with `okta.clientSecretExistingSecretKey`.

## Test plan

- [ ] Verify default deployment (no existing secret values set) renders identically to pre-change
- [ ] Verify `app.existingSecret` set skips secret creation and references the named secret in `envFrom`
- [ ] Verify `okta.clientSecretExistingSecret` set omits `OKTA_CLIENT_SECRET` from the managed secret and adds the `secretKeyRef` env entry in the deployment
- [ ] Verify stack-level `global.existingSharedSecret` skips `shared-secret.yaml` and uses the named secret in deployments

### DIFF
--- a/charts/auth-server/templates/deployment.yaml
+++ b/charts/auth-server/templates/deployment.yaml
@@ -63,8 +63,6 @@ spec:
             - secretRef:
                 name: {{ .Values.global.existingOauthProviderSecret | default .Values.global.oauthProviderSecretName }}
             {{- end }}
-          {{- $hasPerKeySecret := or .Values.entra.clientSecretExistingSecret .Values.okta.clientSecretExistingSecret .Values.okta.m2mClientSecretExistingSecret .Values.okta.apiTokenExistingSecret .Values.auth0.clientSecretExistingSecret .Values.auth0.m2mClientSecretExistingSecret .Values.auth0.managementApiTokenExistingSecret }}
-          {{- if $hasPerKeySecret }}
           env:
             {{- if .Values.entra.clientSecretExistingSecret }}
             - name: ENTRA_CLIENT_SECRET
@@ -115,7 +113,6 @@ spec:
                   name: {{ .Values.auth0.managementApiTokenExistingSecret }}
                   key: {{ .Values.auth0.managementApiTokenExistingSecretKey }}
             {{- end }}
-          {{- end }}
           livenessProbe:
             httpGet:
               path: /health
--- a/charts/auth-server/templates/secret.yaml
+++ b/charts/auth-server/templates/secret.yaml
@@ -1,3 +1,24 @@
+{{- if and .Values.entra.clientSecret .Values.entra.clientSecretExistingSecret }}
+  {{- fail "Cannot set both entra.clientSecret and entra.clientSecretExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
+{{- if and .Values.okta.clientSecret .Values.okta.clientSecretExistingSecret }}
+  {{- fail "Cannot set both okta.clientSecret and okta.clientSecretExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
+{{- if and .Values.okta.m2mClientSecret .Values.okta.m2mClientSecretExistingSecret }}
+  {{- fail "Cannot set both okta.m2mClientSecret and okta.m2mClientSecretExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
+{{- if and .Values.okta.apiToken .Values.okta.apiTokenExistingSecret }}
+  {{- fail "Cannot set both okta.apiToken and okta.apiTokenExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
+{{- if and .Values.auth0.clientSecret .Values.auth0.clientSecretExistingSecret }}
+  {{- fail "Cannot set both auth0.clientSecret and auth0.clientSecretExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
+{{- if and .Values.auth0.m2mClientSecret .Values.auth0.m2mClientSecretExistingSecret }}
+  {{- fail "Cannot set both auth0.m2mClientSecret and auth0.m2mClientSecretExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
+{{- if and .Values.auth0.managementApiToken .Values.auth0.managementApiTokenExistingSecret }}
+  {{- fail "Cannot set both auth0.managementApiToken and auth0.managementApiTokenExistingSecret — env overrides envFrom and can cause confusing behavior" }}
+{{- end }}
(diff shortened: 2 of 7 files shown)


charts/auth-server/templates/deployment.yaml
```diff
@@ -48,21 +48,74 @@ spec:
...
-                name: {{ .Values.global.sharedSecretName }}
+                name: {{ .Values.global.existingSharedSecret | default .Values.global.sharedSecretName }}
             {{- end }}
             {{- if .Values.global.oauthProviderSecretName }}
             - secretRef:
-                name: {{ .Values.global.oauthProviderSecretName }}
+                name: {{ .Values.global.existingOauthProviderSecret | default .Values.global.oauthProviderSecretName }}
             {{- end }}
+          {{- $hasPerKeySecret := or .Values.entra.clientSecretExistingSecret .Values.okta.clientSecretExistingSecret .Values.okta.m2mClientSecretExistingSecret .Values.okta.apiTokenExistingSecret .Values.auth0.clientSecretExistingSecret .Values.auth0.m2mClientSecretExistingSecret .Values.auth0.managementApiTokenExistingSecret }}
+          {{- if $hasPerKeySecret }}
```
omrishiv: This will remove the entire `env` block if we don't use a `$hasPerKeySecret`, do we need this check?

charts/auth-server/templates/secret.yaml
```diff
@@ -61,20 +62,24 @@ data:
   {{- else if eq $authProviderType "entra" }}
   ENTRA_ENABLED: {{ "true" | b64enc | quote }}
   ENTRA_CLIENT_ID: {{ .Values.entra.clientId | b64enc | quote }}
+  {{- if not .Values.entra.clientSecretExistingSecret }}
```
omrishiv, line 86: we may want a check for all of these (including in the registry) to make sure that the existing secret AND the direct value are not set. As Kubernetes takes `env` over `envFrom`, it could make debugging a bit messy if both are accidentally set

charts/mongodb-configure/templates/secret.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.mongodb.existingSecret }}
```
omrishiv:
```
{{- $existingSecret := .Values.mongodb.existingSecret | default .Values.global.existingMongoCredentialsSecret }}                                                                                                                                                                               
{{- if not $existingSecret }}
```

charts/mongodb-configure/templates/job.yaml
omrishiv, line 2:
```  
{{- $existingSecret := .Values.mongodb.existingSecret | default .Values.global.existingMongoCredentialsSecret }}                                                                                                                                                                               
```

charts/mongodb-configure/templates/job.yaml
```diff
@@ -12,7 +12,7 @@ spec:
           command: ['sh', '-c']
           envFrom:
             - secretRef:
-                name: mongo-credentials
+                name: {{ .Values.mongodb.existingSecret | default "mongo-credentials" }}
```
omrishiv: ```{{ $existingSecret | default "mongo-credentials" }}```

charts/mongodb-configure/templates/job.yaml
```diff
@@ -33,7 +33,7 @@ spec:
           ]
           envFrom:
             - secretRef:
-                name: mongo-credentials
+                name: {{ .Values.mongodb.existingSecret | default "mongo-credentials" }}
```
omrishiv: ```{{ $existingSecret | default "mongo-credentials" }}```

charts/mongodb-configure/values.yaml
```diff
@@ -18,3 +18,4 @@ mongodb:
   username: my-user
   use_tls: false
   storage_backend: mongodb-ce
+  existingSecret: ""  # If set, use this existing secret instead of creating one
```
omrishiv: move under `global`

charts/auth-server/templates/deployment.yaml
davidgibbons: Agreed — removed the `$hasPerKeySecret` guard. The `env:` block is now always present with each entry independently guarded by its own `{{- if }}`.

charts/auth-server/templates/secret.yaml
davidgibbons: Added `fail` validations in auth-server, registry, and mcpgw secret templates. Helm will now error if both a direct value and an existingSecret are set for the same credential.

charts/mongodb-configure/templates/secret.yaml
davidgibbons: Done — using `$existingSecret := .Values.mongodb.existingSecret | default .Values.global.existingMongoCredentialsSecret` now.

charts/mongodb-configure/templates/job.yaml
davidgibbons: Done — `$existingSecret` defined at the top of the file combining both sources.

charts/mongodb-configure/templates/job.yaml
davidgibbons: Updated to `{{ $existingSecret | default "mongo-credentials" }}`.

charts/mongodb-configure/templates/job.yaml
davidgibbons: Updated to `{{ $existingSecret | default "mongo-credentials" }}`.

charts/mongodb-configure/values.yaml
davidgibbons: Moved to `global.existingMongoCredentialsSecret` in values.yaml.

PR #790, opened 2026-04-09
feat: add DISABLE_AI_REGISTRY_TOOLS_SERVER env var (#764)

### LINKED ISSUE #764: Feature request: environment variable to disable built-in demo server auto-registration
## Description

The registry currently auto-registers the `airegistry-tools` (mcpgw) demo server on every startup via `initialize_demo_servers()` in the FastAPI lifespan. This is a nice default for getting started, but in production deployments where we manage our own server registrations, it would be great to have a way to opt out.

## Use Case

We run the registry in a managed environment where all MCP server registrations are controlled through a GitOps pipeline. Having `airegistry-tools` appear automatically is confusing for our users since it's not something we've registered or support — they see it in the server list and wonder where it came from.

## Suggestion

An environment variable like `DISABLE_DEMO_SERVERS=true` (defaulting to `false` to preserve current behavior) that skips the `initialize_demo_servers()` call:

```python
# In registry/main.py lifespan
if not settings.disable_demo_servers:
    from registry.services.demo_servers_init import initialize_demo_servers
    await initialize_demo_servers()
```

This keeps the current experience for new users while giving production deployments a clean opt-out.

## Environment

- Version: v1.0.18

shekharprateek:
## Summary

- Adds `DISABLE_AI_REGISTRY_TOOLS_SERVER` boolean env var to opt out of built-in `airegistry-tools` server auto-registration at startup
- Wired across all deployment paths: Docker Compose, Terraform ECS (root + module variables, task definition env), and Helm charts (registry + umbrella values, ConfigMap template)
- Displays the new parameter in the **Settings -> System Config** admin page under a new "Demo Server Configuration" group
- Follows the existing `MCP_TELEMETRY_DISABLED` pattern for consistency
- Default is `false` (existing behavior preserved); set to `true` for production/GitOps deployments that manage server registrations externally

## Files Changed

| Layer | Files |
|-------|-------|
| Core | `registry/core/config.py`, `registry/main.py` |
| Admin UI Config | `registry/api/config_routes.py` |
| Docker | `docker-compose.yml`, `.env.example` |
| Terraform ECS | `variables.tf`, `terraform.tfvars.example`, `main.tf`, `modules/mcp-gateway/variables.tf`, `modules/mcp-gateway/ecs-services.tf` |
| Helm | `charts/registry/values.yaml`, `charts/registry/templates/configmap-otel.yaml`, `charts/mcp-gateway-registry-stack/values.yaml` |

## Test Plan

- [x] Unit tests pass (3 pre-existing failures unrelated to this change)
- [x] Full Docker Compose stack tested with Keycloak auth
- [x] Flag OFF: `airegistry-tools` server appears in UI
- [x] Flag ON: `airegistry-tools` server absent from UI, log confirms `Demo server auto-registration disabled`
- [x] Flag OFF again: `airegistry-tools` server re-registers on restart
- [x] New config param visible in Settings -> System Config page

Closes #764

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -826,3 +826,7 @@ WORKDAY_TOKEN_URL=https://your-tenant.workday.com/ccx/oauth2/your_instance/token
 
 # Debug mode: log telemetry payloads instead of sending (default: false)
 # TELEMETRY_DEBUG=true
+
+# Disable built-in airegistry-tools server auto-registration
+# Set to true for production/GitOps deployments that manage their own server registrations
+# DISABLE_AI_REGISTRY_TOOLS_SERVER=false
--- a/charts/mcp-gateway-registry-stack/values.yaml
+++ b/charts/mcp-gateway-registry-stack/values.yaml
@@ -218,6 +218,9 @@ registry:
     mcpTelemetryOptIn: false  # Set to true to enable daily heartbeat with aggregate counts
     telemetryDebug: false  # Set to true to log payloads instead of sending
 
+    # Demo server configuration
+    disableAiRegistryToolsServer: false  # Set to true to disable auto-registration of the built-in airegistry-tools server
+
   # ANS (Agent Name Service) Integration
   ans:
     enabled: false  # Enable ANS integration for trust verification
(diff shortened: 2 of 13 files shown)


aarora79:
@shekharprateek add functionality that this new config param is displayed in the settings -> system config page.

shekharprateek:
Added in commit becef91 — the `disable_ai_registry_tools_server` parameter now appears in **Settings -> System Config** under a new "Demo Server Configuration" group (order 16). Follows the same pattern as all other config groups in `config_routes.py`.

PR #850, opened 2026-04-17
update to python 3.14

omrishiv:
*Issue #, if available:*

*Description of changes:*

Bumps the minimum python version to 3.14+. Tested by rebuilding/deploying registry and auth-server and making sure no errors in logs and the UI was still working


aarora79:
have tested on the local docker setup on EC2 and also on ECS, works fine. The failed tests will be fixed immediately afterwards via a separate PR, so merging this now.

PR #864, opened 2026-04-20
feat: IPv6 dual-stack support for server bind addresses

### LINKED ISSUE #863: IPv6 support: server bind addresses prevent deployment on IPv6-only Kubernetes clusters
## Problem

All server components hard-code their bind address to `0.0.0.0` or default to `127.0.0.1`, making it impossible to deploy on **IPv6-only Kubernetes clusters** (e.g. EKS in a hybrid VPC where pods are assigned only IPv6 addresses).

This is increasingly common: AWS regions are running out of IPv4 capacity, and teams are running EKS with IPv6-only pod networking. When a pod has no IPv4 address, a process bound to `0.0.0.0` cannot accept any connections.

## Current state

| Component | Current default | Problem |
|---|---|---|
| `registry/main.py` | `os.getenv("REGISTRY_HOST", "127.0.0.1")` | Binds IPv4 loopback only |
| `docker/registry-entrypoint.sh` | `--host 0.0.0.0` (hardcoded) | IPv4 only, not configurable |
| `docker/auth-entrypoint.sh` | `--host 0.0.0.0` (hardcoded) | IPv4 only, not configurable |
| `servers/currenttime/server.py` | `os.environ.get("HOST", "127.0.0.1")` | Binds IPv4 loopback only |
| `charts/mcpgw/deployment.yaml` | `HOST: 0.0.0.0` (hardcoded) | IPv4 only |

Note: The request handling side (`request_utils.py` using `ipaddress.ip_address()` and `auth_server/server.py` `anonymize_ip()`) already handles IPv6 correctly.

## Proposed fix

Change the default bind address from `0.0.0.0` / `127.0.0.1` to `::` (IPv6 any-address).

On Linux, binding to `::` is **dual-stack by default** -- it accepts both IPv4 and IPv6 connections (via IPv4-mapped IPv6 addresses, controlled by `net.ipv6.bindv6only=0` which is the default on all major distros). This means:

- Existing IPv4-only deployments continue to work with zero changes
- IPv6-only deployments (EKS hybrid VPC, GKE dual-stack, etc.) start working
- Users can still override via `BIND_HOST` / `HOST` environment variables

### Files to change (~10 lines total)

1. **`registry/core/config.py`** -- Add `bind_host: str = "::"` setting
2. **`registry/main.py`** -- Use `settings.bind_host` instead of hardcoded env/default
3. **`docker/registry-entrypoint.sh`** -- `BIND_HOST="${BIND_HOST:-::}"` + use variable in uvicorn
4. **`docker/auth-entrypoint.sh`** -- Same pattern
5. **`servers/currenttime/server.py`** -- Change default from `127.0.0.1` to `::`

### Out of scope for this issue

- Kubernetes Service `ipFamilies`/`ipFamilyPolicy` -- handled by the VPC CNI, not the app
- AWS LB `dualstack` annotations -- deployment-specific, users add via Helm values overrides
- OTEL endpoint bracketed IPv6 literals -- deployment-specific Helm template concern

## Environment

- EKS 1.31+ with VPC CNI in IPv6 mode
- Pods receive only IPv6 addresses (no IPv4)
- uvicorn / FastMCP / FastAPI as the runtime

## References

- [Amazon EKS IPv6 networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-ipv6.html)
- [Python socket dual-stack behavior](https://docs.python.org/3/library/socket.html#socket.AF_INET6) -- binding `::` on Linux accepts IPv4 via mapped addresses
- Related: #856 (ARM64 support) -- similar "works on my laptop but not in production infra" class of issue

soheilkhan:
## Summary

Resolves #863

All server components currently hard-code their bind address to `0.0.0.0` or default to `127.0.0.1`, which prevents deployment on **IPv6-only Kubernetes clusters** (e.g. EKS in a hybrid VPC where pods are assigned only IPv6 addresses).

This PR makes deployment bind addresses configurable and uses `::` (dual-stack) for containerized/managed deployments. On Linux, `::` accepts **both IPv4 and IPv6** connections by default (via IPv4-mapped IPv6 addresses), so existing IPv4-only deployments continue to work without any changes.

## Changes

| File | What changed |
|---|---|
| `registry/core/config.py` | Added `bind_host: str = "::"` setting (configurable via `BIND_HOST` env var) |
| `registry/main.py` | Dev registry entrypoint uses `settings.bind_host` instead of hardcoded `127.0.0.1` |
| `docker/registry-entrypoint.sh` | `BIND_HOST` env var with `::` default, passed to uvicorn |
| `docker/auth-entrypoint.sh` | Same `BIND_HOST` pattern for the auth server |
| `.env.example` | Documents `BIND_HOST` and the Linux dual-stack assumption |
| `charts/mcpgw/values.yaml` | Adds configurable `app.bindHost: "::"` |
| `charts/mcpgw/templates/deployment.yaml` | Reads `HOST` from `app.bindHost` instead of hardcoding it |
| `servers/currenttime/server.py` | Keeps local default as `127.0.0.1`; Docker/Helm/ECS can set `HOST=::` explicitly |
| `terraform/aws-ecs/*` | Adds `bind_host` variable and passes `BIND_HOST`/`HOST` to ECS services |
| `registry/api/config_routes.py` | Exposes `bind_host` in the admin config panel |

## Backward compatibility

- `::` on Linux is dual-stack by default (`net.ipv6.bindv6only=0` on all major distros) -- IPv4 clients continue to connect normally
- If a host/container sets `net.ipv6.bindv6only=1`, IPv4 clients will not reach services bound only to `::`; operators can set `BIND_HOST=0.0.0.0` or `HOST=0.0.0.0` for IPv4-only binding
- All bind addresses remain overridable via environment variables (`BIND_HOST`, `HOST`) or Helm/Terraform values
- No Kubernetes Service spec changes needed (`ipFamilies`/`ipFamilyPolicy` is handled by VPC CNI)
- Request parsing (`ipaddress.ip_address()`) and IP anonymization already handle IPv6

## Test plan

- [x] Python syntax: `python3 -m py_compile registry/core/config.py registry/main.py servers/currenttime/server.py registry/api/config_routes.py`
- [x] Shell syntax: `bash -n docker/registry-entrypoint.sh docker/auth-entrypoint.sh`
- [x] Terraform formatting for touched files: `terraform fmt -check terraform/aws-ecs/variables.tf terraform/aws-ecs/main.tf terraform/aws-ecs/modules/mcp-gateway/variables.tf terraform/aws-ecs/modules/mcp-gateway/ecs-services.tf`
- [x] Helm render: `helm template mcpgw charts/mcpgw --set app.secretKey=test-secret` renders `HOST` as `::`


aarora79:
Thanks for the PR, @soheilkhan! Great work enabling IPv6 dual-stack support. The approach of using `::` for dual-stack is solid, and all 2352 tests pass cleanly.

Requesting the following changes before merge:

### 1. Add `BIND_HOST` to `.env.example`

The new setting needs to be documented for operators. Please add:

```bash
# Network bind address: "::" for dual-stack IPv4+IPv6, "0.0.0.0" for IPv4 only
# On Linux, "::" accepts both IPv4 and IPv6 by default (net.ipv6.bindv6only=0)
BIND_HOST=::
```

### 2. Make Helm chart bind host configurable via `values.yaml`

The bind host is currently hardcoded in the template. It should be configurable so operators can override it without forking the chart.

In `charts/mcpgw/values.yaml`, add:

```yaml
app:
  bindHost: "::"
```

In `charts/mcpgw/templates/deployment.yaml`, change:

```yaml
# Before
- name: HOST
  value: "::"

# After
- name: HOST
  value: {{ .Values.app.bindHost | quote }}
```

### 3. Revert `currenttime/server.py` default to `127.0.0.1`

The default changed from `127.0.0.1` (localhost-only) to `::` (all interfaces). This loosens the security posture for local (non-Docker) development. Please keep `127.0.0.1` as the default here and rely on `HOST=::` being set explicitly in Docker/Helm environments, which is already the case via the Helm template and Dockerfile.

### 4. Add Terraform ECS variable for `BIND_HOST`

Per project standards, new config params should be propagated to Terraform ECS (`variables.tf` + `ecs.tf`).

### 5. Add to Config API for admin visibility

The `bind_host` setting should be added to `CONFIG_GROUPS` in `registry/api/config_routes.py` so it's visible in the admin config panel.

### 6. Document `net.ipv6.bindv6only` assumption

The dual-stack behavior depends on the Linux default `net.ipv6.bindv6only=0`. If a host or container has `net.ipv6.bindv6only=1`, IPv4 clients will fail to connect when binding to `::`. Please add a one-liner about this in the PR description or a deployment doc note.

---

Happy to re-review once updated.

soheilkhan:
Thanks for the review. Updated in commit `189133c8`.

Checklist:

- Added `BIND_HOST` to `.env.example` with the Linux dual-stack note.
- Made the `mcpgw` Helm chart configurable via `app.bindHost` and changed the deployment template to use it.
- Reverted `servers/currenttime/server.py` back to the local-safe `127.0.0.1` default; managed deployments now set `HOST` explicitly.
- Added Terraform ECS `bind_host` plumbing and pass `BIND_HOST` / `HOST` into the ECS services.
- Added `bind_host` to `CONFIG_GROUPS` for admin config visibility.
- Updated the PR description with the `net.ipv6.bindv6only=0` assumption and override guidance.

Validation run:

- `python3 -m py_compile registry/core/config.py registry/main.py servers/currenttime/server.py registry/api/config_routes.py`
- `bash -n docker/registry-entrypoint.sh docker/auth-entrypoint.sh`
- `terraform fmt -check` on the touched Terraform files
- `helm template mcpgw charts/mcpgw --set app.secretKey=test-secret` confirms `HOST` renders as `::`

Note: `terraform fmt -check terraform/aws-ecs` still reports a pre-existing unrelated formatting issue in `terraform/aws-ecs/keycloak-database.tf`, so I limited formatting/checks to the Terraform files touched by this PR to avoid unrelated churn.

danielfree:
I'm also interested to test this feature if its been merged

soheilkhan:
@aarora79 Do you want to review yourself?

aarora79:
Heads-up: merged and then hit a regression -- the `::` default broke the stock Docker deployment (nginx has ~20 hardcoded `proxy_pass http://127.0.0.1:7860` directives; the container's uvicorn bound on `[::]:7860` only, so nginx couldn't reach upstream and every request 404'd).

Filed follow-up #1016 flipping the default back to `0.0.0.0` across every surface (entrypoints, Helm, Terraform, settings, `.env.example`) and added the missing `docs/unified-parameter-reference.md` row. IPv6-only operators now opt in explicitly with `BIND_HOST=::`. All your plumbing is intact — only the default changed. No action needed from you.

PR #873, opened 2026-04-22
feat: add external OIDC issuer support for federation / cross-IdP trust

alejandronunezcabello:
## Summary

Adds the ability to validate JWT tokens from external OIDC-compliant identity providers alongside Keycloak tokens. This enables **federation scenarios** where a parent organization's IdP issues tokens that need to be validated by the registry's auth-server.

## Motivation

In enterprise deployments, a registry may need to accept tokens from multiple identity providers. For example:
- A parent company's IdP issues Bearer tokens containing agent roles
- A partner organization needs programmatic access with their own IdP
- Multi-cloud deployments with different IdPs per environment

Currently the Keycloak provider only validates tokens from its own Keycloak instance (plus self-signed tokens). This PR adds a generic external issuer capability.

## Configuration

Two new environment variables:

| Variable | Default | Description |
|----------|---------|-------------|
| `EXTERNAL_ISSUERS` | (empty) | Comma-separated list of trusted OIDC issuer URLs |
| `EXTERNAL_GROUPS_CLAIM` | `roles` | Token claim to map to registry groups for RBAC |

Example:
```bash
EXTERNAL_ISSUERS=https://login.microsoftonline.com/{tenant}/v2.0,https://accounts.google.com
EXTERNAL_GROUPS_CLAIM=roles
```

## How it works

1. Token arrives at auth-server's `validate_token()`
2. If the token's `iss` claim matches a configured external issuer, **or** the `kid` is not found in Keycloak's JWKS, the external issuer path is tried
3. JWKS is fetched via standard OIDC Discovery (`.well-known/openid-configuration` → `jwks_uri`)
4. Token signature is verified against the external issuer's public keys
5. The configured claim (e.g., `roles`) is mapped to registry groups for RBAC
6. External JWKS responses are cached with the same TTL as Keycloak (1 hour)

## Backward compatibility

When `EXTERNAL_ISSUERS` is empty (the default), behavior is **identical** to the current implementation. No existing functionality is affected.

## Changes

- `auth_server/providers/keycloak.py`: Added `EXTERNAL_ISSUERS`, `EXTERNAL_GROUPS_CLAIM` config, per-issuer JWKS cache, `_get_external_jwks()`, `_validate_external_token()`, `_validate_external_token_by_kid()` methods, and updated `validate_token()` flow
- `tests/auth_server/unit/providers/test_keycloak.py`: Added `TestExternalIssuerSupport` test class with 6 tests covering discovery, caching, role mapping, multi-issuer fallback, and error cases

## Testing

- [x] Python syntax validated
- [x] 6 new unit tests added
- [x] Existing tests unmodified (backward compatible)

### DIFF
--- a/tests/auth_server/unit/providers/test_keycloak.py
+++ b/tests/auth_server/unit/providers/test_keycloak.py
@@ -804,3 +804,180 @@ class TestKeycloakProviderInfo:
         assert is_healthy is True
         mock_get.assert_called_once()
         assert "/health/ready" in mock_get.call_args[0][0]
+
+
+# =============================================================================
+# EXTERNAL ISSUER SUPPORT TESTS
+# =============================================================================
+
+
+class TestExternalIssuerSupport:
+    """Tests for external OIDC issuer validation (federation / cross-IdP trust)."""
+
+    def _make_provider(self):
+        """Create a KeycloakProvider for testing."""
+        from auth_server.providers.keycloak import KeycloakProvider
+
+        return KeycloakProvider(
+            keycloak_url="http://keycloak:8080",
+            realm="test-realm",
+            client_id="test-client",
+            client_secret="test-secret",
+        )
+
+    @patch("auth_server.providers.keycloak.EXTERNAL_ISSUERS", [])
+    def test_no_external_issuers_configured(self):
+        """When EXTERNAL_ISSUERS is empty, external path is never tried."""
+        provider = self._make_provider()
+
+        # A token with an unknown kid should raise without trying external
+        with patch.object(provider, "get_jwks", return_value={"keys": []}):
+            with pytest.raises(ValueError, match="No matching key"):
+                provider.validate_token("fake.jwt.token")
+
+    @patch("auth_server.providers.keycloak.EXTERNAL_ISSUERS", ["https://external-idp.example.com"])
+    @patch("requests.get")
+    def test_get_external_jwks_discovery(self, mock_get):
+        """Test OIDC discovery + JWKS fetch for external issuer."""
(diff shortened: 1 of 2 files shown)


tests/auth_server/unit/providers/test_keycloak.py
```diff
@@ -805,3 +804,180 @@ def test_check_keycloak_health(self, mock_get):
...
+
+    @patch("auth_server.providers.keycloak.EXTERNAL_ISSUERS", [])
+    def test_no_external_issuers_configured(self):
+        """When EXTERNAL_ISSUERS is empty, external path is never tried."""
+        provider = self._make_provider()
+
+        # A token with an unknown kid should raise without trying external
+        with patch.object(provider, "get_jwks", return_value={"keys": []}):
+            with pytest.raises(ValueError, match="No matching key"):
+                provider.validate_token("fake.jwt.token")
```
alejandronunezcabello:
```suggestion
            with pytest.raises(ValueError, match="Invalid token"):
                provider.validate_token("fake.jwt.token")
```

aarora79:
thank you for contributing this PR @alejandronunezcabello , so you have tested this with Keycloak and Entra?

alejandronunezcabello:
> thank you for contributing this PR @alejandronunezcabello , so you have tested this with Keycloak and Entra?

Hi @aarora79, sorry for the delayed response. We have not tested with Entra. We were going to test with a custom external IdP but in the end we have changed plans so no actual test has been performed. Feel free to close the PR if the complexity of the contribution is not worth the benefit of the feature.

PR #920, opened 2026-05-02
fix(version): inject BUILD_VERSION from Helm image tag to prevent envFrom override

### LINKED ISSUE #919: UI shows version 1.0.0 instead of actual release version on EKS/Helm deployments
## Problem

Customers deploying on EKS via Helm charts report the UI displays version **1.0.0** instead of the actual release version (e.g., 1.0.21), even though they are pulling the correct versioned image from ECR.

## Root Cause

The Helm deployment templates load multiple K8s secrets via `envFrom`. If any of those secrets contains a `BUILD_VERSION` key (even as an empty string), it overrides the value baked into the Docker image at build time.

In Python, `os.getenv("BUILD_VERSION")` returns `""` which is falsy, so `registry/version.py` falls through to git (not installed in the container), then to `DEFAULT_VERSION = "1.0.0"`.

### Version resolution chain in `registry/version.py`:

1. `BUILD_VERSION` env var (set at Docker build time via `ENV BUILD_VERSION=$BUILD_VERSION`)
2. `git describe --tags` (git not available in container)
3. `DEFAULT_VERSION = "1.0.0"` (fallback)

### Additional affected paths:

- **docker-compose.prebuilt.yml**: Had `BUILD_VERSION=${BUILD_VERSION:-1.0.0}` as a runtime env var, which stomps the value baked into the pulled image if the customer does not set it in their `.env`
- **scripts/publish_containers.sh**: Was not passing `--build-arg BUILD_VERSION` during `docker build`, so DockerHub images always got `1.0.0`

## Fix

1. **Helm charts**: Add explicit `BUILD_VERSION` env var in all three deployment templates (registry, auth-server, mcpgw) sourced from `global.image.tag`. Explicit `env` entries take precedence over `envFrom`, so the version always matches the deployed tag.
2. **docker-compose.prebuilt.yml**: Source `BUILD_VERSION` from `REGISTRY_VERSION` (the same variable that controls the image tag being pulled).
3. **publish_containers.sh**: Pass `--build-arg BUILD_VERSION="$VERSION"` to `docker build`.

## Affected Files

- `charts/registry/templates/deployment.yaml`
- `charts/auth-server/templates/deployment.yaml`
- `charts/mcpgw/templates/deployment.yaml`
- `docker-compose.prebuilt.yml`
- `scripts/publish_containers.sh`

aarora79:
## Summary

- Inject explicit `BUILD_VERSION` env var in all three Helm deployment templates (registry, auth-server, mcpgw) sourced from `global.image.tag`, preventing envFrom secrets from overriding the baked-in image version
- Fix docker-compose.prebuilt.yml runtime env default from hardcoded `1.0.0` to `REGISTRY_VERSION` (the image tag being pulled)
- Add missing `--build-arg BUILD_VERSION` to publish_containers.sh so DockerHub images get the correct version

Fixes #919

## Test plan

- [ ] Deploy on EKS with Helm chart, verify `/api/version` returns the correct tag (e.g., `1.0.21`)
- [ ] Verify the version displayed in the UI footer matches the deployed image tag
- [ ] Test docker-compose.prebuilt.yml: pull a versioned image and confirm version endpoint returns the correct value
- [ ] Verify that customers who previously saw `1.0.0` now see the correct version after `helm upgrade`


aarora79:
Superseded -- closing this in favor of the merged follow-ups that addressed the same root cause:

- #958 (`fix(docker): pass BUILD_VERSION when publishing to Docker Hub`) -- covers the `publish_containers.sh` `--build-arg` fix.
- #965 (`fix overriding the image tag from helm variables`) -- covers the Helm `BUILD_VERSION` injection from `global.image.tag`.

The `docker-compose.prebuilt.yml` default change from this PR is no longer needed since the image itself now carries the correct `BUILD_VERSION` at build time.

PR #925, opened 2026-05-03
docs: add security preflight admission pattern

### LINKED ISSUE #911: Docs idea: security preflight before MCP registry admission / first connection
Hi — this project is tackling exactly the problem many MCP adopters hit next: once discovery becomes easy, admission and first-connection trust become the bottleneck.

Would a short docs/example section for **security preflight before MCP registry admission or first connection** fit here?

Suggested pattern:

1. Discover or propose a remote MCP server.
2. Run an external trust preflight before admitting it to the gateway registry.
3. Store the preflight result with the admission/audit record.
4. Allow automatic connection only when policy passes; otherwise require human review.

Concrete x402 example using ShieldAPI:

```bash
agentcash fetch 'https://shield.vainplex.dev/api/cdp/check/mcp/trust?shieldapi_source=mcp-gateway-registry' \
  -m POST \
  -H 'content-type: application/json' \
  -b '{"endpoint":"https://mcp.example.com"}'

agentcash fetch 'https://shield.vainplex.dev/api/cdp/full/scan?target=mcp.example.com&shieldapi_source=mcp-gateway-registry'
```

The goal is not to replace gateway governance; it is to add one cheap external signal before a discovered server becomes an executable dependency for agents.

If useful, I can turn this into a small vendor-light docs PR and keep ShieldAPI as one concrete example rather than a marketing placement.

zerone0x:
## Summary

- Adds an external trust preflight pattern to the Registration Gate documentation
- Frames third-party trust checks as one admission signal, not a replacement for registry governance
- Adds the registration gate guide to the MkDocs navigation and docs README index

Closes #911

## Validation

- `uv run --extra docs mkdocs build --strict` (blocked: dependency installation ran out of disk space while extracting `torch`)
- `python3` markdown sanity check for balanced fences and MkDocs/docs references

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,7 @@ mkdocs build
 - `complete-setup-guide.md` - Step-by-step setup from scratch
 - `installation.md` - Complete installation guide
 - `auth.md` - Authentication and OAuth setup
+- `registration-webhooks.md` - Registration notifications and admission-control gate integrations
 - `cognito.md` - Amazon Cognito configuration
 - `keycloak-integration.md` - Keycloak integration guide
 - `scopes.md` - Access control and permissions
--- a/docs/registration-webhooks.md
+++ b/docs/registration-webhooks.md
@@ -304,6 +304,35 @@ The gate is checked on the following operations:
 | Skill | Register | `POST /api/skills` |
 | Skill | Update | `PUT /api/skills/{path}` |
 
+### Pattern: External Trust Preflight Before Admission
+
+The registration gate can be used as a security preflight step before a newly discovered MCP server, A2A agent, or skill becomes an executable dependency for users and AI agents. This complements the built-in security scanner: the registry still owns admission, access control, and auditability, while the gate can call external risk services for additional signals.
+
+A typical server admission flow is:
+
+1. Discover or propose a remote MCP server for registration.
+2. Send the sanitized registration payload to the registration gate before persistence.
+3. Have the gate evaluate policy, using internal rules and optional external trust checks.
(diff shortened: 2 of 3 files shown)


aarora79:
Thanks for the contribution, and for framing this as an admission-gate pattern rather than a replacement for registry governance. That framing is on point, and the surrounding guidance (fail-closed, allow/deny lists, human review) reads well.

That said, we can't merge this as written. A couple of concerns:

1. **No third-party endorsements in our docs.** The registration gate is intentionally a pluggable extension point, and we're happy to document the *pattern* of calling an external trust service from the gate. But we can't ship concrete examples that point at a specific commercial product, in this case `shield.vainplex.dev` / ShieldAPI. Readers reasonably treat example URLs in official docs as a recommendation, and we're not in a position to vet or endorse any particular preflight vendor. The `shieldapi_source=mcp-gateway-registry` query parameter also reads as attribution/referral tracking, which reinforces the concern.

2. **Unexplained tooling.** The example uses `agentcash fetch ...` with no context on what it is or how to install it, which will confuse most readers landing on this page.

If you'd like to update the PR so the example uses a generic placeholder (e.g. `https://trust-preflight.example.com/api/check` with a plain `curl` invocation, and no vendor name, product name, or source tag anywhere in the doc), we're glad to take another look. The narrative guidance around the example is fine; it's only the concrete endpoint and tool references that need to go.

If reworking it that way doesn't fit what you're trying to demonstrate, we'll close this out. Either is fine, just let us know which direction you'd like to take.

Thanks.

Issue #500, opened 2026-02-17
Registry: UI Logout function does not redirect correctly when using `routingMode = path`


snorlaX-sleeps:
Related issue: https://github.com/agentic-community/mcp-gateway-registry/issues/390
PR adding path based routing to the Helm Chart: https://github.com/agentic-community/mcp-gateway-registry/pull/395

## Issue
When the routing mode is set to `path` the registry logout function uses the `host` header to determine the `redirect_uri` that is passed to both the `auth-server` and `keycloak`.
This should be updated to include the `root_path` when it is set
Registry Logout function using the `host` header: https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/auth/routes.py#L238
Both the `auth-server` and `keycloak` use the `redirect_uri` from the registry.

### Example logs from logout attempt
Removing the host but leaving the `root_path` in the logs

Registry
```
registry-7c5bc5b86c-vzd4g registry 2026-02-17 19:56:46,558,p114,{routes.py:223},INFO,User was authenticated via OAuth2 provider: keycloak
registry-7c5bc5b86c-vzd4g registry 2026-02-17 19:56:46,558,p114,{routes.py:254},INFO,Redirecting to keycloak logout: https://<host>/auth-server/oauth2/logout/keycloak?redirect_uri=https://<host>/logout
registry-7c5bc5b86c-vzd4g registry 2026-02-17 19:56:46,558,p114,{routes.py:258},INFO,User logged out.
```
Auth Server
```
auth-server-66f6c4f5c7-z7gms auth-server 2026-02-17 19:09:56,496,p1,{server.py:2781},INFO,Redirecting to keycloak logout: https://<host>/keycloak/realms/mcp-gateway/protocol/openid-connect/logout?client_id=mcp-gateway-web&post_logout_redirect_uri=https%3A%2F%2F<host>%2Flogout
auth-server-66f6c4f5c7-z7gms auth-server INFO:     10.0.38.254:47363 - "GET /auth-server/oauth2/logout/keycloak?redirect_uri=https://<host>/logout HTTP/1.1" 302 Found
```
KeyCloak
```
[...]

PR #931, opened 2026-05-05
fix: multiple path-routing, logout, and OAuth redirect bugs on Helm deployments

omrishiv:
## Summary

While verifying #500 locally on a Helm path-routing deployment (routingMode=path, apex-domain host), several pre-existing bugs surfaced that together broke login, logout, and SPA asset loading. This PR fixes them as one coherent set.

### Login / OAuth

- auth_server/server.py:2676 — oauth2_login was concatenating ROOT_PATH onto AUTH_SERVER_EXTERNAL_URL even though the Helm chart already bakes the path prefix into the external URL. Result under path routing:
callback_uri=https://host/auth-server/auth-server/oauth2/callback/.... Removed the double-append.
- auth_server/server.py:~3045 — oauth2_callback's open-redirect check used `hostname.endswith(cookie_domain)`. When `SESSION_COOKIE_DOMAIN=".example.com"`, the apex host `example.com` doesn't satisfy `.endswith(".example.com")`, so legitimate redirects to the apex were rejected and silently replaced with `/`. This manifested in path mode where the apex is the only host. Now accepts exact apex match OR subdomain match, with lowercase normalization. Extracted to a reusable helper `_is_redirect_within_cookie_domain()`.

### Logout

- registry/auth/routes.py — `delete_cookie()` was called without path, domain, or matching attributes. The auth-server sets the session cookie with `Domain=.{domain}; Path=/; Secure`, and browsers silently ignore delete headers that don't match the original `(name, domain, path)`. Result: clicking "Sign Out" returned to `/login` but left the session cookie intact — users could navigate back to `/registry/` and remain logged in. Now passes matching `path="/"` and `domain=settings.session_cookie_domain`. `secure` is intentionally omitted so the delete isn't rejected on any request transported over HTTP (an empty expired cookie has no secrets to protect).
- charts/registry/templates/secret.yaml — added `SESSION_COOKIE_DOMAIN` and `SESSION_COOKIE_SECURE` env vars mirroring what the auth-server chart sets, so the registry pod has the values it needs to delete the cookie it doesn't create.
- charts/keycloak-configure/templates/configmap.yaml — the mcp-gateway-web Keycloak client was missing the `post.logout.redirect.uris` attribute. Keycloak ≥25 requires this; earlier versions fell back to redirectUris. Without it, Keycloak silently rejects or ignores `post_logout_redirect_uri` and the SSO session survives. Added `${REGISTRY_URL}/*` to the allow-list.
- charts/keycloak-configure/templates/configmap.yaml — client creation was not idempotent: POST /clients returns 409 on existing deployments and the script swallowed it, so attribute additions would never apply on upgrades. Added an `upsert_client()` helper that GETs by clientId, POSTs if missing, and PUTs the full representation otherwise.
- auth_server/server.py oauth2_logout : reject absolute `redirect_uri` values outside `SESSION_COOKIE_DOMAIN` before forwarding to the IdP, using the shared helper. This complements the Keycloak-side allow-list and doesn't rely on IdP-client configuration being correct.
- auth_server/server.py:3100 — `delete_cookie("oauth2_temp_session")` now passes `path="/"` to match the `set_cookie` that created it.

### Helm chart

- charts/keycloak-configure/templates/secret.yaml:31 — fixed missing `REGISTRY_URL` variable for the Keycloak setup script.

### SPA under path routing

- registry/main.py:_build_cached_index_html — when `ROOT_PATH` is set, now rewrites ="/static/ and ="/favicon.ico" references in the served index.html to include the prefix, in addition to the existing `<base href>` injection. `<base href>` only affects relative URLs; the CRA build (with homepage:"/") emits absolute asset URLs that bypass the ingress's `/registry/*` rule and 404 at the apex. Verified no false-positive replacements because the =/ anchor is specific to attribute syntax.
- frontend/src/utils/basePath.ts (new) — consolidates `getBaseURL`, `getBasename`, `apiUrl` into a single source of truth. App.tsx and AuthContext.tsx now import from it, dropping their local duplicates.
- frontend/src/components/Layout.tsx, frontend/src/components/UptimeDisplay.tsx — three bare `fetch('/api/...')` calls were bypassing `axios.defaults.baseURL` and hitting the apex root. Now use `apiUrl(...)` to include `ROOT_PATH`. Fixes 404s on /api/stats, /api/version, /api/search/tags.

## Scope

All changes are no-ops in subdomain mode and local development (ROOT_PATH="", no <base> tag, SESSION_COOKIE_DOMAIN as before). Impact is limited to routingMode=path deployments plus the cookie-deletion fix which applies to any deployment where the session cookie has a Domain= attribute.

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -452,6 +452,27 @@ def mask_token(token: str) -> str:
     return "***MASKED***"
 
 
+def _is_redirect_within_cookie_domain(
+    url: str,
+    cookie_domain: str,
+) -> bool:
+    """Check if an absolute redirect URL's host is within SESSION_COOKIE_DOMAIN.
+
+    SESSION_COOKIE_DOMAIN (e.g. ".example.com") defines the trust boundary —
+    the apex host and any subdomain sharing the session cookie are safe
+    redirect targets. The leading dot is conventional but not required.
+    """
+    if not cookie_domain:
+        return False
+    parsed = urlparse(url)
+    if parsed.scheme not in ("http", "https"):
+        return False
+    # urlparse().hostname is lowercased by Python; normalize apex to match.
+    hostname = parsed.hostname or ""
+    apex = cookie_domain.lstrip(".").lower()
+    return hostname == apex or hostname.endswith(f".{apex}")
+
+
 def _is_safe_redirect_url(
     url: str,
     allowed_hosts: set[str] | None = None,
@@ -2678,7 +2699,10 @@ async def oauth2_login(provider: str, request: Request, redirect_uri: str = None
         # The callback_uri MUST match exactly between authorization and token exchange
         auth_server_external_url = os.environ.get("AUTH_SERVER_EXTERNAL_URL", "").rstrip("/")
         if auth_server_external_url:
-            auth_server_url = f"{auth_server_external_url}{ROOT_PATH}"
+            # AUTH_SERVER_EXTERNAL_URL is the complete public base URL and
+            # already includes any path prefix (e.g. "/auth-server" in path
+            # routing mode); ROOT_PATH is only used in the host-fallback branch.
+            auth_server_url = auth_server_external_url
             scheme = "https" if auth_server_external_url.startswith("https") else "http"
             logger.info(f"OAuth2 login - using AUTH_SERVER_EXTERNAL_URL: {auth_server_url}")
         else:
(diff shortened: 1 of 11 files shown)


aarora79:
**Post-merge regression report: `oauth2_logout` redirect_uri guard rejects documented default configuration**

The new redirect_uri guard in `oauth2_logout` at [auth_server/server.py:3228](https://github.com/agentic-community/mcp-gateway-registry/blob/main/auth_server/server.py#L3228) relies on `_is_redirect_within_cookie_domain()`, which returns `False` unconditionally when `SESSION_COOKIE_DOMAIN` is empty. Per `.env.example:629-643`, an empty `SESSION_COOKIE_DOMAIN` is the **recommended** default for single-domain deployments ("cookie scoped to exact host only - safest option"), so every deployment using that default now sees its own legitimate SPA logout URL (e.g. `https://<host>/logout`) blocked:

```
WARNING: Blocked unsafe logout redirect_uri for keycloak: https://mcpgateway.ddns.net/logout
```

End-user symptom: Keycloak strips the blocked `post_logout_redirect_uri` and falls back to its own default page, so users bypass the "Successfully Logged Out" 5-second countdown page and land straight on the login screen.

`oauth2_callback` has the same issue at [auth_server/server.py:3064](https://github.com/agentic-community/mcp-gateway-registry/blob/main/auth_server/server.py#L3064), but it happens to be masked because `OAUTH2_CONFIG.registry.success_redirect` is typically a relative path, which short-circuits the guard.

Reproduction (single-domain docker-compose, `SESSION_COOKIE_DOMAIN=` empty):
1. Log in via Keycloak.
2. Click "Sign Out" in the UI.
3. Auth-server logs the "Blocked unsafe logout redirect_uri" warning; browser lands on `/login` instead of `/logout`.

Fix (teaches the helper that same-origin is always safe, independent of `SESSION_COOKIE_DOMAIN`): [commit a6551a06 on PR #926](https://github.com/agentic-community/mcp-gateway-registry/pull/926/commits/a6551a06). Verified locally:

- Same-origin `https://mcpgateway.ddns.net/logout` now passes -> 5-second page renders.
- Cross-origin `https://evil.example.com/steal` is still blocked; `redirect_uri` stripped before forwarding to Keycloak.
- 8-case unit sanity check: same-origin empty cookie_domain, cross-origin empty cookie_domain, subdomain match, apex match, outside cookie domain, TLS-terminated upstream via `x-forwarded-*`, scheme mismatch, non-http scheme -- all behave as expected.

PR #932, opened 2026-05-05
fix: handle null description in virtual server documents

alejandronunezcabello:
## Problem

Virtual MCP servers disappear from the UI after editing. The root cause is a Pydantic validation error when reading documents from MongoDB:

```
Failed to parse virtual server document: 1 validation error for VirtualServerConfig
description
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

### Root Cause

1. The frontend edit form sends `description: null` when the description field is empty (line 207 of VirtualServerForm.tsx)
2. The update service stores this `null` in MongoDB via `$set`
3. On subsequent reads, `VirtualServerConfig` rejects `None` for the `str`-typed `description` field
4. The repository silently skips unparseable documents → virtual servers vanish from listings
5. The update endpoint also fails with 500 because it reads the document first

### Fixes

- **Backend model** (`virtual_server_models.py`): Add `field_validator` on `VirtualServerConfig` and `VirtualServerInfo` that coerces `None` → `""` (defensive read path)
- **Backend service** (`virtual_server_service.py`): Coerce `None` description to empty string before writing to MongoDB (defensive write path)
- **Frontend** (`VirtualServerForm.tsx`): Send `''` instead of `null` for empty description in edit mode
- **Repository** (`virtual_server_repository.py`): Include document `_id` in parse error logs for diagnosability

### Testing

- Verified via `mongosh` that existing documents with `description: null` are now parseable
- Pydantic validator handles all cases: `None` → `""`, missing → `""`, valid string → kept as-is

### DIFF
--- a/frontend/src/components/VirtualServerForm.tsx
+++ b/frontend/src/components/VirtualServerForm.tsx
@@ -204,7 +204,7 @@ const VirtualServerForm: React.FC<VirtualServerFormProps> = ({
       if (isEditMode) {
         const updateData: UpdateVirtualServerRequest = {
           server_name: name.trim(),
-          description: description.trim() || null,
+          description: description.trim() || '',
           tool_mappings: allMappings,
           required_scopes: parsedScopes,
           tags: parsedTags,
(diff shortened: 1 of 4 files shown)


aarora79:
Thanks @alejandronunezcabello for the fix -- shipped in #932 and merged. I opened a small follow-up in #940 to remove an unused `model_validator` import (ruff F401 that slipped through because our CI runs ruff with `continue-on-error: true`), apply `ruff format` to the repository file, and add four unit tests that pin the null-description coercion on both the read and write paths.

PR #942, opened 2026-05-06
fix: deduplicate IAM user list and handle orphaned M2M deletes

alejandronunezcabello:
## Problem

1. **Duplicate users in IAM list**: M2M service accounts appear twice — once from the IdP listing (Keycloak already includes service account clients) and once from MongoDB's `idp_m2m_clients` collection.

2. **Cannot delete orphaned M2M accounts**: If a record exists in MongoDB but was deleted from the IdP (or never properly synced), the delete endpoint fails with "User or M2M account 'X' not found" because it only checks the IdP.

## Root Cause

- `management_list_users` appends ALL MongoDB `idp_m2m_clients` entries without checking if they already appear in the IdP results
- `management_delete_user` only calls `iam.delete_user()` (IdP-only) and never cleans up MongoDB

### List endpoint (`GET /api/management/iam/users`)
- Build a set of existing usernames from the IdP results
- Skip MongoDB M2M entries whose `client_id` or `name` already appears in that set

### Delete endpoint (`DELETE /api/management/iam/users/{username}`)
- After attempting IdP deletion, also remove from MongoDB `idp_m2m_clients`
- If user is not found in the IdP, don't immediately fail — check MongoDB too
- Only return 400 if user was found in neither store

### DIFF
--- a/registry/api/management_routes.py
+++ b/registry/api/management_routes.py
@@ -152,22 +152,31 @@ async def management_list_users(
         cursor = collection.find({})
         m2m_docs = await cursor.to_list(length=None)
 
-        # Add M2M clients as users with special email pattern
+        # Deduplicate: skip MongoDB entries whose client_id already appears
+        # in the IdP results (e.g. Keycloak already lists M2M service accounts)
+        existing_usernames = {u.get("username", "").lower() for u in raw_users}
+
+        # Add only M2M clients that are NOT already in the IdP listing
+        added = 0
         for doc in m2m_docs:
             client_id = doc.get("client_id", "")
+            name = doc.get("name", client_id)
+            if name.lower() in existing_usernames or client_id.lower() in existing_usernames:
+                continue
             raw_users.append(
                 {
                     "id": client_id,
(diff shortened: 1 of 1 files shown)


aarora79:
Follow-up PR #1005 (merged) shipped the non-blocking post-merge items:

- **P1.1** unit tests for dedup + orphan-delete branches
- **P1.2** per-session isolated test MongoDB database name (fixes `TestManagementListUsers` flakiness against local Mongo)
- **P1.3** case-insensitive regex on the MongoDB delete filter (parity with the case-insensitive dedup on list)
- **P2.1** audit event on MongoDB-only delete with `idp_skip_reason=not_found`
- **P2.2** `m2m_orphan_cleanups_total` Prometheus counter
- **P2.3** enriched MongoDB-cleanup warning log with username + `idp_deleted`
- **P2.4** hoisted `get_documentdb_client` import to module scope

Verified end-to-end on local Compose: case-insensitive delete, dedup, audit event persisted with the expected fields, truly-missing still -> 400.

PR #958, opened 2026-05-07
fix(docker): pass BUILD_VERSION when publishing to Docker Hub (#919)

### LINKED ISSUE #919: UI shows version 1.0.0 instead of actual release version on EKS/Helm deployments

aarora79:
Closes #919 (milestone v1.0.22-p1).

## Summary

- Customers pulling from Docker Hub see \`BUILD_VERSION=1.0.0\` inside the registry container even when they pulled \`mcpgateway/registry:v1.0.22\`. The registry UI and telemetry then report the wrong version.
- Root cause: [scripts/publish_containers.sh](scripts/publish_containers.sh) calls \`docker build\` without \`--build-arg BUILD_VERSION\`, so the Dockerfile ARG default (\`1.0.0\`) is baked into the image. ECR images are unaffected because the GitHub Actions workflow passes the arg directly.
- Fix: pass \`--build-arg BUILD_VERSION=\"\$VERSION\"\` from the Docker Hub publish script, and add the standard \`ARG BUILD_VERSION / ENV BUILD_VERSION\` pair to the three Dockerfiles that were missing it so every component image carries its real version.

## Evidence (from the customer)

\`\`\`
$ docker inspect mcpgateway/registry:v1.0.22 | grep BUILD_VERSION
                    \"BUILD_VERSION=1.0.0\"        # wrong

$ docker inspect public.ecr.aws/ai-registry/registry | grep BUILD_VERSION
                    \"BUILD_VERSION=v1.0.22\"     # correct
\`\`\`

## Files changed

| File | Change |
|------|--------|
| [scripts/publish_containers.sh](scripts/publish_containers.sh) | Pass \`--build-arg BUILD_VERSION=\$VERSION\` alongside the existing \`SERVER_DIR\` build-arg. Harmless for Dockerfiles that do not declare the ARG. |
| [docker/Dockerfile.auth](docker/Dockerfile.auth) | Added \`ARG BUILD_VERSION=\"1.0.0\"\` + \`ENV BUILD_VERSION=\$BUILD_VERSION\` in the runtime stage (mirrors the existing pattern in \`docker/Dockerfile.registry\`). |
| [docker/Dockerfile.mcp-server](docker/Dockerfile.mcp-server) | Same. |
| [docker/Dockerfile.mcp-server-light](docker/Dockerfile.mcp-server-light) | Same. |
| [metrics-service/Dockerfile](metrics-service/Dockerfile) | Same. |

## Why this minimal approach (and not the original #919 scope)

The original issue proposed fixing three paths: Helm charts (explicit \`BUILD_VERSION\` env var from \`global.image.tag\`), \`docker-compose.prebuilt.yml\` (source from \`REGISTRY_VERSION\`), and \`publish_containers.sh\`. That covers both build-time and run-time version stomping.

This PR scopes to the **build-time** fix only, per the customer trace showing the problem is already present in the image itself (\`docker inspect\` shows \`1.0.0\` before any env var is applied). Fixing the build-time path means every Docker Hub image from v1.0.22-p1 onward reports the correct version under \`docker inspect\` and in any runtime path that reads \`BUILD_VERSION\`, including future Helm / \`docker-compose.prebuilt\` deployments. The Helm and compose env-precedence follow-ups can be tracked separately if we want defense-in-depth against misconfigured \`envFrom\` secrets, but they are no longer required to fix the symptom.

## Test plan

- [x] \`bash -n scripts/publish_containers.sh\` passes
- [ ] Reviewer to trigger a publish to Docker Hub (or run the script locally with \`--dockerhub --version v1.0.22-p1\`) and confirm:
  \`\`\`
  docker inspect mcpgateway/registry:v1.0.22-p1 | grep BUILD_VERSION
  \"BUILD_VERSION=v1.0.22-p1\"
  \`\`\`
- [ ] Spot-check \`docker inspect mcpgateway/auth-server:v1.0.22-p1 | grep BUILD_VERSION\` reports the correct tag (tests the Dockerfile.auth addition)
- [ ] No change to ECR publish path (GitHub Actions workflow unaffected)

### DIFF
--- a/docker-compose.podman.yml
+++ b/docker-compose.podman.yml
@@ -196,6 +196,8 @@ services:
     build:
       context: metrics-service
       dockerfile: Dockerfile
+      args:
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - METRICS_SERVICE_PORT=8890
       - METRICS_SERVICE_HOST=0.0.0.0
@@ -232,6 +234,8 @@ services:
     build:
       context: .
       dockerfile: docker/Dockerfile.auth
+      args:
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - REGISTRY_URL=${REGISTRY_URL}
       - SECRET_KEY=${SECRET_KEY}
@@ -325,6 +329,8 @@ services:
     build:
       context: servers/currenttime
       dockerfile: ../../docker/Dockerfile.mcp-server
+      args:
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - PORT=8000
       - MCP_TRANSPORT=streamable-http
@@ -337,6 +343,8 @@ services:
     build:
       context: servers/fininfo
       dockerfile: ../../docker/Dockerfile.mcp-server
+      args:
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - PORT=8001
       - SECRET_KEY=${SECRET_KEY}
@@ -353,6 +361,7 @@ services:
       dockerfile: docker/Dockerfile.mcp-server
       args:
         SERVER_DIR: servers/mcpgw
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - HOST=0.0.0.0
       - PORT=8003
@@ -372,6 +381,8 @@ services:
     build:
       context: servers/realserverfaketools
       dockerfile: ../../docker/Dockerfile.mcp-server
+      args:
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - PORT=8002
     ports:
--- a/docker-compose.prebuilt.yml
+++ b/docker-compose.prebuilt.yml
@@ -31,7 +31,11 @@ services:
       - REGISTRY_DESCRIPTION=${REGISTRY_DESCRIPTION:-}
       - REGISTRY_CONTACT_EMAIL=${REGISTRY_CONTACT_EMAIL:-}
       - REGISTRY_CONTACT_URL=${REGISTRY_CONTACT_URL:-}
-      - BUILD_VERSION=${BUILD_VERSION:-1.0.0}
+      # BUILD_VERSION is NOT set here: it is baked into the image by the
+      # publish pipeline (issue #919). Setting it at the compose level with
+      # a "${BUILD_VERSION:-1.0.0}" default would stomp the image-baked
+      # value whenever the operator's .env did not set BUILD_VERSION,
+      # which is the silent-fallback symptom this PR fixes.
       - SECRET_KEY=${SECRET_KEY}
       - AUTH_SERVER_URL=${AUTH_SERVER_URL}
       - AUTH_SERVER_EXTERNAL_URL=${AUTH_SERVER_EXTERNAL_URL}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -293,6 +293,8 @@ services:
     build:
       context: metrics-service
       dockerfile: Dockerfile
+      args:
+        BUILD_VERSION: ${BUILD_VERSION:-1.0.0}
     environment:
       - METRICS_SERVICE_PORT=8890
       - METRICS_SERVICE_HOST=0.0.0.0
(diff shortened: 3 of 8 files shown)


aarora79:
## Extension: covering the local-build path (`build_and_run.sh` / `docker compose build`)

The original commit fixed the Docker Hub **publish** path (`scripts/publish_containers.sh`). Reviewing the local-build path surfaced two more code paths with the same bug:

### 1. `docker compose build` (used by `build_and_run.sh`) was missing the build-arg on 6 of 7 services

Pre-extension, only the `registry` service in `docker-compose.yml` passed `BUILD_VERSION` as a build-arg. So any local build via `build_and_run.sh` produced auth-server / currenttime-server / fininfo-server / realserverfaketools-server / mcpgw-server / metrics-service images with `BUILD_VERSION=1.0.0` baked in, regardless of the git tag.

Fix: added `BUILD_VERSION: \${BUILD_VERSION:-1.0.0}` to each service's `build.args` in both [`docker-compose.yml`](docker-compose.yml) and [`docker-compose.podman.yml`](docker-compose.podman.yml).

### 2. `docker-compose.prebuilt.yml` was stomping the image-baked value at runtime

The prebuilt compose file set `- BUILD_VERSION=\${BUILD_VERSION:-1.0.0}` as a runtime env var on the registry service. For any operator whose `.env` didn't explicitly set `BUILD_VERSION` (the common case), this forced the container runtime to see `1.0.0` even after the image pulled from Docker Hub correctly bakes the right version.

Fix: removed the line in [`docker-compose.prebuilt.yml`](docker-compose.prebuilt.yml). The image-baked `ENV BUILD_VERSION` now wins at runtime, which is what we want.

### Verification

\`\`\`
$ BUILD_VERSION=vTEST docker compose config | grep 'BUILD_VERSION: '
        BUILD_VERSION: vTEST   # x7 — registry + 6 non-registry services
$ docker compose -f docker-compose.prebuilt.yml config | grep BUILD_VERSION
(no output — image-baked value wins at runtime)
\`\`\`

## How to verify locally

From this branch:

\`\`\`bash
./build_and_run.sh
docker inspect mcp-gateway-registry-registry:latest | grep BUILD_VERSION
docker inspect mcp-gateway-registry-auth-server:latest | grep BUILD_VERSION
docker inspect mcp-gateway-registry-metrics-service:latest | grep BUILD_VERSION
\`\`\`

All three should report the current git-describe version (e.g. something like \`BUILD_VERSION=1.0.22-fix/919-dockerhub-build-version\`, not \`1.0.0\`). Pre-extension, only the first would have been correct.

Commit: `e33c2e3f`.

PR #977, opened 2026-05-08
feat(helm): support extraEnv and extraEnvFrom in subcharts and stack

omrishiv:
Add `extraEnv` and `extraEnvFrom` values to auth-server, registry, and mcpgw subcharts (and the parent stack), letting operators inject arbitrary env vars or envFrom sources into each service without forking the charts.

User entries are appended to the end of the container's `env:` and `envFrom:` lists, so they win over chart-managed envFrom sources on name collision. A template-render-time validator rejects `extraEnv` entries that are missing a `name`, duplicate another user entry, or collide with a chart-reserved name. Reserved names are the union of every env: name the chart may render and every key sourced via envFrom from stack-level and per-chart secrets/configmaps, giving over-rejection as the safe default.

`extraEnvFrom` is intentionally not validated (Helm cannot introspect external ConfigMap/Secret contents); the values.yaml documentation names the catastrophic keys to avoid.

Adds helm-unittest suites for each chart (41 tests total) covering baseline render, append ordering, reserved-name rejection, duplicate rejection, missing-name rejection, and cross-subchart isolation at the stack level. Wires `helm unittest` into the existing .github/workflows/helm-test.yml so these tests gate PRs.

Updates CLAUDE.md with a Helm Chart Development section explaining the unittest workflow and the invariant that the reserved-name list and deployment template must be kept in sync.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/helm-test.yml
+++ b/.github/workflows/helm-test.yml
@@ -26,12 +26,12 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
       - name: Set up Helm
-        uses: azure/setup-helm@v4
+        uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2
         with:
-          version: '3.14.0'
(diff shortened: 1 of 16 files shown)


aarora79:
Tracked by issue #985.

PR #1010, opened 2026-05-11
feat(auth): resource-bound JWT tokens (#944)

### LINKED ISSUE #944: Resource-Bound Tokens
## Problem

Bearer tokens minted from the UI are scoped only to the user's group/scope memberships, not to the specific resource the user was viewing when they generated the token. There is no `resource_id` (or similar) claim in the JWT.

Authorization is a broad "does this user's scope grant access to server X?" check — not "was this token bound to server X?"

An agent holding a token generated for Virtual MCP Server *A* can reach Virtual MCP Server *B*, or any regular MCP server, agent, or skill the user has access to, by changing the path in its request. The virtual server is a convenience/routing layer, not a security boundary. The same is true for tokens minted from any resource detail page.

### Sources

- **Mint:** `auth_server/server.py:2167` — `GenerateTokenRequest` takes `user_context`, `requested_scopes`, `expires_in_hours`. No resource field. JWT claims at lines 2245–2262 include `sub`, `groups`, `scope`, but no `resource_type` / `resource_id`.
- **Frontend (sidebar "Get JWT token"):** `frontend/src/components/Sidebar.tsx:284`, handler `fetchAdminTokens` (lines 92–116). Posts `{description, expires_in_hours}` — no resource.
- **Frontend (token page):** `frontend/src/pages/TokenGeneration.tsx:55` — user picks scopes + expiration; no resource path is passed.
- **Proxy:** `docker/nginx_rev_proxy_http_only.conf:77,93` — forwards `X-Scopes` to every upstream, no per-location restriction.
- **Authorization:** `registry/auth/dependencies.py:442-454` — `user_can_access_server()` asks "is `server_name` in the list of servers these scopes grant?"; does not ask "was this token bound to `server_name`?"
- **Resource detail routes** (all use `enhanced_auth` with scope-superset checks, no token-to-resource binding):
  - Server: `registry/api/server_routes.py:4294` (GET `/api/servers/{path}`)
  - Virtual server: `registry/api/virtual_server_routes.py:130-162, 278-293`
  - Agent: `registry/api/agent_routes.py:1198, 1247, 1409`
  - Skill: `registry/api/skill_routes.py:787, 889, 973`

## Resource types in scope

| Type | Identifier | Example |
|------|------------|---------|
| `server` | service path | `/my-server` |
| `virtual_server` | slug under `/virtual/` | `/virtual/my-agg` |
| `agent` | path | `/code-reviewer` |
| `skill` | path | `/python-linter` |

### Two token kinds, one mint endpoint

`GenerateTokenRequest` gains an optional `resource` field:

```json
{
  "description": "...",
  "expires_in_hours": 8,
  "resource": {
    "type": "server | virtual_server | agent | skill",
    "id": "/path"
  }
}
```

- **`resource` present** → mint a **resource-bound token**. JWT includes `token_kind: "resource"`, `resource_type`, `resource_id`.
- **`resource` absent** → mint a **user token** (today's behavior). JWT includes `token_kind: "user"`. Grants everything the user has access to. This is what the sidebar button continues to produce.

### Mint-time validation

In `auth_server/server.py:2167` (`generate_user_token` path):

- If `resource` is present, verify the requested resource is in the user's accessible list for that type (`accessible_servers`, `accessible_agents`, etc. from `registry/auth/dependencies.py`). Reject with 403 otherwise.
- If `resource` is absent, mint as today (`token_kind: "user"`).

### Frontend split

| UI surface | Behavior |
|------------|----------|
| Sidebar "Get JWT token" (`Sidebar.tsx:284`) | No `resource` field → user token. Label clearly: "Personal access token — full access." |
| Resource detail pages (server / virtual / agent / skill) | Always pass `resource` when minting. Remove any code path that lets a resource detail page produce an unbound token. |
| `TokenGeneration.tsx` | Either (a) require the user to pick a resource before minting, or (b) reserve the page for user-token flows only.

### Enforcement at the edge

Route guards (behind a shared FastAPI dependency, e.g. `require_resource_binding(resource_type)`) inspect `token_kind`:

- `token_kind == "user"` → allow any resource the user's scopes permit (status quo).
- `token_kind == "resource"` → require `resource_type` and `resource_id` to match the requested path. Mismatch = 403.
- Missing `token_kind` (legacy tokens) → treat as `user` during a deprecation window; then reject.

Apply to:

- `registry/api/server_routes.py` (including the `/mcp/{path}` proxy path)
- `registry/api/virtual_server_routes.py`
- `registry/api/agent_routes.py`
- `registry/api/skill_routes.py`

Also update the nginx `auth_request` subrequest (`docker/nginx_rev_proxy_http_only.conf:77`) so the auth server compares the request URI against the token's `resource_id`. Without this, traffic that hits nginx before FastAPI bypasses the guard.

### Migration

Legacy tokens have no `token_kind`.

Two options:

- **(a)** Treat missing claim as `user` and require admin scope to use. Safer operationally.
- **(b)** Hard-invalidate at a cutover date. Safer security-wise.

Recommendation: **(b)** with a short deprecation window (e.g. 2 weeks), since this is a security fix.

## Acceptance criteria

- [ ] `GenerateTokenRequest` accepts optional `resource: {type, id}`.
- [ ] Mint rejects resource-bound requests for resources the user cannot access.
- [ ] JWT includes `token_kind`, and when bound, `resource_type` + `resource_id`.
- [ ] Sidebar "Get JWT token" still produces a user token.
- [ ] Every resource detail page mints bound tokens only.
- [ ] Route guards enforce `token_kind == "resource"` ⇒ path match.
- [ ] Nginx auth subrequest enforces the same check at the proxy layer.
- [ ] Legacy tokens handled per chosen migration path.
- [ ] Tests cover: bound-token path match, bound-token path mismatch (403), user-token unrestricted access, mint rejection for inaccessible resource.

omrishiv:
## Summary                                                                                                                             
- Adds an optional `resource: {type, id}` field to JWT mint so tokens can be scoped to a single resource at issuance; nginx edge enforces the binding at `/validate` (403 on mismatch).           
- New `registry/auth/resource_binding.py` is the single source of truth for `ResourceType` / `TokenKind` enums, JWT claim-name constants, deny/allow list, URL classifier, normalizer, and the pre-mint authorization helper (with `asyncio.wait_for` timeout on DB lookups).                                                                                                                            
- auth-server `/validate` handles every `token_kind` case fail-closed — no claim + self-signed → 403, no claim + external IdP → accepted as user, `user` → pass, `resource` → require type+id match vs. URL, else → 403.                                               
- Registry `/api/tokens/generate` accepts optional `resource`, runs pre-mint authorization, forwards to `/internal/tokens`, decodes the returned JWT to detect a silent rolling-upgrade downgrade (returns 500 with `Connection: close`), and records distinct audit actions for user vs. resource-bound mints.                        
- Frontend: new `ResourceBoundTokenButton` in the four detail modals; `ServerConfigModal` (Connect flow) now mints resource-bound tokens for the server being connected to.                                                                                              
- Docs updated: `docs/api-reference.md`, `docs/jwt-token-vending.md` (includes deploy-ordering guidance).                            
                                                                                                            
## Breaking change                                                                                                                     
Self-signed tokens without a `token_kind` claim (minted by a pre-#944 gateway) are now rejected at `/validate` with 403. External IdP tokens (Cognito, Keycloak, Entra, Okta, Auth0) and session cookies are unaffected — they never carry the claim and continue to flow through as user tokens. Cached self-signed tokens from before the upgrade must be re-minted.                                         
                                                                                     
**Deploy order:** auth-server first (or blue/green). Otherwise there's a window where a new registry mints tokens against an old auth-server that silently drops the unknown claim.                                                                                   
                                                                                                                            
## Test plan                                              
- [ ] `uv run pytest tests/unit/auth/test_resource_binding.py tests/auth_server/unit/test_resource_bound_tokens.py` — new unit suites pass (1,428 LOC added).
- [ ] Full suite: `uv run pytest tests/ -n 8` — no regressions.
- [ ] Mint a user token (no `resource` field) — `/validate` accepts it for any URL the user can reach.
- [ ] Mint a resource-bound token for `service:foo` — `/validate` accepts matching URL, rejects any other service URL with 403.
- [ ] Mint with `resource.id` containing `..` or `%` — rejected at the API boundary (Pydantic validator).
- [ ] Pre-mint DB lookup timeout — token mint fails cleanly, no hang.
- [ ] External IdP JWT (Cognito/Keycloak) — passes `/validate` as a user token.
- [ ] Pre-#944 self-signed token — rejected 403 at `/validate`.
- [ ] Rolling-upgrade downgrade detection: point new registry at old auth-server, mint — registry returns 500 with `Connection: close`.
- [ ] Frontend: `ResourceBoundTokenButton` appears in Server/Agent/Skill/VirtualServer detail modals; Connect flow in `ServerConfigModal` produces a resource-bound token; `VirtualServerCard` passes `resourceType='virtual_server'`. 

Fixes #944


aarora79:
Reviewed end-to-end against main. Strong PR — single source of truth in `registry/auth/resource_binding.py`, fail-closed everywhere, 116 new tests pass, and broader auth suites green (419 passed, no regressions).

Two small should-fix items and the ruff tidy-up below. Nothing blocking.

## Should fix

### 1. `auth_server/server.py` lines 2090-2097 — `isinstance` before truthiness in claim check

Today Pydantic `ResourceBinding` gates input on the mint side, so the claim is always a `StrictStr`. But `/validate` trusts whatever is in the JWT — a forged claim (attacker with `SECRET_KEY`, or a future mint path that skips the Pydantic model) could set `resource_id: []`, which is truthy-falsy (`not []` → `True` so the check catches it today) but `resource_id: ["x"]` would pass `not claim_id` and fall through as a non-string. Tighten the ordering:

```python
# replace lines 2090-2097 with:
if (
    not isinstance(claim_type, str) or not claim_type.strip()
    or not isinstance(claim_id, str) or not claim_id.strip()
):
```

Same intent, explicit "must be non-empty string", no reliance on truthy semantics for non-string types.

### 2. `frontend/src/components/ResourceBoundTokenButton.tsx` line 60 — narrow the catch

```tsx
// current
} catch (err: any) {
  setError(err.response?.data?.detail || err.message || 'Failed to generate token');

// suggested
} catch (err: unknown) {
  if (axios.isAxiosError(err)) {
    setError(err.response?.data?.detail || err.message || 'Failed to generate token');
  } else if (err instanceof Error) {
    setError(err.message);
  } else {
    setError('Failed to generate token');
  }
```

Matches the pattern used in `TokenGeneration.tsx` and avoids `any` leaking through strict TS configs.

## Ruff tidy-up

Seven lint items — five auto-fixable, two pre-existing `UP042` patterns matching the rest of the codebase (leave them). Run:

```bash
uv run ruff check --fix registry/auth/resource_binding.py registry/api/server_routes.py auth_server/server.py
```

That clears:
- `registry/auth/resource_binding.py:457,488` — `asyncio.TimeoutError` → builtin `TimeoutError`
- `registry/auth/resource_binding.py:517` — missing trailing newline
- `auth_server/server.py:1026` and `registry/api/server_routes.py:1` — import block sort

## Consider (follow-ups, not this PR)

- `auth_server/server.py:2176` — `req_type.value != claim_type` is asymmetric with `token_kind == TokenKind.USER` four lines up; drop `.value` for consistency (both sides of `ResourceType` inherit from `str`).
- `registry/api/server_routes.py:2591` — `from ..auth.resource_binding import TOKEN_KIND_CLAIM, TokenKind` duplicates the one at line 2421 inside the same function. Hoist to module scope.
- Add a unit test for the federated-server classifier caveat documented at `resource_binding.py:168-177` (mint `server:peer-registry/cf-docs` → request `/api/servers/peer-registry/cf-docs` → 403).
- Prometheus counter `mcp_registry_resource_binding_total{outcome=mint_success|mint_authz_denied|validate_mismatch|validate_blocked}` would be a great observability hook.
- `ServerConfigModal` Connect flow now silently mints bound tokens — consider a one-line callout in the modal body so the UX shift isn't invisible.

Issue #933, opened 2026-05-05
Bug: inconsistent admin authorization between proxied auth and session-cookie `enhanced_auth` paths


zoltan-fedor:
### Summary
The registry can derive different admin authorization results for the same authenticated user depending on which auth dependency a route uses.

Routes using the proxied auth path treat the user as admin, while routes using the session-cookie `enhanced_auth` path can treat the same user as non-admin.

### Impact
A user can successfully authenticate and appear as admin in the main UI/auth context, but still get `403 Admin access required` on some admin-only routes.

Example impact:
- main auth context shows admin access
- admin UI is visible
- some admin-only configuration routes still fail

### Expected behavior
Equivalent authenticated users should resolve to the same effective scopes, UI permissions, and `is_admin` status regardless of whether a route uses proxied auth or session-cookie auth.

### Actual behavior
The registry has multiple auth resolution paths, and they do not consistently derive the same authorization result from the same underlying user session/group data.

### Reproduction
1. Authenticate with an external IdP user that should resolve to admin access.
2. Confirm the main auth context reports admin access.
3. Call a route that uses proxied auth and verify it succeeds.
4. Call an admin-only route that uses `enhanced_auth`.
5. Observe that the second route can return `403 Admin access required` for the same user.

### Likely root cause
Different route dependencies rebuild authorization differently:
- proxied auth uses one path for scopes/admin derivation
- `enhanced_auth` uses another path

Those paths can diverge, which leads to inconsistent `is_admin` results.

### Suggested fix
Unify admin/scope derivation so all auth dependencies produce the same effective authorization result for the same user/session, or refactor both paths to share the same resolution logic.

### Notes
This is distinct from the separate Entra issue where sessions may initially be created without group claims. This bug remains even after the user session contains the correct group information.

Issue #929, opened 2026-05-04
Entra login can create sessions without groups when ID token omits inline group claims


zoltan-fedor:
### Summary

When authenticating with Microsoft Entra ID, the OAuth callback can create a valid session cookie without any group IDs if the ID token does not contain inline `groups` or `roles`.

### Impact

Users can log in successfully but receive no scopes or UI permissions, so admin users cannot register MCP servers.

### Actual behavior

A successful Entra login may produce a session where the effective auth context is empty:
- `groups: []`
- `scopes: []`
- `ui_permissions: {}`

### Expected behavior

If Entra omits inline group claims, the auth flow should still recover the user's group Object IDs and create a session with usable group membership.

### Notes

This appears to happen with Entra group overage-style behavior, where the token does not include direct group IDs.

PR #1042, opened 2026-05-15
Fix oversized session cookie + bundle related auth fixes

### LINKED ISSUE #399: feat: Implement server-side token storage to resolve session cookie size limit

### LINKED ISSUE #915: OAUTH_STORE_TOKENS_IN_SESSION=false still stores id_token in OAuth session cookie, causing large-cookie login failures with Entra ID
## Summary

When `OAUTH_STORE_TOKENS_IN_SESSION=false`, the OAuth callback still stores `id_token` in the session cookie. For providers like Microsoft Entra ID, this can make the session cookie too large and cause login to fail after the IdP redirect completes.

## Observed behavior

1. User completes Entra login successfully.
2. App returns from the IdP redirect.
3. Session is not established reliably and the UI falls back to the login screen.

In our case this was reproducible in deployed environments using Entra.

## Expected behavior

When `OAUTH_STORE_TOKENS_IN_SESSION=false`, the session cookie should not contain token-derived fields, including `id_token`.

## Why this looks incorrect

The documented purpose of `OAUTH_STORE_TOKENS_IN_SESSION=false` is to avoid cookie-size issues for OAuth/OIDC providers with large tokens. However, current behavior still includes `id_token` in the session, which defeats that purpose.

There also appears to be a mismatch between intent and implementation:

- PR/release history says token storage should be disabled by default to prevent cookie-size issues
- current callback behavior still persists `id_token`
- current tests appear to expect that behavior

## Suspected root cause

In `auth_server/server.py`, `oauth2_callback()` builds `session_data` with `id_token` unconditionally, while only token metadata is gated by `OAUTH_STORE_TOKENS_IN_SESSION`.

## Proposed fix

Only include `id_token` in the session when `OAUTH_STORE_TOKENS_IN_SESSION=true`.

That would make the default path fully token-free again and align the code with the intent of the flag.

## Notes

This is especially relevant for Entra ID, where ID tokens can be large due to group or role claims.

If `id_token` is needed for OIDC logout via `id_token_hint`, that likely needs to be treated as an explicit tradeoff when token storage is enabled, rather than happening in the default disabled path.

### LINKED ISSUE #971: Oversized OAuth session cookie breaks login for Entra users with large group memberships
## Summary

Users authenticated via Microsoft Entra ID can fail to log in when their IdP token contains a large `groups` claim.

The auth server currently serializes the full OAuth session payload, including all external group IDs, into the `mcp_gateway_session` browser cookie. For users with large group memberships, the resulting `Set-Cookie` header exceeds common browser limits such as 4096 bytes. Microsoft Edge rejects the cookie, and the user appears to authenticate successfully at the IdP but is not logged into the application.

This was reproduced in a downstream deployment using Entra ID, where one user with a very large group list could not log in while another user with fewer groups could.

## Impact

- OAuth login appears to succeed, but the browser does not persist the session
- The first authenticated UI bootstrap request fails with 401 because no session cookie is sent
- This is user-specific and depends on the size of the external `groups` claim
- Entra ID is the easiest provider to hit this with, but the issue is not Entra-specific in principle

## Observed Behavior

The auth server logs show successful token exchange and successful user extraction from the Entra ID token, followed by successful authentication and a redirect.

Immediately after that, the first authenticated request fails because no valid session cookie is present.

In the browser, the `Set-Cookie` value for `mcp_gateway_session` is longer than 4096 characters, and Edge refuses to save it.

## Expected Behavior

OAuth-authenticated users should be able to log in regardless of how many external groups they belong to.

Large IdP group membership should not make session establishment fail at the browser layer.

## Root Cause

The long-lived OAuth session cookie currently stores the full session payload client-side, including the raw `groups` array.

Relevant code path:

- `auth_server/server.py`
- `oauth2_callback()` builds `session_data` with `groups`
- `oauth2_callback()` serializes that payload into `mcp_gateway_session`

The current design is especially vulnerable for Entra users because Entra can return large `groups` arrays in ID tokens.

## Why This Is Hard To Fix By Simply Removing `groups`

At first glance, it seems attractive to just drop `groups` from the cookie, but raw group identity is used for runtime behavior in several places.

Examples include:

- Registry cookie-auth scope derivation from session data
- Group-restricted agent and skill visibility checks
- Federation export visibility filtering by peer groups
- Some admin and permission checks that still fall back to groups

So the issue is not just cookie size. The underlying problem is that the system is storing too much authorization state in a browser cookie.

## Proposed Direction

The most robust fix appears to be moving the long-lived OAuth session server-side and keeping only an opaque session ID in the browser cookie.

A practical implementation could be:

1. Keep `oauth2_temp_session` as-is for the short-lived OAuth handshake.
2. Replace the long-lived `mcp_gateway_session` payload with a random opaque `session_id`.
3. Store the actual session server-side, including:
   - username
   - email
   - name
   - groups
   - provider
   - auth_method
   - optional `id_token`
   - expiry metadata
   - optionally precomputed scopes
4. Make both auth-server and registry resolve the session from shared storage.
5. Support both legacy self-contained cookies and the new opaque-session format during rollout.

## Possible Storage Options

The least disruptive option seems to be reusing the existing shared DocumentDB or MongoDB integration already used elsewhere in the project, rather than introducing a brand-new dependency just for sessions.

## Minimal Reproduction

1. Configure OAuth login with Entra ID.
2. Use a user whose Entra ID token contains a very large `groups` claim.
3. Complete the OAuth login flow.
4. Inspect the callback response in the browser dev tools.
5. Observe that `Set-Cookie: mcp_gateway_session=...` exceeds browser size limits.
6. Observe that the browser does not store the cookie.
7. Observe that the next authenticated request fails because no session cookie is sent.

## Additional Notes

There is already recognition in the project that oversized OAuth session cookies are a problem when storing tokens client-side. This issue is the same class of problem, but caused by storing large external group claims in the browser session cookie.

## Questions

1. Would you be open to a PR that introduces server-side session storage for the long-lived OAuth session while keeping a compatibility path for existing signed cookies?
2. If so, is reusing the existing DocumentDB or MongoDB abstraction the preferred first implementation, or would you prefer a storage-agnostic interface from the start?

omrishiv:
Fixes: #915 #929 #933 #971 #399 

## Summary

This branch fixes the "session cookie too large" failure that broke login for Entra ID users with many group memberships, plus three related auth issues that surfaced while debugging it. The browser cookie now carries only an opaque, signed `session_id`; the full session record (username, groups, encrypted `id_token`) lives server-side.

Closes:
- **#971** — Oversized OAuth session cookie breaks login for Entra users with large group memberships
- **#399** — Implement server-side token storage to resolve session cookie size limit
- **#915** — `OAUTH_STORE_TOKENS_IN_SESSION=false` still stores `id_token` in session cookie
- **#929** — Entra login can create sessions without groups when ID token omits inline group claims (group overage)
- **#933** — Inconsistent admin authorization between proxied auth and `enhanced_auth` paths

### Server-side session store (#971, #399, #915)
- New `oauth_sessions_<namespace>` MongoDB/DocumentDB collection with TTL on `expires_at`. Auth-server writes the record at the end of `oauth2_callback`; registry reads it on every authenticated request via the opaque session_id in the signed cookie.
- `auth_server/session_store.py` — writer (create/resolve/delete + index management).
- `registry/auth/session_store.py` — reader (resolve/delete).
- `registry/auth/session_crypto.py` — shared crypto helper. `id_token` is encrypted at rest with AES-GCM under a 32-byte key derived from `SECRET_KEY` via HKDF-SHA256. Random 96-bit nonce per record.
- Logout deletes the server-side record before clearing the cookie, closing the cookie-replay window.
- `OAUTH_STORE_TOKENS_IN_SESSION` is **removed** entirely (was a flag that gated `id_token` storage to keep cookies small; with the move to server-side, the flag's purpose is gone). Removed from auth-server, registry config, all 3 docker-compose files, all charts, all 5 terraform files, `.env.example`, `.env.template`, the System Config UI, and the unified parameter reference.

### Strict `SECRET_KEY` (#971 root-cause cleanup)
- Both `auth_server/server.py` and `registry/core/config.py` now refuse to start if `SECRET_KEY` is unset, replacing the per-replica `secrets.token_hex(32)` fallback that silently caused cross-replica `BadSignature` errors.
- `docs/unified-parameter-reference.md` clarifies that the Helm chart auto-generates at install time but Docker/Terraform require an explicit value.

### Entra group-overage fallback (#929)
- New `EntraIdProvider.has_group_overage()` and `fetch_groups_via_graph()` in `auth_server/providers/entra.py`. When the ID token signals overage (`hasgroups: true` or `_claim_names.groups`), the auth-server falls back to `GET /me/memberOf` over Microsoft Graph and pages results, capped at 1000 group IDs.
- Sovereign-cloud aware (public, US Gov, China) via `_LOGIN_TO_GRAPH_HOST`. Honors `ENTRA_GRAPH_BASE_URL` override for proxied/air-gapped deployments.

### Unified admin/scope derivation across cookie + header auth (#933)
- New `_derive_user_context` in `registry/auth/dependencies.py` is the single source of truth for scope→ui_permissions→is_admin derivation. Both `enhanced_auth` (cookie) and `nginx_proxied_auth` (header) call into it.
- The previous header path synthesized groups from a scope heuristic when `X-Groups` was missing. Fixed by trusting `X-Groups` only and deriving admin status purely from the auth_server-computed scopes.

### Other notable changes
- `cli_user_auth.py` `create_session_cookie` is now `NotImplementedError` with a clear migration message — locally minted cookies cannot work with server-side sessions.
- New `tests/auth_server/unit/test_session_store.py` (18 tests): encrypt/decrypt round-trip, nonce uniqueness, SECRET_KEY rotation invalidation, missing-key error, full create/resolve/delete control flow, expired-record handling, naive datetime coercion, store-outage degradation, and silent decryption-failure fallback.
- Cookie-size regression guard: `test_session_cookie_stays_well_under_size_ceiling` asserts the OAuth callback's cookie value stays < 512 bytes.

## Migration / breaking changes

- **All currently-active users will be force-logged-out on rollout.** The new resolver rejects pre-rollout dict-payload cookies and redirects to `/login`. This is intentional — it collapses the migration into one PR with no parallel-format compatibility code.
- **`SECRET_KEY` is now required.** Operators on Docker/Terraform who relied on the runtime fallback must set it explicitly before deploying. Helm chart users are unaffected (chart auto-generates).
- **`OAUTH_STORE_TOKENS_IN_SESSION` no longer exists.** The variable is removed from all deployment surfaces; setting it has no effect.
- **`agents/cli_user_auth.py` is hard-broken.** The CLI's local cookie-minting path is incompatible with server-side sessions; use the auth-server's browser OAuth flow instead.


aarora79:
Merged `main` into this branch to resolve conflicts. The only conflict was in `terraform/aws-ecs/modules/mcp-gateway/ecs-services.tf` (kept `UI_TITLE` from main, dropped `OAUTH_STORE_TOKENS_IN_SESSION` per this PR). Tests passing post-merge (3072 passed / 68 skipped / 0 failed).

A few minor follow-ups will be addressed in an immediate fast-follow PR after this one merges (`ENTRA_GRAPH_BASE_URL` documentation, observability counter, breaking-change release notes, a couple of defensive code touches). None block this merge.

PR #1100, opened 2026-05-21
Preserve MCP proxy response headers

### LINKED ISSUE #1096: bug(mcp-proxy): auth_server drops Mcp-Session-Id and other upstream response headers, breaking streamable-http MCP sessions through the gateway
### Summary

The `/mcp-proxy/{server_name}` hop in [`auth_server/server.py`](https://github.com/agentic-community/mcp-gateway-registry/blob/main/auth_server/server.py) buffers the upstream MCP response and re-emits it via `starlette.responses.Response` / `JSONResponse`, but it only carries `body + status + content-type`. **Every other header from the upstream response is silently dropped — including `Mcp-Session-Id`**, which streamable-http MCP servers emit on `initialize` and which the client requires for follow-up requests.

Effect: any MCP client that uses streamable-http transport (LibreChat, Roo Code, Cursor, etc.) cannot establish a session with any MCP server that sits behind the gateway. The `initialize` call returns 200 with an empty `Mcp-Session-Id`, and the next JSON-RPC call fails with `"Missing session ID"`.

### Affected versions

- v1.24.0 / v1.24.1 (where the `/mcp-proxy` hop was introduced for Issue #1026)
- `main` as of today — the `mcp_proxy` function is **byte-identical** to v1.24.1 (`diff -q` between the v1.24.1 release tarball and the current `main` `auth_server/server.py` is empty).
- Pre-v1.24 releases are NOT affected because nginx forwarded MCP POSTs directly to the upstream MCP server; the auth_server was never on the response path.

### Reproduction

1. Deploy v1.24.1 (or current `main`) to any environment.
2. Register any streamable-http MCP server (e.g. `office-docs-mcp-server`).
3. Point a streamable-http MCP client at the gateway (e.g. LibreChat).
4. Observe `initialize` returns 200 OK with the body unchanged, but the response **omits** the `Mcp-Session-Id` header that the upstream emitted.
5. Any subsequent JSON-RPC call from the client fails with `"Missing session ID"` because the client never received a session to bind to.

### Root cause

[`auth_server/server.py:3875-3907`](https://github.com/agentic-community/mcp-gateway-registry/blob/main/auth_server/server.py#L3875-L3907):

```python
if not should_filter:
    return Response(
        content=body_bytes,
        status_code=status_code,
        media_type=content_type,        # <-- no headers= passed
    )
...
return JSONResponse(
    content=_safe_parse_body(body_bytes),
    status_code=status_code,            # <-- no headers= passed
)
...
return JSONResponse(content=parsed, status_code=status_code)   # <-- no headers= passed
```

All three return paths construct a Starlette response without a `headers=` argument. Starlette then only sets `content-type` (from `media_type`) and `content-length` (from the body) — everything else from the upstream HTTP response is lost.

A secondary contributing bug: even if the response builders were fixed naively, reading `upstream_response.headers` outside the `async with client.stream(...)` block returns an empty mapping because httpx releases the response headers when the stream context manager exits. The fix must snapshot the headers **inside** the `with`.

### Empirical proof (Starlette `Response` behavior)

```python
from starlette.responses import Response

upstream_headers = {
    "content-type": "application/json",
    "mcp-session-id": "abc-123",
    "x-custom-something": "keep-me",
}

# Current main:
r1 = Response(content=b"{}", status_code=200, media_type="application/json")
# raw_headers -> [('content-length', '2'), ('content-type', 'application/json')]
# Mcp-Session-Id is GONE.

# With headers= argument:
r2 = Response(content=b"{}", status_code=200, media_type="application/json",
              headers=upstream_headers)
# raw_headers -> [('content-type', 'application/json'),
#                 ('mcp-session-id', 'abc-123'),
#                 ('x-custom-something', 'keep-me'),
#                 ('content-length', '2')]
```

### Why `MCP_TOOLS_LIST_FILTER_ENABLED=false` is NOT a workaround

It might seem the bug only affects the tools/list filtering path, so disabling the filter should bypass it. **It does not.** Two structural reasons:

1. [`registry/core/nginx_service.py:1166-1202`](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/core/nginx_service.py#L1166-L1202) routes every MCP POST to `http://auth-server:8888/mcp-proxy/...` **unconditionally**. The comment at line 1170 is explicit: *"Nginx never inspects the body or flag; auth_server decides."* No code path skips the proxy hop based on the filter flag.
2. Inside `mcp_proxy`, when `filter_enabled=False` the code reaches `if not should_filter:` and returns the bare `Response(content, status_code, media_type)` at line 3880 — the same header-dropping construct. The flag changes whether `tools/list` arrays get pruned; it does not change which Starlette response object is built.

So today's deployment options for users hitting this are:
- Downgrade to a pre-v1.24 release (loses the per-scope `tools/list` filtering of Issue #1026).
- Patch their own image with the fix below.

### Proposed fix

Symmetric with the existing `_forward_headers` helper used on the request direction. Three edits to `mcp_proxy`:

1. Add `_RESPONSE_FRAMING_HEADERS = frozenset({"content-length", "content-encoding", "transfer-encoding", "connection"})` and a `_passthrough_response_headers(upstream_headers)` helper that returns the upstream dict minus those framing headers (so Starlette can recompute them — important because httpx may have already decoded `content-encoding: gzip`).
2. Snapshot `upstream_headers = dict(upstream_response.headers)` **inside** the `async with client.stream(...)` block.
3. Pass `headers=_passthrough_response_headers(upstream_headers)` on all three returns (passthrough branch, non-JSON tools/list fallback, filtered JSON branch).

Patch totals 59 added / 6 removed lines and changes no existing behavior other than restoring upstream response headers. PR with full diff + 9 new unit tests in `tests/auth_server/unit/test_server.py` (helper unit tests + endpoint-level tests via `TestClient` with a mocked httpx upstream, including a regression guard that explicitly asserts the fix works when `MCP_TOOLS_LIST_FILTER_ENABLED=false`).

### Related history

- #1026 introduced the proxy hop.
- #1051 / PR #1052 fixed a related bug where the same code path forced `Content-Type: application/json` and wrapped SSE bodies in `{"raw": "..."}`. That fix added `media_type=content_type` to the same `Response(...)` calls — but stopped there, leaving every other upstream header still dropped. This issue is the next-shoe-drop after #1051.

Baijack-star:
## Summary
- snapshot upstream MCP response headers while still inside the `httpx` stream context
- pass end-to-end headers such as `Mcp-Session-Id` through all `/mcp-proxy` response branches
- strip framing headers so Starlette can recompute body framing safely
- add regression coverage for passthrough responses and `MCP_TOOLS_LIST_FILTER_ENABLED=false`

## Why
Fixes #1096. Streamable-http MCP clients need the upstream `Mcp-Session-Id` from `initialize` before they can send follow-up requests. The auth proxy currently forwards body/status/content-type but drops that session header, so clients see a successful initialize response but cannot bind later JSON-RPC calls to the session.

## Verification
- `uv run pytest tests/auth_server/unit/test_server.py::TestMcpProxyResponseHeaders -q --no-cov`
- `uv run --with ruff ruff check auth_server/server.py tests/auth_server/unit/test_server.py`

Prepared by an AI agent operator; scope kept to the response-header regression described in #1096.

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -3737,6 +3737,15 @@ _HOP_BY_HOP_HEADERS: frozenset[str] = frozenset(
     }
 )
 
+_RESPONSE_FRAMING_HEADERS: frozenset[str] = frozenset(
+    {
+        "content-length",
+        "content-encoding",
+        "transfer-encoding",
+        "connection",
+    }
+)
+
 
 async def _read_bounded(
     response: httpx.Response,
@@ -3777,6 +3786,17 @@ def _forward_headers(
             continue
         forwarded[key] = value
     return forwarded
+
+
+def _passthrough_response_headers(
+    upstream_headers: dict[str, str],
+) -> dict[str, str]:
+    """Preserve upstream end-to-end headers while letting Starlette frame the body."""
+    return {
+        key: value
+        for key, value in upstream_headers.items()
+        if key.lower() not in _RESPONSE_FRAMING_HEADERS
+    }
 
 
 @app.post("/mcp-proxy/{server_name:path}")
@@ -3848,6 +3868,7 @@ async def mcp_proxy(
                 # pathological upstream cannot DoS the proxy.
                 body_bytes = await _read_bounded(upstream_response, max_body_bytes)
                 status_code = upstream_response.status_code
+                upstream_headers = dict(upstream_response.headers)
                 content_type = upstream_response.headers.get("content-type", "application/json")
     except HTTPException:
         raise
@@ -3881,6 +3902,7 @@ async def mcp_proxy(
             content=body_bytes,
             status_code=status_code,
             media_type=content_type,
+            headers=_passthrough_response_headers(upstream_headers),
         )
 
     try:
@@ -3892,6 +3914,7 @@ async def mcp_proxy(
         return JSONResponse(
             content=_safe_parse_body(body_bytes),
             status_code=status_code,
+            headers=_passthrough_response_headers(upstream_headers),
         )
 
     result = parsed.get("result") if isinstance(parsed, dict) else None
@@ -3904,7 +3927,11 @@ async def mcp_proxy(
         result["tools"] = filtered
         parsed["result"] = result
 
-    return JSONResponse(content=parsed, status_code=status_code)
+    return JSONResponse(
+        content=parsed,
+        status_code=status_code,
+        headers=_passthrough_response_headers(upstream_headers),
+    )
 
 
 def _safe_parse_body(
--- a/tests/auth_server/unit/test_server.py
+++ b/tests/auth_server/unit/test_server.py
@@ -413,6 +413,129 @@ class TestRateLimiting:
         assert f"{username}:{old_hour}" not in user_token_generation_counts
 
 
+class TestMcpProxyResponseHeaders:
+    """Regression tests for streamable-http MCP response headers."""
+
+    class _FakeUpstreamResponse:
+        def __init__(
+            self,
+            body: bytes,
+            headers: dict[str, str],
+            status_code: int = 200,
+        ):
+            self._body = body
+            self.headers = headers
+            self.status_code = status_code
+
+        async def aiter_bytes(self, chunk_size: int):
+            yield self._body
+
+    class _FakeStream:
+        def __init__(self, response):
+            self.response = response
+
+        async def __aenter__(self):
+            return self.response
+
+        async def __aexit__(self, exc_type, exc, tb):
+            return False
+
+    class _FakeAsyncClient:
+        def __init__(self, response):
+            self.response = response
+
+        async def __aenter__(self):
+            return self
+
+        async def __aexit__(self, exc_type, exc, tb):
+            return False
+
+        def stream(self, *args, **kwargs):
+            return TestMcpProxyResponseHeaders._FakeStream(self.response)
+
+    def test_passthrough_response_headers_preserves_mcp_session_id(self):
+        import auth_server.server as server_module
+
+        headers = server_module._passthrough_response_headers(
+            {
+                "content-type": "application/json",
+                "Mcp-Session-Id": "session-123",
+                "Retry-After": "2",
+                "content-length": "999",
+                "transfer-encoding": "chunked",
+                "connection": "keep-alive",
+                "content-encoding": "gzip",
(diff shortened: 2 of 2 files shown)


aarora79:
Thanks, will review today.

aarora79:
Hi @Baijack-star — first off, thanks for the thorough writeup on #1096 and the carefully scoped PR. The root-cause analysis (especially the empirical Starlette `Response` behavior demo and the explanation of why `MCP_TOOLS_LIST_FILTER_ENABLED=false` isn't a workaround) was excellent and made it easy to reason about.

The core symptom (`Mcp-Session-Id` being dropped, breaking streamable-http clients) was fixed on `main` in commit 19b81320 ("fix(auth): forward Mcp-Session-Id header from upstream through mcp-proxy"). That change snapshots the upstream headers inside the `async with client.stream(...)` block and forwards `mcp-session-id` / `x-mcp-session-id` on all three return paths in `mcp_proxy`, so streamable-http MCP clients can now establish and reuse sessions through the gateway.

Where we'd like to push back is on the denylist approach in this PR. The auth-server sits on a trust boundary in front of arbitrary upstream MCP servers, and forwarding every upstream header except the 4 framing ones means an upstream can dictate response headers we'd rather control at the gateway: `Set-Cookie`, `Strict-Transport-Security`, `Content-Security-Policy`, `Access-Control-Allow-*`, `Location`, `Server`, etc. We don't think we can safely take that on without a much longer-tail audit of every header browsers and MCP clients act on.

If you'd be open to converting the implementation to an allowlist, we'd be happy to take this in. Concretely something like:

```python
_FORWARDED_RESPONSE_HEADERS = frozenset({
    "mcp-session-id",
    "x-mcp-session-id",
    "www-authenticate",   # for the PRM/OAuth flow added in #1115
    "retry-after",        # so clients honor upstream 429/503 backoff
})
```

...paired with the test suite you already wrote, plus a regression test asserting that headers like `Set-Cookie` and `Location` are dropped. That gives us the generalization beyond just the two session-id headers (which is a real gap on `main`: `WWW-Authenticate` and `Retry-After` aren't forwarded today) and the test coverage, without surrendering response-header policy to upstream.

Thanks again for the work, happy to review a revised version.

aarora79:
Closing this as #1126 has been merged.

Issue #867, opened 2026-04-21
feat: expose /metrics endpoint on registry for in-process Prometheus counters


aarora79:
## Problem

The registry process declares `prometheus_client` `Counter` objects in several modules (e.g. [registry/auth/routes.py](registry/auth/routes.py) for login/logout tracking, and would-be counters for other features like the M2M management routes in #851), but **there is no HTTP endpoint on the registry that exposes them**. A `curl $REGISTRY_URL/metrics` today falls through nginx's SPA catch-all and returns the React `index.html`. As a result:

- Counters declared in the registry process are dead code — incremented but never scraped.
- Prometheus (local `prometheus-1` container) only scrapes the separate `metrics-service:9465`, not the registry.
- Features like #851 that would benefit from per-operation metrics cannot ship them without relying on the audit log alone.

## Proposed Solution

1. **Mount the default Prometheus collector** at `/api/metrics` inside the registry FastAPI app:
   ```python
   from prometheus_client import make_asgi_app

   app.mount("/api/metrics", make_asgi_app())
   ```
   Path `/api/metrics` (rather than `/metrics`) avoids conflicting with the nginx SPA catch-all and matches how other `/api/*` routes are already exposed.

2. **Add a scrape job** to `prometheus.yml` for the registry container:
   ```yaml
   - job_name: 'mcp-registry'
     static_configs:
       - targets: ['registry:7860']   # or the appropriate internal port
     scrape_interval: 10s
     metrics_path: /api/metrics
   ```

3. **Decide on auth posture** for the endpoint. Options:
   - Leave it unauthenticated but only accessible on the internal docker network (simplest; matches how `metrics-service:9465` works today).
   - Require a static token / mTLS (more secure; more work).

4. **Audit existing in-process counters** so the first scrape exposes useful data:
   - `logout_id_token_hint_present` / `logout_id_token_hint_missing` / `logout_jwt_validation_failed` / `logout_url_length_warning` in [registry/auth/routes.py](registry/auth/routes.py)
   - Any additional counters introduced by feature PRs (e.g. #851's `m2m_management_requests_total` once restored).

5. **Update operator documentation** to describe the new endpoint and basic dashboard queries.

## Motivation

- Closes the gap where counters exist but produce no signal.
- Enables per-feature observability (rate of 4xx/5xx on specific endpoints, abuse detection on admin APIs, etc.) without building new telemetry infrastructure.
[...]

PR #1109, opened 2026-05-22
feat(metrics): expose /api/metrics Prometheus endpoint on registry

NandiniKodali988:
**Issue #:** #867

**Description of changes:**

- Adds `GET /api/metrics` endpoint to the registry that serves Prometheus-format metrics
- Aggregates registry-local metrics (deployment mode, nginx config writes, logout counters, cloud detection) with application metrics from the metrics-service (auth requests, tool discovery, tool executions, health checks, protocol latency)
- Added unauthenticated NGINX location block (`location ^~ /api/metrics`) in both HTTP-only and HTTP+HTTPS templates so the endpoint is reachable without a JWT
- Updated `config/prometheus.yml` to add a scrape job targeting `registry:7860/api/metrics`
- Documented the endpoint in `docs/OBSERVABILITY.md`

**Test plan:**

- [ ] `uv run pytest tests/unit/api/test_metrics_endpoint.py` passes (5 tests)
- [ ] `curl http://localhost/api/metrics` returns HTTP 200 with Prometheus text
- [ ] Prometheus target `mcp-registry` shows state `UP` at `/targets`
- [ ] PromQL queries like `mcp_tool_executions_total` return results

### DIFF
--- a/config/prometheus.yml
+++ b/config/prometheus.yml
@@ -12,4 +12,12 @@ scrape_configs:
       - targets: ['metrics-service:9465']
     scrape_interval: 10s
     metrics_path: /metrics
+
+  # Scrape in-process counters from the registry FastAPI app (issue #867).
+  # Requires the registry container to be on the same Docker network as Prometheus.
+  - job_name: 'mcp-registry'
+    static_configs:
+      - targets: ['registry:7860']
+    scrape_interval: 10s
+    metrics_path: /api/metrics
     
--- a/docker/nginx_rev_proxy_http_and_https.conf
+++ b/docker/nginx_rev_proxy_http_and_https.conf
@@ -462,6 +462,22 @@ server {
         proxy_read_timeout 30s;
     }
 
+    # Prometheus metrics endpoint - no authentication required (issue #867)
+    location ^~ {{ROOT_PATH}}/api/metrics {
+        proxy_pass http://127.0.0.1:7860;
+        proxy_http_version 1.1;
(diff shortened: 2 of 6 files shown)


aarora79:
Thanks @NandiniKodali988 for the work on this — the implementation, tests, and documentation are clean and well-structured.

After reviewing the deployment-surface implications, we've decided to take a different architectural direction. The core issue: this PR exposes the registry's in-process Prometheus counters at /api/metrics, but on ECS those counters can't reach AMP without a separate ADOT collector wired up to scrape the registry, and on EKS there's no Prometheus stack at all. So the new endpoint adds genuine value only on Compose — and even there, it lives alongside (rather than replacing) the existing two-codepath pattern (HTTP POST to metrics-service plus auto-instrumentation OTLP push).

The cleaner solution is to migrate **all** custom application metrics to native OpenTelemetry meter emission directly inside registry, auth-server, and mcpgw. They'd ride the same OTLP push the auto-instrumentation already uses — which means EKS customers finally see auth_request, tool_execution, and the in-process counters (nginx_config_writes_total, peer_sync_failures_total, etc.) for the first time. The metrics-service container, its SQLite store, the dual scrape pattern, and this PR's /api/metrics endpoint all go away.

This is tracked in #1122. Closing in favor of that direction. Issue #867 (which this PR was closing) is also being closed because it's obsoleted by the same work.

Again, thank you for the contribution — the analysis and discussion this PR triggered is what surfaced the architectural simplification opportunity.

PR #1126, opened 2026-05-26
feedback on review #1097 has been addressed by using response header allowlist so that upstream MCP response headers are safe to forward back to the client

zahangirbu:
*Issue #1096*

*Description of changes:*
Feedback on review #1097 has been addressed by using response header allowlist so that upstream MCP response headers are safe to forward back to the client.

Allowed list:
```
_FORWARDED_RESPONSE_HEADERS: frozenset[str] = frozenset(
    {
        "mcp-session-id",
        "x-mcp-session-id",
        "www-authenticate",
        "retry-after",
    }
)
```

This mechanism prevents an upstream from dictating response-header policy that browsers / MCP clients honor (Set-Cookie, Strict-Transport-Security, Content-Security-Policy, Access-Control-Allow-*, Location, Server, etc.) 


zahangirbu:
Hi @aarora79,
Based on your feedback on #1097, I have addressed your feedback by resolving the conflicts. Please have a look and let me know if you need anything else.

PR #1128, opened 2026-05-27
feat(gitlab): translate GitLab /-/raw/ URLs to API v4 for private repos

madhuc-ghub:
## Summary

GitLab's `/-/raw/` web URLs require session cookies and therefore can't be
fetched from a private repository with only a Personal Access Token. The
API v4 raw file endpoint
(`/api/v4/projects/:id/repository/files/:path/raw`) accepts a
`PRIVATE-TOKEN` header instead, which is exactly what the registry already
configures via `auth_credential`.

This PR adds a small **optional** helper module,
`registry/utils/gitlab_url_utils.py`, plus three defensive hook points in
`skill_service` and `url_utils`. With the helper present, the registry can
fetch SKILL.md, sibling resource files, and recursive tree listings from
private GitLab projects using the same PAT-style auth flow that already
works for GitHub. With the helper absent (e.g. you delete the file), the
hooks fall through and existing GitHub / raw URL behaviour is completely
unchanged.

### What's in the helper

- `parse_gitlab_url` — parses both `/-/raw/` web URLs and API v4 raw-file
  URLs into base, project, ref and filepath.
- `translate_gitlab_to_api_url` — rewrites a `/-/raw/` URL to the
  equivalent API v4 raw-file URL so `PRIVATE-TOKEN` works.
- `derive_gitlab_resource_url` — resolves a sibling resource (e.g. a file
  referenced from a SKILL.md) to its API v4 URL with proper percent
  encoding.
- `translate_gitlab_tree_api_url` — builds a recursive tree API URL so
  multi-file skill discovery works against GitLab the same way it does
  against GitHub.

### Why this is safe for upstream

Every call site uses a `try/except ImportError` guard:

- `skill_service._build_fetch_headers` only attempts the rewrite when both
  auth headers are set **and** the URL contains `gitlab`.
- `skill_service._resolve_tree_api` tries the GitLab translator first and
  falls through to the existing GitHub / GHES handling on `None`.
- `url_utils.derive_resource_url` tries GitLab first and falls back to the
  current `SKILL.md`-stripping behaviour.

There are no new required dependencies, no changes to existing GitHub
code paths, and no behavioural change for any non-GitLab URL.

## Test plan

- [x] 20 new unit tests in `tests/unit/utils/test_gitlab_url_utils.py`
      covering:
  - parsing `/-/raw/` and API v4 file URLs
  - `file_dir` / `encoded_project` helpers (nested groups, root files)
  - translation `/-/raw/` → API v4
  - idempotency when the input is already an API v4 URL
  - sibling resource derivation with subdirectories and at the repo root
  - percent-encoding of spaces and reserved characters in resource paths
  - tree-API URL construction and `skill_dir` extraction
  - `None` fall-through for empty / non-GitLab / blob-style URLs
- [x] Tests pass locally (`pytest tests/unit/utils/test_gitlab_url_utils.py`
      → 20 passed).
- [ ] Reviewer sanity-check that GitHub-only paths are unaffected (the
      defensive imports are the only edits in `skill_service.py` and
      `url_utils.py`).

### DIFF
--- a/registry/services/skill_service.py
+++ b/registry/services/skill_service.py
@@ -196,9 +196,10 @@ def _build_fetch_headers(
 ) -> tuple[str, dict[str, str]]:
     """Build fetch URL and auth headers for a SKILL.md request.
 
-    Returns (fetch_url, headers) tuple.  The URL is returned unchanged;
-    platform-specific modules may override to translate web URLs to
-    authenticated API endpoints.
+    For GitLab URLs with credentials, translates /-/raw/ web URLs to
+    API v4 endpoints that accept PRIVATE-TOKEN headers.
+
+    Returns (fetch_url, headers) tuple.
     """
     headers: dict[str, str] = {}
     fetch_url = url
@@ -212,6 +213,16 @@ def _build_fetch_headers(
     elif auth_scheme == "api_key":
         header_name = auth_header_name or "PRIVATE-TOKEN"
         headers[header_name] = auth_credential
+
+    if headers and "gitlab" in url.lower():
+        try:
+            from ..utils.gitlab_url_utils import translate_gitlab_to_api_url
+
+            api_url = translate_gitlab_to_api_url(url)
+            if api_url:
+                fetch_url = api_url
+        except ImportError:
+            pass
 
     return fetch_url, headers
 
@@ -290,6 +301,18 @@ def _resolve_tree_api(
         - ``skill_dir``: directory portion of the path leading up to
           ``SKILL.md`` (``""`` when SKILL.md is at the repo root).
     """
+    # Try GitLab translation first (Meraki: defensive import so removing
+    # gitlab_url_utils leaves upstream features fully functional for GitHub).
+    try:
+        from ..utils.gitlab_url_utils import translate_gitlab_tree_api_url
+
+        gitlab_result = translate_gitlab_tree_api_url(skill_md_url)
+        if gitlab_result is not None:
+            return gitlab_result
+    except ImportError:
+        pass
+
+    # Fall through to upstream GitHub / GHES handling.
     parsed = urlparse(skill_md_url)
     hostname = parsed.hostname or ""
     if not hostname:
--- a/registry/utils/gitlab_url_utils.py
+++ b/registry/utils/gitlab_url_utils.py
@@ -1,0 +1,118 @@
+"""
+GitLab-specific URL utilities for API v4 endpoint translation.
+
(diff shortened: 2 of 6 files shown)


registry/services/skill_service.py
```diff
@@ -213,6 +214,16 @@ def _build_fetch_headers(
         header_name = auth_header_name or "PRIVATE-TOKEN"
         headers[header_name] = auth_credential
 
+    if headers and "gitlab" in url.lower():
```
omrishiv, line 217: Can we use different matching for gitlab? The way it is currently written means that any self-hsoted gitlab that doesn't have `gitlab` in the name will fail to use the `translate_gitlab_to_api_url`.

tests/unit/services/test_skill_service_gitlab.py
```diff
@@ -0,0 +1,184 @@
...
+    _resolve_tree_api,
+)
+
+
+# =============================================================================
+# _build_fetch_headers — GitLab URL translation
+# =============================================================================
+
+
+class TestBuildFetchHeadersGitlab:
```
omrishiv, line 33: Can we add some tests for domains that do not include `gitlab` in the hostname?

registry/services/skill_service.py
omrishiv, line 217:
```suggestion
    if headers and parse_gitlab_url(url):
```

registry/utils/gitlab_url_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    if not parts:
+        return None
+
+    skill_dir = parts.file_dir
+    if not skill_dir:
+        return None
+
+    tree_url = (
+        f"{parts.base}/api/v4/projects/{parts.encoded_project}/repository/tree"
+        f"?path={url_quote(skill_dir, safe='')}&ref={parts.ref}&recursive=true&per_page=100"
```
omrishiv, line 116: If there are more than 100 skills, this will drop anything over 100. https://github.com/agentic-community/mcp-gateway-registry/blob/ec1aa9718b3ca864e45e769edb8a29da80d62313/registry/services/skill_service.py#L427-L433 for reference

omrishiv (changes requested):
Hi @madhuc-ghub, thanks for the PR. I tested it on my end and was able to pull a gitlab skill from a private, self-hosted registry. Added a few comments/suggestions, please take a look. Also, is it possible to update the documentation for the `GITHUB_EXTRA_HOSTS`? The variable name may need to be changed, but at least the documentation should highlight adding the Gitlab host in there

PR #1173, opened 2026-06-03
Feat/custom entities

### LINKED ISSUE #1172: Feature: Custom Entity Types (admin-defined, schema-driven catalog types)
## Summary

Today the registry catalogs three first-class entity types (MCP servers, A2A
agents, skills), each bespoke top-to-bottom: its own Pydantic model, MongoDB
collection, routes, repository, and React components. Adding a new catalog type
(n8n workflows, rules, policies, etc.) currently requires a code change and
deploy per type.

This feature adds an admin-defined, schema-driven mechanism so a
`registry-admin` can define a new **catalog** entity type at runtime — a name
plus a field schema — and users can then create, list, edit, search, and delete
records of that type. No per-type code, no deploy.

The feature ships behind a flag (`CUSTOM_ENTITY_TYPES_ENABLED`, default off), so
existing deployments see no change until they opt in.

## Problem statement

Users want to catalog additional asset types alongside the built-in ones
without waiting on an engineering change for each new type. The three built-in
types are each fully hand-written, so every new "thing to catalog" is a
multi-file PR. We need a runtime, schema-driven path for inert catalog types.

## Goals

- Let a `registry-admin` define a new entity type (name + field schema) at
  runtime, with no deploy.
- Persist records of custom types in MongoDB, governed by the existing
  RBAC/visibility model (public / private / group-restricted).
- Make custom records discoverable through the existing unified semantic search.
- Render list / detail / create / edit UI for custom types entirely from the
  schema — no per-type React code.

## Non-goals (v1)

- Custom entities are **catalog-only** — never proxied, executed, or
  health-checked by the gateway.
- No conversion of the three built-in types to schema (they keep their bespoke
  code and ingestion behavior).
- No federation of custom types.
- No edit history / versioning (destructive edits, matching the built-ins).
- No nested-object fields, entity references, file uploads, conditional fields,
  user-selectable sorting, or markdown rendering.
- Type descriptors are **immutable** in v1 (rename/restructure = delete +
  recreate). An additive-only field-add is a possible fast-follow.

## High-level approach

1. **Two new collections.**
   - `mcp_custom_types` — admin-authored type descriptors (`_id` = type name).
   - `mcp_custom_entities` — all records across all custom types in one
     collection, keyed by an `entity_type` discriminator; synthetic path
     `/{type}/{uuid}` is the `_id`. Single collection + discriminator matches
     the proven `mcp_embeddings_{dim}` pattern and keeps list code in one place.

2. **Descriptor-driven validation and UI.** A `CustomTypeDescriptor` lists
   typed fields (`string`, `text`, `number`, `bool`, `enum`, `date`,
   `array<string>`). The same descriptor drives server-side attribute
   validation and the generic React form/detail/card via a
   `datatype -> widget` switch.

3. **Shared envelope, typed payload.** Every record carries a uniform envelope
   (`name`, `description`, `visibility`, `allowed_groups`, `owner`, `tags`,
   `is_enabled`, timestamps) plus an `attributes` bag validated against the
   descriptor. `name`/`description` are promoted to the envelope so the list
   view, search text, and ordering have a stable display name regardless of the
   custom schema.

4. **Semantic search integration.** Records index into the existing
   `mcp_embeddings_{dim}` collection with `entity_type` set to the custom type
   name. Fields flagged `semantic` contribute to the embedding text; other
   fields contribute to keyword-search metadata. Custom records appear in the
   default "search everything" results; the existing soft-cap + RRF ranking
   prevents one large type from crowding out others.

5. **RBAC / visibility.** Type definition (create/delete) is admin-only. Record
   CRUD uses normal auth plus the existing visibility model. Custom-entity list
   and count apply an **in-query** visibility filter (a deliberate divergence
   from the built-ins' post-fetch filtering) so `total_count` reflects what the
   caller can actually see, while staying index-covered.

6. **Descriptor cache.** A small in-process cache fronts type descriptors with
   two read contracts: a best-effort, TTL'd read for hot paths (config tab list,
   default search scope) and an authoritative always-Mongo read on the write
   path (so a freshly created type can't false-404 on another replica).

## API surface

Type definitions (admin-only):

- `POST /api/custom-types` — define a type (name + fields). 409 on duplicate.
- `GET /api/custom-types` — list descriptors (any authenticated user; needed to
  build tabs/forms).
- `GET /api/custom-types/{name}` — full descriptor (the single source for the
  field schema).
- `DELETE /api/custom-types/{name}` — cascading delete of descriptor + records +
  embeddings; requires `?force=true` when records exist.

Records (normal auth + visibility filter):

- `POST /api/custom/{type}` — create a record; validates `attributes` against
  the descriptor.
- `GET /api/custom/{type}` — list with `limit`/`offset`; returns visible slice +
  matching `total_count`.
- `GET /api/custom/{type}/{uuid}` — single record (404, not 403, if not visible).
- `PUT /api/custom/{type}/{uuid}` — merge-then-validate update; re-indexes.
- `DELETE /api/custom/{type}/{uuid}` — delete record + its embedding.

Config:

- `GET /api/config` gains `features.custom_types` (master on/off) and a
  lightweight `custom_types` tab list (`name` + `display_name` only) so the
  frontend can build dynamic tabs.

All `{type}` / `{name}` / `{uuid}` path params are pattern-constrained at the
route signature as a NoSQL-injection guard before reaching any query.

## Configuration

| Variable | Default | Purpose |
|----------|---------|---------|
| `CUSTOM_ENTITY_TYPES_ENABLED` | `false` | Master switch (tabs + endpoints). Off = no behavior change for existing deployments. |
| `CUSTOM_TYPE_CACHE_TTL_SECONDS` | `60.0` | TTL for the in-process descriptor cache (best-effort reads). |
| `MAX_CUSTOM_RECORDS_PER_TYPE` | `0` | Soft cap per type (0 = unlimited); rejects create at cap. |

`CUSTOM_ENTITY_TYPES_ENABLED` must be propagated to all deployment surfaces
(Docker Compose, Terraform/ECS, Helm) and the unified parameter reference.

## Scope and deliverables

**Backend:** new models, descriptor-driven validator, two repositories, service
(validate + persist + index), search indexing extension, two routers, factory
wiring, config endpoint change, feature flag.

**Frontend:** generic `CustomEntityCard` / `CustomEntityDetail` /
`CustomEntityForm` components driven by the descriptor, dynamic Dashboard tabs
from the config tab list, and a `humanize`/`labelFor` helper. The descriptor is
fetched once per tab mount; record data is sourced from the shared stats context
(no per-tab refetch).

**Tests:** validator units (each datatype, enum, required, length bounds, date
rejection), model units (reserved names, caps), security-critical visibility
filter units (list filter and single-record check must agree), repository units
(entity_type scoping, filtered count/list, cascade delete), route units (admin
gate, param guards, mass-assignment protection, `total_count` correctness,
audit), and an end-to-end integration test.

**Docs/CLI:** unified parameter reference row, admin guidance that custom fields
are stored unencrypted (no secrets), and CLI commands in the registry client.

## Key constraints and risks

- **Custom fields are stored unencrypted** and emitted verbatim in API
  responses and search text. Admins must not define secret-bearing fields; this
  is an authoring-time constraint documented in the UI and docs, not enforced
  programmatically in v1.
- **Embedding collection is the scaling chokepoint.** All types share
  `mcp_embeddings_{dim}`; vector kNN runs before the `entity_type` match.
  Practical ceiling ~50-100k records per type and ~20-30 enabled custom types
  before search P99 degrades. The `custom_entities_total{entity_type}` gauge is
  the early-warning signal.
- **Last-write-wins on concurrent PUT** (no If-Match in v1). Merge-then-validate
  limits blast radius to two users editing the same attribute — the same gap the
  built-in types have. Optimistic concurrency deferred to v1.1.

## Kill switch / rollback

Setting `CUSTOM_ENTITY_TYPES_ENABLED=false` hides all custom-type tabs, disables
the endpoints, and excludes custom records from search — but **preserves all
data** in both collections. Re-enabling restores everything with no migration
and no data loss. This is the intended rollback path.

## Rollout

1. **Backend** — models, validator, repos, service, search indexing, routes
   behind the flag (default off); unit + integration tests green.
2. **Frontend** — generic components, dynamic tabs, manual UX pass (define a
   type as admin, CRUD records as a user).
3. **Deployment + docs** — propagate the flag to all surfaces and the unified
   reference, ship CLI, enable in staging, verify, then prod.

omrishiv:
*Issue #, if available:*

Fixes #1172 

*Description of changes:*

This PR introduces the ability to dynamically register custom entities and schemas in the registry. An admin can create any number of custom entities with a schema that's specific to the entity and admins and users can register custom entities based on that schema. 

Custom entities create their own tabs dynamically, support lexical/semantic search and voting, and register their own cards. 

image

image

image

image

image


aarora79:
Demo video: 

https://github.com/user-attachments/assets/3858e990-e26e-4d49-b408-1e64ca42e2fd

PR #1175, opened 2026-06-03
fix: allow empty metadata server fields defaults to {}

### LINKED ISSUE #1165: Server registration fails when metadata field is omitted
## Summary

The OpenAPI spec correctly marks `metadata` as an optional field for server registration. However, when the field is omitted from the request, the registration can fail because downstream code expects `metadata` to be a `dict`, not `None`.

## Root Cause

The `ServerInfo` Pydantic model at `registry/core/schemas.py:225` declares:
```python
metadata: dict[str, Any] = Field(default_factory=dict, ...)
```

This type annotation means `None` is not a valid value. When metadata is not provided in the registration form, the server entry dict does not include a `metadata` key. Depending on the storage backend and read path, this can result in `None` being assigned to the field, which fails Pydantic validation.

## Fix

Ensure that the registration flow in `server_routes.py` explicitly sets `server_entry["metadata"] = {}` when metadata is not provided, before persisting to storage. This guarantees the field is always a dict.

## Acceptance Criteria

- [ ] Server registration succeeds when `metadata` field is omitted entirely
- [ ] Server registration succeeds when `metadata` is provided as `{}`
- [ ] No regression for existing servers with metadata populated
- [ ] Test covering registration without metadata field

omrishiv:
*Issue #, if available:*

Fixes #1165 

*Description of changes:*

Defaults `metadata` fields to empty dicts, `{}`. Guards against `null` json metadata

### DIFF
--- a/registry/api/server_routes.py
+++ b/registry/api/server_routes.py
@@ -1035,6 +1035,7 @@ async def register_service(
         "allowed_groups": allowed_groups_list,
         "deployment": deployment,
         "registered_by": user_context["username"],
+        "metadata": {},
     }
 
     # Deployment-specific fields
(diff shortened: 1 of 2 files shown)


aarora79:
Thanks for landing this. Opened #1188 as a small follow-up: extends the test matrix to the other two register handlers, tightens the guard to also catch non-dict JSON (e.g. `metadata=[]` was still being stored), and mirrors the read-side default onto the single-GET so legacy rows without a `metadata` key match the list response.

Issue #987, opened 2026-05-09
feat(logging): migrate container logs to /var/log/containers/ai-registry/ with JSONL format for Splunk ingestion


aarora79:
Customers deploying the registry run Splunk universal forwarders configured to scrape files under `/var/log`. Today:

- The **registry** and **auth-server** containers write rotated log files to `/app/logs/{registry,auth-server}.log` inside the container, backed by either a named Docker volume (`docker-compose.yml`) or a bind mount to `${HOME}/mcp-gateway/logs` (`docker-compose.prebuilt.yml`, `docker-compose.podman.yml`).
- The **mcpgw** container (the MCP Gateway Server, [servers/mcpgw/server.py:21-25](../../servers/mcpgw/server.py#L21-L25)) logs **only to stdout** via `logging.basicConfig(...)`. There is no file handler and no log volume mount.
- Nginx inside the registry container writes to `/var/log/nginx/{access,error}.log` with no rotation and no host-side bind mount (229 MB of unrotated `error.log` observed).
- The Python log format is comma-separated plain text ([LOG_FORMAT in registry/utils/logging_setup.py:13](../../registry/utils/logging_setup.py#L13)): `%(asctime)s,p%(process)s,{%(filename)s:%(lineno)d},%(levelname)s,%(message)s`. Splunk's default CSV parser breaks on this format because the timestamp itself contains a comma (millisecond separator) and message text is not escaped.

Because the log files live inside a Docker named volume or an arbitrary home directory, the customer's Splunk forwarder sees nothing. They want the files to land under `/var/log/containers/` on the host, and they want a documented, parser-friendly log format standard.

### Proposed Solution

**1. Move all service log files to a single flat directory on the host:**

| Service | Host file path (same as in-container path) |
|---------|---------------------------------------------|
| registry | `/var/log/containers/ai-registry/registry.log` |
| auth-server | `/var/log/containers/ai-registry/auth-server.log` |
| mcpgw | `/var/log/containers/ai-registry/ai-registry-tools.log` |
| nginx access (in registry container) | `/var/log/containers/ai-registry/nginx-access.log` |
| nginx error (in registry container) | `/var/log/containers/ai-registry/nginx-error.log` |

[...]

PR #1201, opened 2026-06-05
macOS setup skill: add missing gettext prereq, fix /var/log bind mount, correct token file format

vrindabhandari:
### 1. Add `gettext` to the prerequisites check
The Phase 13 build needs `gettext` (for `envsubst` and related utilities), but Phase 1's prereq check doesn't verify it. On a fresh Mac, the build fails partway through with a non-obvious error. Adds the check and a `brew install gettext` install instruction alongside the other tools.

### 2. Fix Docker bind mount for macOS file-sharing
The `/var/log/containers/ai-registry` host bind mount used by `registry`, `auth-server`, and `mcpgw-server` isn't on Docker Desktop's default macOS file-sharing allowlist, so all three containers fail to start on a fresh Mac. Moves the host side to `${HOME}/mcp-gateway/logs` — same convention as the other `${HOME}/mcp-gateway/...` mounts in this file, and the directory is already created by Phase 7 of the skill. In-container path is unchanged so `APP_LOG_DIR` and app code stay the same.

### 3. Correct token file references in Phase 15/16
The `registry_management.py` CLI only accepts JSON token files (`*-token.json`), but the skill points `--token-file` at the `.env` env files. Phase 15 (Cloudflare server registration) and Phase 16 (verification) both fail when run literally. Updates the three `--token-file` references to use the JSON files generated alongside the `.env` ones by Phase 14.

### DIFF
--- a/.claude/skills/macos-setup/SKILL.md
+++ b/.claude/skills/macos-setup/SKILL.md
@@ -372,8 +372,13 @@ Log: `{ 6, "Embeddings Model Download", DONE/FAILED, "~/mcp-gateway/models/all-M
 **Announce:** "Creating Docker volume mount directories..."
 
 ```bash
-mkdir -p "${HOME}/mcp-gateway/{servers,models,auth_server,secrets/fininfo,logs,ssl}"
-ls -la "${HOME}/mcp-gateway/"
+# Splunk log directory (Issue #987). On Linux this exists by default
+# via the distro / log forwarder; on macOS we have to create it with
+# sudo since /var/log is root-owned.
+if [ ! -d /var/log/containers/ai-registry ]; then
+    sudo mkdir -p /var/log/containers/ai-registry
+    sudo chown "$(whoami)" /var/log/containers/ai-registry
+fis
 ```
 
 Log: `{ 7, "Directory Creation", DONE, "~/mcp-gateway/{servers,models,auth_server,secrets/fininfo,logs,ssl}" }`
--- a/.claude/skills/macos-setup/SKILL.md
+++ b/.claude/skills/macos-setup/SKILL.md
@@ -140,6 +140,9 @@ git --version 2>/dev/null && echo "GIT_OK" || echo "GIT_FAIL"
 
 echo "=== jq ==="
 jq --version 2>/dev/null && echo "JQ_OK" || echo "JQ_FAIL"
+
+echo "=== gettext ==="
+gettext --version 2>/dev/null && echo "GETTEXT_OK" || echo "GETTEXT_FAIL"
 ```
 
 For any failed check, display the install instructions:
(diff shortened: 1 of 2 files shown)


aarora79:
@vrindabhandari Thanks for catching the missing `gettext` prereq and the token-file format fix — both are good catches.

However, the `docker-compose.yml` change in this PR is a regression for everyone running on Linux. The path `/var/log/containers/ai-registry` is the convention for Splunk-style log ingestion (see Issue #987), and Linux deployments rely on it being a real host path that log forwarders watch. Switching the bind mount to `${HOME}/mcp-gateway/logs` silently breaks that integration with no warning.

The right place to fix the macOS-specific failure is in the macOS setup skill itself, not in `docker-compose.yml`. Could you:

1. **Revert** all three `docker-compose.yml` changes (lines 321, 503, 584) so the bind mount stays `/var/log/containers/ai-registry:/var/log/containers/ai-registry`.

2. **Add directory creation in `.claude/skills/macos-setup/SKILL.md`**, in **Phase 7: Create Required Directories** (right after the existing `mkdir -p "${HOME}/mcp-gateway/{...}"` block around line 375). Something like:

   ```bash
   # Splunk log directory (Issue #987). On Linux this exists by default
   # via the distro / log forwarder; on macOS we have to create it with
   # sudo since /var/log is root-owned.
   if [ ! -d /var/log/containers/ai-registry ]; then
       sudo mkdir -p /var/log/containers/ai-registry
       sudo chown "$(whoami)" /var/log/containers/ai-registry
   fi
   ```

That keeps the macOS fix local to the macOS skill and leaves the Linux/Splunk path unchanged. The `gettext` prereq and `agent-test-agent-m2m-token.json` filename fixes can stay as-is — both are correct.

vrindabhandari:
Thanks, that makes sense - I've reverted all three docker-compose.yml lines back to the original bind mount and added the sudo mkdir/chown block to Phase 7 of the macOS skill.

aarora79:
The Phase 7 change in this PR replaces the `${HOME}/mcp-gateway/...` directory tree with a `sudo mkdir` for `/var/log/containers/ai-registry`. Two problems with that:

1. It deletes the `mkdir -p "${HOME}/mcp-gateway/{servers,models,auth_server,secrets/fininfo,logs,ssl}"` line that Phase 6 (embeddings model download) and the compose bind mounts (`docker-compose.yml` lines 332-337) depend on. Removing it breaks those later phases on a fresh Mac.

2. The `/var/log/containers/ai-registry` directory is already created with the correct ownership by `scripts/prepare-log-dirs.sh`, which `build_and_run.sh` (line 490) invokes automatically during Phase 13. That script chowns to `1000:1000` (the container `appuser` uid). The `sudo chown "$(whoami)"` added here sets the wrong owner (a macOS user is typically uid 501, not 1000), so the containers still can't write to the directory. The skill also doesn't need `sudo` for this at all, since `prepare-log-dirs.sh` handles it.

Posting a follow-up commit directly to main that restores Phase 7 to:

```bash
mkdir -p "${HOME}/mcp-gateway/{servers,models,auth_server,secrets/fininfo,logs,ssl}"
ls -la "${HOME}/mcp-gateway/"
```

No handling of `/var/log/containers/ai-registry` is needed in the skill: `build_and_run.sh` -> `prepare-log-dirs.sh` creates it with the right uid/gid in Phase 13. Fixes #1 (gettext prereq) and #3 (token JSON format) from this PR are correct and stay.

PR #1203, opened 2026-06-07
docs: add dynamic discovery receipt guidance

caioribeiroclw-pixel:
## Summary

Adds a compact "Discovery Receipts and Context Budgets" section to the dynamic tool discovery docs. The section describes what operators should capture after dynamic tool discovery: query/correlation IDs, exposed tools, withheld candidates, invocation outcome, redacted argument/result shapes, and stop reason.

Why: dynamic discovery reduces upfront MCP/tool context bloat, but production and multi-tenant operators still need an audit-friendly way to review what the agent actually saw without copying raw payloads into logs.

## Checks

- `git diff --check`
- lightweight doc smoke verifying the new TOC anchor and JSON example are present

Docs-only change; I did not run the full app test suite.

### DIFF
--- a/docs/dynamic-tool-discovery.md
+++ b/docs/dynamic-tool-discovery.md
@@ -6,6 +6,7 @@ The MCP Gateway & Registry provides a powerful **Dynamic Tool Discovery and Invo
 
 - [Overview](#overview)
 - [How It Works](#how-it-works)
+- [Discovery Receipts and Context Budgets](#discovery-receipts-and-context-budgets)
 - [Architecture](#architecture)
 - [Usage Examples](#usage-examples)
 - [Agent Integration](#agent-integration)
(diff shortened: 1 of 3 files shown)


aarora79:
Hi @caioribeiroclw-pixel, are you proposing to add this as a functionality? I only see the md file being included in the PR, did you intend to commit code for this as well?

caioribeiroclw-pixel:
Thanks for the nudge — I intended the original PR as docs/spec only, but you were right that it was ambiguous.

I pushed a small implementation commit now (`3dc12acd`) so the PR is concrete:

- `search_registry` returns a compact `discovery_receipt` on successful semantic search responses
- deprecated `intelligent_tool_finder` returns the same shape for compatibility
- receipt records query, max-results limit, exposed tool names/paths/scores, withheld candidate count, status, and stop reason
- added unit coverage for exposed/withheld accounting and keeping the helper to metadata/counts rather than raw args/results
- updated the doc wording to say the receipt is returned with the search response and can be persisted by operators alongside audit logs

Local checks I could run in this environment:

```bash
python3 -m py_compile servers/mcpgw/server.py tests/unit/servers/mcpgw/test_intelligent_tool_finder.py
git diff --check
```

I could not run the pytest target locally because this checkout does not have `pytest` installed, so I left the unit tests in the PR for CI/maintainer validation. If you prefer this to stay docs-only, I can split the code into a follow-up PR instead.

caioribeiroclw-pixel:
Small follow-up after self-review: I noticed `search_registry` was counting withheld candidates from returned entity groups instead of the actual candidate tool list, so a server with multiple `matching_tools` could under-report withheld tools.

Pushed `13ead8a9` to fix that accounting and added tests for both nested `servers[].matching_tools` and top-level `tools[]` search results.

Checks I could run here remain:

```bash
python3 -m py_compile servers/mcpgw/server.py tests/unit/servers/mcpgw/test_intelligent_tool_finder.py
git diff --check
```

I still cannot run pytest in this checkout because the environment is missing pytest/project deps, but the new tests are in the PR for your CI/local validation.

aarora79:
Thanks @caioribeiroclw-pixel , will review tomorrow and merge if all looks good.

aarora79:
Thanks for this, the "withheld candidates as a first-class signal" idea is genuinely useful, especially for evals and agent tuning where you want to know whether the right tool got starved out by a tight `top_n`. A few things I'd like us to align on before this merges:

**1. Let's make the receipt opt-in rather than always-on.**
The consumer of an MCP tool response is the agent/LLM, and most production calls don't need this metadata, it just adds tokens to the model-facing response for data the model won't use. Since the real value here is for eval / agent-development workflows, I'd propose gating it behind an optional parameter that defaults to off:

```python
async def search_registry(
    query: str,
    max_results: int = 10,
    include_discovery_receipt: bool = False,
    ctx: Context | None = None,
) -> dict[str, Any]:
```

When `False` (the default), the response is byte-for-byte what it is today, so this becomes a purely additive, backwards-compatible change. When `True`, the eval harness gets the receipt. Same treatment for `intelligent_tool_finder`.

**2. Separate audit from this eval signal.**
The doc section frames receipts as something "operators can persist alongside audit logs." If operator audit is a goal, that data should be emitted server-side (structured `logger.info()` audit line or an OTel span attribute/event), not returned to the caller, since the model can't be trusted to persist it and shouldn't be the audit transport. The repo already has OTel-native metrics and `@track_tool()` wired up, so audit is better served there. I'd suggest narrowing the doc to position the receipt as an eval/agent-dev signal, and handle operator audit (if we want it) as a separate server-side change.

**3. One correctness nit in `search_registry`.**
`exposed_tools` is built only from `tools` + servers' `matching_tools`, but `total_results` (and the actual exposed surface) also counts `agents` and `skills`. So `withheld.candidate_tool_count` can be off whenever agents/skills are in the result set. Worth either including them in the candidate set or documenting that the receipt is tools-only.

If we go with the opt-in parameter, this is a clean, additive feature and I'm happy to approve. Happy to pair on the flag wiring if useful.

caioribeiroclw-pixel:
Thanks — agreed on all three points. I pushed `14ae46a3` to align the PR with that shape:

- added `include_discovery_receipt: bool = False` to both `search_registry` and the deprecated `intelligent_tool_finder`
- default responses now omit `discovery_receipt`, so production/model-facing calls stay unchanged unless an eval/dev caller opts in
- narrowed the docs: receipt is positioned as an eval / agent-development debugging signal, not operator audit transport; production audit should be server-side logs/OTel
- changed the receipt from tool-only accounting to result-surface accounting (`exposed_results`, `candidate_result_count`) so agents/skills are counted alongside tools instead of under-reporting withheld candidates
- added tests for opt-in behavior and agents/skills accounting

Checks I could run here:

```bash
python3 -m py_compile servers/mcpgw/server.py tests/unit/servers/mcpgw/test_intelligent_tool_finder.py
git diff --check
custom test harness for tests/unit/servers/mcpgw/test_intelligent_tool_finder.py: 13 tests passed
```

This environment does not have `pytest`, `pip`, or project deps available, so I used the existing lightweight stubs in that test module plus local stubs for missing runtime imports to execute the unit test functions directly. GitHub has not reported checks on the new fork commit yet from what I can see.

aarora79:
Follow-up: deduped discovery candidates and itemized the withheld set (`top_withheld`) in #1227 (merged).

PR #1209, opened 2026-06-09
fix(auth0): configurable username claim with lowercase normalisation

### LINKED ISSUE #930: Valid group-based admin scopes can still result in empty UI permissions
### Summary

Even after the correct Entra admin group is resolved and mapped to admin scopes, the Registry can still return empty `ui_permissions`.

### Impact

The user receives valid scopes such as:
- `registry-admins`
- `mcp-servers-unrestricted/read`
- `mcp-servers-unrestricted/execute`

but still cannot access the Register page because `register_service` is missing from `ui_permissions`.

### Actual behavior

The auth context contains valid admin scopes, but:
- `ui_permissions: {}`

The UI then shows:
- "You do not have permission to register servers or agents."

### Expected behavior

If a user has a valid admin scope such as `registry-admins`, the Registry should consistently derive the corresponding UI permissions, including `register_service`.

### Notes

This looks like either:
- repository-backed scope/UI data is not being loaded consistently, or
- the product needs a more robust fallback / failure mode when scope documents are incomplete.

srinivas-kannepalli:
## Problem

Auth0 provider hardcodes `nickname` as the username for session storage and DB group lookups.

For **enterprise federations** (ADFS → Okta → Auth0) the `nickname` claim contains only the local-part of the email address and preserves the original directory capitalisation (e.g. `Srinivas.Kannepalli`). This fails to match lower-cased full-email addresses stored in the `idp_user_groups` collection, so users always fall back to read-only access even after their groups are seeded.

Closes #930 (empty `ui_permissions` for enterprise federation users).

## Root cause

`auth0.py` ignores the `username_claim` field already present in `oauth2_providers.yml` and uses a hardcoded lookup chain (`nickname or email or sub`).  All other providers (keycloak, okta, entra, pingfederate) honour `provider_config["username_claim"]`.

### `oauth2_providers.yml`
- Auth0 `username_claim` default changed from `nickname` → `${AUTH0_USERNAME_CLAIM:-email}`
- `email` is the most portable default; works with both simple and enterprise Auth0 tenants
- Set `AUTH0_USERNAME_CLAIM=nickname` to restore previous behaviour

### `server.py` – `substitute_env_vars()`
- Generalised to support `${VAR:-default}` syntax for **any** variable in the YAML
- Replaces the one-off Cognito-domain special case; backwards-compatible

### `providers/auth0.py`
- `Auth0Provider.__init__` accepts a new `username_claim: str = "email"` parameter
- `validate_token` and `extract_user_from_tokens` now use `self.username_claim` + `.lower()`
- Prevents case-sensitivity mismatches regardless of upstream IdP capitalisation

### `providers/factory.py`
- Reads `AUTH0_USERNAME_CLAIM` env var (default `"email"`) and passes it to `Auth0Provider`

## Backwards compatibility

Deployments without `AUTH0_USERNAME_CLAIM` switch from `nickname` → `email` as the username key. Operators who relied on nickname-keyed sessions/DB records should:
1. Set `AUTH0_USERNAME_CLAIM=nickname`  *or*
2. Re-run their user-group seed job with the new username format

## Testing

- Tested against a Cimpress Auth0 tenant (enterprise ADFS→Okta→Auth0 federation)
- `email` claim produces `srinivas.kannepalli@vista.com` after `.lower()`, matching Cortex-seeded DB records
- Group enrichment now succeeds and `ui_permissions` is populated correctly

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -720,6 +720,14 @@ AUTH0_CLIENT_SECRET=your_auth0_client_secret_here
 # Default: https://mcp-gateway/groups
 AUTH0_GROUPS_CLAIM=https://mcp-gateway/groups
 
+# Auth0 username claim — the ID-token claim used as the canonical username for
+# session storage and idp_user_groups lookups.
+# Default: nickname (preserves existing behaviour for simple Auth0 tenants).
+# Set to "email" for enterprise federations (ADFS → Okta → Auth0) where
+# "nickname" contains only the local-part of the address in the upstream
+# directory's casing, causing group-lookup mismatches.
+AUTH0_USERNAME_CLAIM=nickname
+
 # Enable Auth0 in OAuth2 providers
 AUTH0_ENABLED=false
 
--- a/auth_server/oauth2_providers.yml
+++ b/auth_server/oauth2_providers.yml
@@ -102,10 +102,9 @@ providers:
     scopes: ["openid", "email", "profile"]
     response_type: "code"
     grant_type: "authorization_code"
-    # Claims mapping for user info
-    # Auth0 uses 'nickname' for display name and requires a custom
-    # Rule/Action to add groups to tokens as a namespaced claim
-    username_claim: "nickname"
+    # AUTH0_USERNAME_CLAIM selects which ID-token claim becomes the username
+    # used for session storage and DB group lookups.  Defaults to "nickname".
+    username_claim: "${AUTH0_USERNAME_CLAIM:-nickname}"
     groups_claim: "${AUTH0_GROUPS_CLAIM}"
     email_claim: "email"
     name_claim: "name"
--- a/auth_server/providers/auth0.py
+++ b/auth_server/providers/auth0.py
@@ -45,6 +45,7 @@ class Auth0Provider(AuthProvider):
         m2m_client_id: str | None = None,
         m2m_client_secret: str | None = None,
         groups_claim: str = "https://mcp-gateway/groups",
+        username_claim: str = "nickname",
     ):
         """Initialize Auth0 provider.
 
@@ -59,6 +60,9 @@ class Auth0Provider(AuthProvider):
                 Auth0 requires a namespaced claim via a Rule/Action
                 (e.g., 'https://mcp-gateway/groups'). Defaults to
                 'https://mcp-gateway/groups'.
+            username_claim: ID-token claim used as the canonical username for
+                session storage and DB group lookups.  Defaults to
+                ``"nickname"``.  Set ``AUTH0_USERNAME_CLAIM`` to override.
         """
         self.domain = domain.rstrip("/")
         self.client_id = client_id
@@ -67,6 +71,7 @@ class Auth0Provider(AuthProvider):
         self.m2m_client_id = m2m_client_id or client_id
         self.m2m_client_secret = m2m_client_secret or client_secret
         self.groups_claim = groups_claim
+        self.username_claim = username_claim
 
         # JWKS cache
         self._jwks_cache: dict[str, Any] | None = None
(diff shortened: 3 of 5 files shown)


aarora79:
Thanks @srinivas-kannepalli. The diagnosis is correct: Auth0 was the only provider that ignored `username_claim`, and `nickname` carries just the local-part with the upstream directory's casing, so it misses the lower-cased email keys in `idp_user_groups`.

Two CI items to fix first:

- Ruff fails: `F401 string.Template imported but unused` at `auth_server/server.py:23`. The `substitute_env_vars` rewrite removed the last use of `Template`. Run `uv run ruff check auth_server/ --fix`.
- One test fails: `tests/auth_server/unit/providers/test_auth0.py::TestAuth0TokenValidation::test_validate_token_success` asserts `username == "testuser"`, but the new default returns `testuser@example.com`. Update it to match the default we agree on below.

## Default behavior

We want to keep `nickname` as the default and make `email` opt-in.

The username is the identity key. It keys the session and the `idp_user_groups` lookup. Switching the default from `nickname` to `email` re-keys every existing Auth0 user on upgrade. Their seeded group records use the old key, the lookup misses, and they drop to read-only until the operator sets `AUTH0_USERNAME_CLAIM=nickname` or re-seeds. That turns a fix for enterprise federations into a breaking change for everyone else.

A number of existing deployments run Auth0 as the IdP. That is the set this default would break on upgrade.

Proposed split:

1. Keep the lowercase normalization on by default. It is safe for any claim and it fixes the casing half of #930 (`Srinivas.Kannepalli` to `srinivas.kannepalli`) with no migration. Keep it as you have it.

2. Default the claim to `nickname`; make `email` opt-in. In `auth_server/oauth2_providers.yml`, under the `auth0` provider, set:

   ```yaml
   username_claim: "${AUTH0_USERNAME_CLAIM:-nickname}"
   ```

   and default `username_claim` to `"nickname"` in `_create_auth0_provider()` (`auth_server/providers/factory.py`) and in `Auth0Provider.__init__` (`auth_server/providers/auth0.py`). Enterprise federations set `AUTH0_USERNAME_CLAIM=email` to get the full-email key.

Result: simple tenants upgrade with no change and no migration. Enterprise federations set one env var, re-seed once, and the #930 case works.

## Smaller items

- Add `AUTH0_USERNAME_CLAIM` to `.env.example` next to `AUTH0_GROUPS_CLAIM`, and to `docs/unified-parameter-reference.md` section 12e. State the default (`nickname`) and that `email` is the setting for enterprise IdP federations. If Auth0 runs on ECS/Helm, wire it through those surfaces or note that it is auth-server env only.
- Move `import re` to module scope in `server.py` instead of importing it inside `substitute_env_vars` on every call.
- Add a test for `substitute_env_vars`. It now branches across `${VAR}`, `${VAR:-default}`, unset-without-default, and `COGNITO_DOMAIN:-auto`, and it parses every provider value at startup with no coverage.
- Add a What's New line for the new `AUTH0_USERNAME_CLAIM` knob and the lowercase normalization.

The `${VAR:-default}` generalization in `substitute_env_vars` is a good cleanup and removes the one-off Cognito special case. That can land as-is.

The approach is right. We want `email` opt-in via `AUTH0_USERNAME_CLAIM` so existing Auth0 installs upgrade without losing access. Thanks for the contribution.

aarora79:
Moving this back to draft for now, I'm not sure it's ready for review yet. Please mark it ready again once you've finished the changes.

Also, the Anthropic API changes (`/v0.1/servers` to `/api/servers` and the related CLI commits) don't belong in this PR, and I'm not sure they're needed. Let's keep this PR scoped to the Auth0 username-claim fix and handle those separately if they turn out to be necessary.

aarora79:
Closing due to inactivity, please feel free to reopen with comments incorporated.

PR #1224, opened 2026-06-12
feat(connect-config): per-server OAuth client_id login for Cursor/Claude/Codex + /mcp path override

go-faustino:
## Summary

When a deployment runs an OAuth provider but has **anonymous Dynamic Client Registration disabled** (common with enterprise Keycloak), the Connect dialog can't show an IDE login button — it falls back to embedding a static gateway token. This PR lets operators advertise a **pre-registered public OAuth client_id** so the Connect dialog emits a token-less, login-button config instead, for the IDEs that support a fixed public client.

Two levels of configuration:

- **Registry-wide default:** `IDE_OAUTH_CLIENT_ID` (`ide_oauth_client_id`). Empty by default → unchanged static-token behavior.
- **Per-server override:** `oauth_client_id` on the server entry, which wins over the global default for that server.

Resolution happens **server-side** in `GET /api/servers/{path}/connect-config` (`server_info.oauth_client_id or settings.ide_oauth_client_id`), so the frontend receives a single resolved value.

When a client id resolves for a server, the Connect dialog drops the static gateway token and emits OAuth-aware output per IDE:

| IDE | Output |
|-----|--------|
| Cursor (JSON) | `"auth": { "CLIENT_ID": "<id>" }` |
| Claude Code | `claude mcp add --transport http --client-id <id> <name> <url>` |
| Codex | `codex mcp add <name> --url <url> --oauth-client-id <id>` |

Other IDEs (Roo Code / Kiro / Goose / VS Code default / curl) are intentionally left on the existing behavior, since they have no verified fixed-public-client OAuth config syntax.

### Also: per-server `/mcp` path override

Added `append_mcp_path: bool | null` on the server entry to control the trailing `/mcp` transport segment of the gateway Connect URL:

- `null` (default) — auto-detect from `proxy_pass_url` (unchanged behavior).
- `false` — strip `/mcp` for root-endpoint servers (e.g. AWS Knowledge) that serve MCP at the server path itself and break on `/mcp`.
- `true` — force the suffix.

The five duplicated inline URL builders in the Connect modal were consolidated into a single `buildConnectUrl()` helper that honors this override.

## Changes

- `registry/core/schemas.py` — new `oauth_client_id` + `append_mcp_path` fields on `ServerInfo`.
- `registry/api/server_routes.py` — `connect-config` resolves the effective client id (per-server ‖ global) and returns both fields.
- `frontend/src/components/ServerConfigModal.tsx` — consume the connect-config fields; `buildConnectUrl()` refactor; OAuth output for Cursor/Claude/Codex with token drop; UI hints.
- Reverted the earlier global-only `/api/config` + `useRegistryConfig` plumbing (resolution now lives server-side in connect-config).
- Docs: `docs/unified-parameter-reference.md` (global default + per-server overrides table), `docs/ai-coding-assistants-setup.md` (per-IDE OAuth examples + `append_mcp_path`).
- Tests: `frontend/.../ServerConfigModal.test.tsx` (OAuth + append_mcp_path via mocked connect-config; kept URL-generation tests), `tests/unit/api/test_config_ide_oauth.py` (model fields, admin-config group, env export).

## Notes

- The client id is **public** (not a secret); the registered public client must allow the loopback redirect URIs each IDE uses.
- Backward compatible: with no global setting and no per-server value, the Connect config is byte-for-byte unchanged (static token).

## Test plan

- [x] `tsc --noEmit` clean
- [x] Frontend `jest` ServerConfigModal suite (8/8): URL generation, append_mcp_path strip/force, Cursor OAuth emit/omit, registry-only never-OAuth
- [x] `ServerInfo` round-trips the new fields (not stripped by validation)
- [x] Backend unit suite (CI) — `tests/unit/api/test_config_ide_oauth.py`


aarora79:
Thanks @go-faustino for submitting this, will test this with a few IdPs and then merge it. I have a few follow-ups to this which will add as a separate PR. Thanks for the super useful contribution.

aarora79:
Thanks for this contribution, @go-faustino — the per-server `oauth_client_id` and `/mcp` path override were a great foundation. We've now built on it in a much bigger way in #1235 (merged): the IDE OAuth login flow is wired across all deployment surfaces (Docker, Compose variants, Terraform, Helm), validated end-to-end on Keycloak, Okta, and Amazon Cognito, and rounded out with a fixed callback-port option (`IDE_OAUTH_CALLBACK_PORT`) for IdPs that match redirect URIs literally — plus a few pre-existing auth bug fixes surfaced along the way (CSRF validation, PRM scope advertisement, Cognito access-token validation).

The three connection methods are now documented in [`docs/connection-methods/`](https://github.com/agentic-community/mcp-gateway-registry/tree/main/docs/connection-methods) — pre-registered client id, DCR, and CIMD (roadmap) — and per-IdP setup scripts live in [`setup/idp/<idp>/`](https://github.com/agentic-community/mcp-gateway-registry/tree/main/setup/idp). There's also a quick-start FAQ: [How do I get my AI coding assistant to work with this registry?](https://github.com/agentic-community/mcp-gateway-registry/blob/main/docs/faq/connect-ai-coding-assistant.md)

Appreciate you kicking this off!

go-faustino:
Thanks @aarora79 — glad it was a useful starting point, and #1235 looks like the right way to round it out.

For context: we're running the registry in front of an enterprise Keycloak with anonymous DCR disabled, so the pre-registered public client_id path is exactly the gap we needed for Cursor/Claude/Codex. Two of your follow-ups line up directly with that setup — the verify_csrf_token_header_only fix (silent fallback to the static token on cookie-authenticated GETs is a nasty one) and the PRM derive_supported_scopes() fix, since Keycloak rejects unknown scopes with invalid_scope. Defaulting to basic OIDC scopes with MCP_ADVERTISED_SCOPES as the override is the right call.

We deploy via our own downstream chart rather than the in-repo surfaces, but the per-IdP scripts and connection-methods/ docs are a great reference. Appreciate you building on it so quickly.

PR #1260, opened 2026-06-17
fix: validate signed jwt for mcp-proxy

omrishiv:
*Issue #, if available:*

*Description of changes:*

This PR enforces that requests sent to the mcp-proxy have a valid, signed JWT token. For existing deployments, it is recommended to update the registry first before updating the auth-server, that way requests are signed before arriving at the auth-server. If the reverse happens, the auth-server will reject the unsigned requests until the registry is updated. 


omrishiv:
Registry test fix is in #1258

PR #1263, opened 2026-06-19
feat(agents): on-demand A2A agent card pull with dry-run preview (#1030)

### LINKED ISSUE #1030: Add on-demand A2A agent card pull with dry-run preview and overwrite
## Problem Statement

When an A2A agent is registered in the registry, its metadata (name, description, version, skills, capabilities, security schemes, etc.) is captured at registration time and becomes static. If the remote agent updates its card (adds skills, changes version, updates provider info), the registry has no way to pull the latest card and update the local record.

Today, the health check endpoint (`POST /agents/{path}/health`) already fetches `/.well-known/agent-card.json` but discards the response body, only checking reachability. We should reuse this mechanism to also parse and optionally apply the remote card data.

## Proposed Solution

Add a new backend API endpoint `POST /api/agents/{path}/pull-card` that:

1. Fetches the A2A agent card from `/.well-known/agent-card.json` on the agent's host.
2. In **dry-run mode** (default): returns a diff of what would change (field-by-field comparison).
3. In **overwrite mode**: applies the remote card fields to the local record (only A2A-spec fields, preserving registry-specific metadata).

The UI adds a "Pull Card" icon button on the AgentCard component. Clicking it calls the API in dry-run mode first, shows a diff/preview modal, and lets the user confirm to apply changes.

## User Stories

- As a registry operator, I want to pull the latest A2A agent card so that the registry reflects the agent's current capabilities.
- As a registry operator, I want to preview what would change before applying, so I don't accidentally overwrite custom metadata.
- As a registry operator, I want the pull to only update A2A-spec fields (name, description, version, skills, capabilities, etc.) and preserve registry-specific fields (tags, visibility, trust_level, ratings, etc.).

## Acceptance Criteria

- [ ] `POST /api/agents/{path}/pull-card` endpoint exists with `dry_run` query parameter (default: true).
- [ ] Dry-run response includes: remote card data, field-by-field diff (current vs remote), and list of fields that would change.
- [ ] Overwrite mode (`dry_run=false`) updates only A2A-spec fields, preserving registry metadata.
- [ ] UI shows a "Pull Card" icon on each agent card (next to existing health check button).
- [ ] Clicking the icon opens a preview modal showing the diff.
- [ ] User can confirm to apply changes from the modal.
- [ ] Endpoint respects permissions (modify_service permission, owner or admin).
- [ ] Federated / read-only agents cannot be overwritten.
- [ ] Agents without a valid URL or non-A2A agents show an appropriate error.
- [ ] Health check is updated as a side effect of a successful pull.

## Out of Scope

- Automatic / scheduled pulling (this is on-demand only).
- Pulling from non-standard endpoints (only `/.well-known/agent-card.json`).
- Conflict resolution for concurrent edits (last write wins, same as existing update endpoint).

## Dependencies

- None (builds on existing health check infrastructure).

## Related Issues

- Health check endpoint already fetches `/.well-known/agent-card.json` but discards body.

## Design Documents

Detailed design lives under `.scratchpad/a2a-card-pull/` (LLD, expert review, and testing plan).

vrindabhandari:
*Issue #, if available:* Closes #1030

*Description of changes:* 

Adds `POST /api/agents/{path}/pull-card`:
- **Dry-run (default):** fetches the remote A2A card from `/.well-known/agent-card.json` and returns a field-by-field diff (current vs remote) without writing.
- **Apply (`dry_run=false`):** updates only A2A-spec fields, preserving registry-managed metadata (tags, ratings, visibility, trust_level, sync_metadata, etc.).
- Reuses the existing health-check fetch path; a successful pull also refreshes health status in the same write.
- **UI:** a "Pull Card" button on owned, non-federated A2A agents that opens a preview modal showing the diff, with an Apply action.

Design-review items folded in: 1 MiB cap on the remote read (S1), registrant-only protection via the shared `REGISTRANT_ONLY_FIELDS` contract (S5), a single merged `update_agent` write (P1), a structured `pull_card` log line (R4), and a warning when the remote would change the agent URL (S3).

**Testing**
- `tests/unit/api/test_pull_card_helpers.py` — normalize / diff / safe-updates / fetch + the A2A↔registrant disjoint invariant.
- `tests/unit/api/test_pull_card_endpoint.py` — dry-run, apply, permission/federated/protocol/URL guards, single-write, and registrant-only protection.
- All 35 pass locally.

**Notes for reviewer (a few conscious calls)**
- Endpoint tests use the existing `unittest.mock` style rather than respx, since respx isn't a project dependency. A respx + live-Mongo integration variant can be a follow-up.
- Last-write-wins (no `If-Match`), matching current PUT/PATCH. The LLD flagged If-Match as an open decision — happy to add optimistic concurrency if preferred.
- The local pre-commit `mypy` and `python-syntax` hooks flag two pre-existing items unrelated to this change (`BatchItemOp.replace` shadowing `str.replace`, and a `python` executable not on PATH). Left untouched to keep the diff scoped.


aarora79:
Hey @vrindabhandari, took this for a full test ride today and pushed four follow-up commits onto the branch (PR head now `17bc8f0c`). Heads-up: I force-pushed (with lease) — if you have local work on this branch, you'll want to `git fetch origin && git reset --hard origin/feat/1030-a2a-pull-card`.

What's on top of your original commit:

1. **`9249b508` — docstring + skills null-key spurious diff.** Two reviewer-feedback follow-ups in one commit:
   - Docstring on `pull_agent_card` now mentions the `dry_run=True` health-fields side effect (was only in the PR description, not OpenAPI-visible).
   - `_compute_card_diff` was reporting a spurious `skills` change on real A2A agents. Reproduced live on a real card: local `model_dump()` emits `examples` / `input_modes` / `output_modes` / `security` as explicit nulls, the remote omits them — `dict != dict` even though the lists are semantically equal. The per-field `hasattr(s, "model_dump")` check at line 359 was dead code (by then `s` is already a plain dict from the top-level `model_dump()` at line 344, so the `else s` branch ran and the nulls survived). Fix drops `None`-valued keys per field on both sides for `provider`, each `skill`, each `security_scheme` value, and sorts `skills` by stable `id` so a reordered remote also doesn't false-positive. All 35 existing tests still pass.

2. **`7242a862` — CLI parity.** New `agent-pull-card` subcommand on `registry_management.py` mirroring the new endpoint, with `--apply` (default dry-run) and `--json`. Mirror models in `registry_client.py` (`PullCardFieldChange`, `PullCardResponse`) and a `pull_card_agent()` client method.

3. **`17bc8f0c` — README blurb in What's New.** One-sentence summary.

Plus your original was rebased onto current `main` so the merge is clean.

End-to-end testing done:
- Positive path: UI dry-run + Apply on a real A2A agent (Jewel Homes) — diff renders, applied, registry-managed fields untouched.
- Negative paths: dead remote (DNS fail), upstream 502 — both surface as clean HTTP 502 from the endpoint with the right detail strings.
- CLI: dry-run, `--apply`, `--json`, error exit code on dead remote — all clean.

Filed #1282 as a follow-up for SSRF hardening on the shared `_fetch_remote_agent_card` / health-check fetch path — not a #1263 regression (same surface existed pre-PR), but worth tightening. Full code change + acceptance criteria are in the issue if you'd like to pick it up.

Thanks for the PR!

Issue #889, opened 2026-04-25
Bearer token TTL: MAX_TOKEN_LIFETIME_HOURS not enforced, and request for per-MCP max lifetime at registration


WPrintz:
## Summary

Two related issues with the `/api/tokens/generate` endpoint that affect the bearer token issued by the "Connect" button on the registry UI:

1. **Bug**: `MAX_TOKEN_LIFETIME_HOURS` is declared but never enforced, and `expires_in_hours` in the request payload is ignored on the `oauth2` / `network-trusted` code path — the lifetime is hardcoded to 8 hours regardless of what the caller asked for.
2. **Feature request**: allow a **per-MCP-server maximum token lifetime** set at server registration time, so operators can allow long-lived tokens for low-risk servers (e.g. public docs search) while keeping sensitive servers on short TTLs.

Background context: for IDE-based MCP usage (Claude Code, Cursor, Kiro, etc.), an 8-hour bearer token means users must click Connect, copy the JSON, paste into `mcp.json`, and restart their IDE multiple times per day. MCP clients don't hot-reload config, so there's no mitigation short of longer-lived tokens or an out-of-band refresh mechanism.

---

### Current behavior

In [`auth_server/server.py`](https://github.com/agentic-community/mcp-gateway-registry/blob/main/auth_server/server.py):

```python
# Lines 75-76
MAX_TOKEN_LIFETIME_HOURS = 24
DEFAULT_TOKEN_LIFETIME_HOURS = 8
```

```python
# Lines 965-971 - request model accepts an override
class GenerateTokenRequest(BaseModel):
    user_context: dict[str, Any]
    requested_scopes: list[str] = []
    expires_in_hours: int = DEFAULT_TOKEN_LIFETIME_HOURS   # accepts override from caller
    description: str | None = None
```

```python
# Line 2219 - ignores the request value
if auth_method in ("oauth2", "network-trusted"):
    ...
    expires_in = DEFAULT_TOKEN_LIFETIME_HOURS * 3600  # always 8 hours, ignores request.expires_in_hours
```

`MAX_TOKEN_LIFETIME_HOURS = 24` is declared but never referenced anywhere in the codebase, so it enforces nothing.

### Impact

- Callers of `POST /api/tokens/generate` that pass `expires_in_hours: N` silently get 8h regardless
- The frontend **"Token Ready"** modal shown on Connect always says "expires in 8 hours" because the backend can't return anything else
[...]

Issue #959, opened 2026-05-07
Self-signed user tokens include full groups claim even though authorization is scope-based, causing oversized/malformed bearer tokens


zoltan-fedor:
## Summary

The auth server generates self-signed user tokens with the full `groups` claim embedded, but the authorization path for those same self-signed tokens uses the embedded `scope` claim directly rather than mapping groups to scopes.

In environments where users belong to many groups, this can make the generated bearer token unnecessarily large and fragile. We hit a production failure where programmatic MCP calls returned an auth failure and the auth-server logs showed:

- `Invalid token - Invalid payload padding`

The root cause was not incorrect scopes. It was token bloat from embedding large group lists in a token whose authorization path is already scope-based.

## Why this looks like an upstream bug

Upstream currently appears to do both of the following:

1. Validate self-signed tokens by reading scopes directly from the token's `scope` claim.
2. Only perform group-to-scope mapping for IdP-backed auth methods such as Keycloak / Entra / Cognito / Okta / Auth0.
3. Still generate self-signed user tokens with the full `groups` array included.

That means the `groups` claim is redundant for self-signed user-token authorization, but it still increases token size significantly.

## Impact

- Large self-signed JWTs for users with many groups
- Increased risk of bearer token truncation/corruption in transit
- Failures that surface as malformed token parsing, for example:
  - `Invalid payload padding`
- Hard-to-diagnose production behavior because the token contents are valid in principle, but the token is too large in practice

## Reproduction

1. Log in as a user who is a member of many IdP groups.
2. Generate a self-signed user token via the internal token-vending path.
3. Inspect the token payload and note that it contains the full `groups` array.
4. Use the token for programmatic MCP access.
5. In some environments / transport paths, the request fails and the auth server logs a malformed JWT error such as:
   - `Invalid token - Invalid payload padding`

## Expected behavior

Self-signed user tokens should include only the claims required for downstream authorization.

Since self-signed authorization already uses the embedded `scope` claim directly, the token should not need to embed the full external `groups` list.

Malformed or truncated bearer tokens should also be classified as authentication failures (`401`), not server/configuration failures.

## Actual behavior

- Self-signed user tokens include full `groups`
- Self-signed authorization uses `scope`, not `groups`
- Tokens become larger than necessary
- Invalid/malformed bearer tokens can surface during validation as parsing failures

## Suggested fix

For self-signed user-generated tokens:

1. Stop embedding the full `groups` claim when authorization is scope-based.
[...]

PR #1272, opened 2026-06-20
Fix datetime serialization and token lifetime enforcement (#573, #889)

### LINKED ISSUE #573: GET /api/servers/groups/{group_name} returns 500 due to datetime serialization
## Description

The `GET /api/servers/groups/{group_name}` endpoint returns a 500 Internal Server Error with the message:

```json
{
  "detail": "Internal error: Object of type datetime is not JSON serializable"
}
```

## Root Cause

The DocumentDB `scope_repository.get_group()` method returns the raw MongoDB document which contains Python `datetime` objects for `created_at` and `updated_at` fields. The route handler passes this dict directly to `JSONResponse`, which uses `json.dumps` internally and cannot serialize `datetime` objects.

The `list_groups` endpoint avoids this by only extracting specific scalar fields from the documents.

## Affected Code

- `registry/api/server_routes.py` line 3086: `return JSONResponse(status_code=200, content=group_data)`
- `registry/repositories/documentdb/scope_repository.py` line 294: returns raw MongoDB doc

## Fix

Round-trip the group data through `json.dumps(group_data, default=str)` before passing to `JSONResponse`, converting `datetime` objects to ISO strings. This matches the pattern used elsewhere in the codebase (e.g., `config_routes.py:535`).

Both storage backends are safe:
- **DocumentDB / MongoDB CE**: `datetime` objects converted to strings
- **File-based**: Already returns empty strings for timestamp fields

## Steps to Reproduce

1. Deploy with DocumentDB or MongoDB CE backend
2. Create or import a group scope
3. Call `GET /api/servers/groups/{group_name}` with a valid Bearer token
4. Observe 500 error response

ajmsra:
datetime serialization and token lifetime enforcement (fixes #573, #889)

- **#573**: `GET /api/servers/groups/{group_name}` returned 500 because datetime objects from DocumentDB are not JSON-serializable. Fixed by round-tripping through `json.dumps(default=str)`, matching the pattern used elsewhere in the codebase.
- **#889**: `MAX_TOKEN_LIFETIME_HOURS` (24 h) was declared but never enforced, and the caller-provided `expires_in_hours` was ignored — all self-signed tokens were hardcoded to 8 hours. Fixed by honouring the request field and clamping to `[1, MAX_TOKEN_LIFETIME_HOURS]`.

Note: the original version of this PR also addressed #959 (drop the `groups` claim from self-signed JWTs to fight token bloat). That change was reverted during review because it broke group-restricted resource visibility — the registry's visibility filter reads `user_context["groups"]` at request time, and dropping the claim left every group-restricted resource invisible for non-PingFederate IdPs. The bloat motivation is independently addressed by #1279 (group filtering at login) and #1281 (bulk group→scope lookup), both merged 2026-06-21. #959 will get a separate PR that doesn't regress visibility.

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -2983,7 +2983,20 @@ async def generate_user_token(
             )
 
             current_time = int(time.time())
-            expires_in = DEFAULT_TOKEN_LIFETIME_HOURS * 3600  # 8 hours default
+            # Honour the caller's requested lifetime, clamped to the
+            # server-wide maximum (#889).  Values <= 0 or above the cap
+            # are silently clamped; omitted values fall back to the
+            # default (8 h).
+            effective_hours = min(
+                max(request.expires_in_hours, 1),
+                MAX_TOKEN_LIFETIME_HOURS,
+            )
+            expires_in = effective_hours * 3600
+            if request.expires_in_hours != DEFAULT_TOKEN_LIFETIME_HOURS:
+                logger.info(
+                    f"Token lifetime: requested={request.expires_in_hours}h, "
+                    f"effective={effective_hours}h (max={MAX_TOKEN_LIFETIME_HOURS}h)"
+                )
 
             # Build JWT claims
             jwt_claims = {
--- a/registry/api/server_routes.py
+++ b/registry/api/server_routes.py
@@ -4756,7 +4756,12 @@ async def get_group_api(
                 detail=f"Group '{group_name}' not found",
             )
 
-        return JSONResponse(status_code=200, content=group_data)
+        # Round-trip through json.dumps with default=str to convert
+        # datetime objects from DocumentDB into ISO-format strings (#573).
+        return JSONResponse(
+            status_code=200,
+            content=json.loads(json.dumps(group_data, default=str)),
+        )
 
     except HTTPException:
         raise
--- a/tests/auth_server/unit/test_server.py
+++ b/tests/auth_server/unit/test_server.py
@@ -2914,3 +2914,76 @@ class TestLogScopesLoaded:
         mock_logger.warning.assert_not_called()
         mock_logger.info.assert_called_once()
         assert "1 group mappings" in mock_logger.info.call_args[0][0]
+
+
+# =============================================================================
+# TOKEN LIFETIME ENFORCEMENT (#889)
+# =============================================================================
+
+
+class TestTokenLifetimeEnforcement:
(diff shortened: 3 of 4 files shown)


aarora79:
Hey @ajmsra, took this for a full test ride and ended up rewriting the branch on your fork — heads up that I force-pushed (with lease) and the branch is now at `c402ebb8`. If you have local work on it, you'll want to `git fetch origin && git reset --hard origin/datetime-serialization-token-lifetime-groups-bloat`.

**What's the same:** fix #573 (datetime serialization) and fix #889 (token-lifetime clamping), both verified live on a freshly-rebuilt local stack:

- `GET /api/servers/groups/registry-admins` → HTTP 200 (was 500, exact `TypeError: Object of type datetime is not JSON serializable` reproduced at the line you patched).
- `POST /internal/tokens` lifetime matrix: input 4→4h, 8→8h, 24→24h, 72→24h (clamped), 0→1h, -5→1h. Structured `Token lifetime: requested=…` log line fires for every off-default value as intended.

**What I dropped: the #959 change** (omit `groups` from self-signed JWTs). Reason: I reproduced live that the registry's group-restricted resource visibility check at [`registry/services/visibility.py:90`](registry/services/visibility.py#L90) reads `user_context["groups"]` at request time. With the claim omitted, `validate_self_signed_token` returns `groups=[]`, the registry gets `X-Groups: ""`, and **every group-restricted agent disappears** for non-PingFederate IdPs (the per-provider MongoDB enrichment fallback at [`auth_server/server.py:2306-2353`](auth_server/server.py#L2306-L2353) only defaults `pingfederate`). I also confirmed via the lifetime-test matrix that minted tokens come back with `groups_in_jwt=False`, so the regression would have shipped.

The token-bloat motivation behind #959 should now be covered by:
- **#1279** (merged 2026-06-21) — filters the groups claim down to scope-relevant groups before persisting.
- **#1281** (merged 2026-06-21) — bulk-resolves group → scope mappings so per-request latency no longer scales with group count.

Both landed AFTER you opened this PR. Worth checking against your environment to see if the original "Invalid payload padding" symptom is already gone with current main — if so, #959 may not need a separate PR at all.

**Also rebased onto current main** so the merge will be clean (the branch was behind main by 5 commits including #1279/#1281, which the previous head would have inadvertently reverted).

Thanks for the PR!

ajmsra:
Thanks @aarora79 good catch, thanks again for the quick review on this!

PR #1275, opened 2026-06-20
Refactor/frontend card primitives

omrishiv:
*Issue #, if available:*

*Description of changes:*

Extracts common components for the frontend UI to enable reuse
Updates the Kiro instructions for servers to support DCR


aarora79:
Follow-up: #1319 (merged) builds on this PR's partial primitive extraction by deduplicating ServerCard and CustomEntityCard onto the shared `cards/` primitives with their look kept byte-identical to main (guarded by a new class-set snapshot test). The remaining three cards (Agent, Skill, VirtualServer) were deferred because finishing the full dedup would require homogenizing card layout and actions, which is a much bigger lift.

Issue #553, opened 2026-02-26
nginx config hardcodes bare auth-server hostname, breaks Cloud Map DNS deployments


WPrintz:
## Summary

The nginx config templates (`docker/nginx_rev_proxy_http_only.conf`, `docker/nginx_rev_proxy_http_and_https.conf`) hardcode `auth-server` as a bare hostname in all `proxy_pass` directives:

```nginx
proxy_pass http://auth-server:8888/validate;
proxy_pass http://auth-server:8888/oauth2/callback/keycloak;
# ... ~12 more locations
```

This pattern only resolves in environments where bare container/service names are in the DNS path:
- **docker-compose**: Docker embedded DNS resolves container names
- **Terraform + ECS Service Connect**: Envoy sidecar registers short names in `/etc/hosts`

It **breaks** in deployments using standard DNS-based service discovery (e.g., AWS Cloud Map with `PrivateDnsNamespace`), where only the FQDN resolves (e.g., `auth-server.mcp-gateway.local`). The ECS task crash-loops with:

```
[emerg] host not found in upstream "auth-server" in /etc/nginx/conf.d/nginx_rev_proxy.conf:202
```

## Origin

This has been present since the file was created in commit 073c9a8 (PR #163). The `AUTH_SERVER_URL` env var already exists and carries the correct resolvable hostname for every deployment mode, but the nginx config doesn't use it.

## Suggested Fix

Restore the variable-based pattern that was used in earlier CloudFormation workshop branches:

```nginx
set $auth_server_upstream http://auth-server:8888;   # default
# Override from AUTH_SERVER_URL if set (handled by nginx_service.py or entrypoint)

proxy_pass $auth_server_upstream/validate;
proxy_pass $auth_server_upstream/oauth2/callback/keycloak$is_args$args;
```

Or derive the upstream at config generation time in `nginx_service.py` from the `AUTH_SERVER_URL` environment variable, which is already set in every deployment mode (docker-compose, Terraform, CloudFormation).

## Workaround

In `registry-entrypoint.sh`, resolve the FQDN from `AUTH_SERVER_URL` and add a bare-name `/etc/hosts` entry before nginx starts:

```bash
if [ -n "${AUTH_SERVER_URL:-}" ]; then
    auth_host=$(echo "$AUTH_SERVER_URL" | sed 's|http://||;s|:.*||')
    if echo "$auth_host" | grep -q '\.'; then
[...]

PR #1278, opened 2026-06-20
Update auth Server DNS Resolution in NGINX Templates and Align FileScopeRepository Group Mapping Behavior

ajmsra:
Fixes issues:

#553: Nginx templates hardcoded `auth-server:8888`, breaking Cloud Map DNS deployments where only FQDNs resolve. Replaced all 54 occurrences across both HTTP-only and HTTP+HTTPS templates with `{{AUTH_SERVER_HOST}}:{{AUTH_SERVER_PORT}}` placeholders, and added AUTH_SERVER_URL parsing in nginx_service.py following the existing Keycloak/PingFederate pattern. Defaults to `auth-server:8888` for backward compatibility.

#930: FileScopeRepository.get_all_group_mappings() returned the raw YAML `{keycloak_group: [scope_names]}` while the DocumentDB backend returns the inverted `{scope_name: [keycloak_groups]}`. Since map_cognito_groups_to_scopes() inverts the result expecting the DocumentDB shape, file-backend users got zero scopes and empty ui_permissions. Fixed by inverting in the file backend to match the canonical shape.


aarora79:
Follow-up: PR #1284 (issue #1283) completes the ECS side of this change. It exposes `AUTH_SERVER_URL` as a Terraform variable (`auth_server_url`, default `http://auth-server:8888`) so ECS operators can point the registry/nginx at a Cloud Map / Service Connect FQDN without editing module code. Verified on a live ECS deployment. With #1284 merged, all three deployment surfaces (Docker Compose, Helm, Terraform/ECS) now expose or derive the internal auth-server URL that this PR's nginx templates honor.

PR #1300, opened 2026-06-22
fix(connect-config): omit server auth header under OAuth login for all IDEs (gateway injects egress)

go-faustino:
## Summary

Follow-up to #1224 (per-server OAuth `client_id` login). When a server resolves an OAuth client id (`useOAuthLogin`), the gateway does two things: it runs the IDE OAuth/PKCE login **and** injects the server's stored egress credential upstream server-side. The client should therefore only authenticate to the gateway — it must not send the upstream server's `Authorization`/API-key header.

But the Connect dialog still emitted a client-side server-auth header for any `auth_scheme` server, even in OAuth-login mode:

```jsonc
{
  "mcpServers": {
    "my-server": {
      "url": "https://gw/my-server/mcp",
      "headers": { "Authorization": "Bearer [YOUR_SERVER_AUTH_TOKEN]" },  // ← wrong in OAuth-login mode
      "auth": { "CLIENT_ID": "mcp-gateway" }
    }
  }
}
```

The client forwards that placeholder verbatim, which collides with the gateway-injected egress credential and breaks the connection. Users have to manually delete the header.

## Fix

Drop the server-auth header **whenever an OAuth client id is configured (`useOAuthLogin`)**, across every IDE surface that emits it — not just Cursor:

| IDE | Before (OAuth-login deployment) | After |
|-----|----------------------|-------|
| Cursor (JSON) | `headers.Authorization` + `auth.CLIENT_ID` | `auth.CLIENT_ID` only, no `headers` |
| Roo Code (JSON) | `headers: { X-Authorization, Authorization }` | `headers: { X-Authorization }` (keeps gateway token) |
| Kiro (JSON) | `headers: { X-Authorization, Authorization }` | `headers: { X-Authorization }` |
| Claude Code (CLI) | `--client-id <id>` + `--header "Authorization: ..."` | `--client-id <id>` only |
| Goose (YAML) | `Authorization: Bearer [...]` + `X-Authorization` | `X-Authorization` only |
| VS Code default (JSON) | `headers.Authorization` | dropped |

Codex already omitted the server-auth header. IDEs that can't run the OAuth-login config (Roo Code / Kiro / VS Code default / Goose) **keep** the static `X-Authorization` gateway token — only the upstream server-auth header is removed. All **non-OAuth (static-token)** paths are byte-for-byte unchanged.

Implementation: `buildHeaders()`'s `includeServerAuth` now defaults to `!useOAuthLogin` (one place covers all the JSON IDEs); the Claude Code CLI and Goose builders get the same `!useOAuthLogin` guard.

## Changes

- `frontend/src/components/ServerConfigModal.tsx`
  - `buildHeaders()` `includeServerAuth` defaults to `!useOAuthLogin` (covers Cursor / Roo Code / Claude Code JSON / Kiro / default).
  - `generateClaudeCodeCommand()` and `generateGooseConfig()` skip the server-auth header when `useOAuthLogin` (Goose deps updated).
- `frontend/src/components/__tests__/ServerConfigModal.test.tsx`
  - Regression tests: Cursor (no `headers`), Roo Code & Kiro (server `Authorization` omitted, `X-Authorization` gateway token kept) under OAuth login.

## Notes

Backward compatible: with no resolved OAuth client id, output is unchanged.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Frontend `jest` ServerConfigModal suite (11/11), incl. Cursor / Roo Code / Kiro under OAuth login

### DIFF
--- a/frontend/src/components/ServerConfigModal.tsx
+++ b/frontend/src/components/ServerConfigModal.tsx
@@ -322,11 +322,14 @@ const ServerConfigModal: React.FC<ServerConfigModalProps> = ({
     // Build headers object: custom first, then auth_scheme, then gateway auth.
     // When the IDE handles login via OAuth (useOAuthLogin), the static gateway
     // token is omitted - the IDE obtains it through the OAuth/PKCE flow.
-    // includeServerAuth is also dropped in that mode: the gateway injects the
-    // server's stored egress credential upstream, so the client must not send a
-    // server Authorization/API-key header (the [YOUR_SERVER_AUTH_TOKEN]
-    // placeholder would be forwarded as-is and break the connection).
-    const buildHeaders = (includeGatewayToken = true, includeServerAuth = true) => {
+    // The server-auth header is dropped in that mode too (default
+    // includeServerAuth = !useOAuthLogin), for every IDE that emits it: the
+    // gateway injects the server's stored egress credential upstream, so the
+    // client must not send a server Authorization/API-key header (the
+    // [YOUR_SERVER_AUTH_TOKEN] placeholder would be forwarded as-is and break
+    // the connection). IDEs that can't run OAuth login (Roo Code / Kiro / VS
+    // Code default) still keep the static gateway token, just not server auth.
+    const buildHeaders = (includeGatewayToken = true, includeServerAuth = !useOAuthLogin) => {
       const headers: Record<string, string> = {};
 
       // Custom headers go first so auth_scheme and gateway auth overwrite collisions
@@ -528,7 +531,10 @@ const ServerConfigModal: React.FC<ServerConfigModalProps> = ({
     for (const h of customHeaders) {
       headerLines.push(`      ${h.name}: ${h.value}`);
     }
-    if (server.auth_scheme && server.auth_scheme !== 'none') {
+    // Server auth header. Skipped in OAuth-login mode: the gateway injects the
+    // server's stored egress credential upstream, so the client must not send a
+    // server Authorization/API-key header (the placeholder would break it).
+    if (!useOAuthLogin && server.auth_scheme && server.auth_scheme !== 'none') {
       if (server.auth_scheme === 'bearer') {
         headerLines.push(`      Authorization: Bearer [YOUR_SERVER_AUTH_TOKEN]`);
       } else if (server.auth_scheme === 'api_key') {
@@ -556,7 +562,7 @@ const ServerConfigModal: React.FC<ServerConfigModalProps> = ({
     lines.push('    timeout: 300');
 
     return lines.join('\n');
-  }, [server.name, server.auth_scheme, server.description, server.auth_header_name, isRegistryOnly, jwtToken, customHeaders, buildConnectUrl]);
+  }, [server.name, server.auth_scheme, server.description, server.auth_header_name, isRegistryOnly, useOAuthLogin, jwtToken, customHeaders, buildConnectUrl]);
 
   const generateClaudeCodeCommand = useCallback(() => {
     const serverName = server.name.toLowerCase().replace(/\s+/g, '-').replace(/[^a-z0-9-]/g, '');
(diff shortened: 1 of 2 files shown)


aarora79:
Hey @go-faustino, so we could merge this but this is somewhat a forward looking functionality that we are working on. As in, we currently have no code on main which injects egress credentials for an MCP server (this would be there soon, but does not exist at this time). So if we remove the placeholder then how would an end user know that say an additional header needed to be included. So merging this PR does not as such hurt anything but could you please help me understand how it helps you today. And yes, we would remove this as soon as we merge the egress auth PR.

go-faustino:
Thanks @aarora79 — happy to explain, and your concern is fair.

**Use case.** We're in the #1224 scenario (enterprise Keycloak, anonymous DCR disabled), and the upstream credential for these servers is managed by the gateway side, not the end user: the server is configured with `auth_scheme: bearer` and an egress credential, and the gateway presents it to the upstream when proxying. The user only does the OAuth login to the gateway (the `auth.CLIENT_ID` from #1224). So the hard-coded `Authorization: Bearer [YOUR_SERVER_AUTH_TOKEN]` line is a placeholder they can't fill, and leaving it in collides with the gateway-managed credential and breaks the connection. Stripping it is what makes the copy-paste config work.

**Addressing your point.** You're right that for servers where the user genuinely must supply the header, the placeholder is a useful hint and shouldn't be dropped. So rather than change the default, I'll gate this behind a config flag — the same pattern we used for `oauth_client_id` / `append_mcp_path` in #1224 / #1241:

- a per-server `suppress_server_auth_header: bool | null` on the server entry, plus a registry-wide default, resolved **server-side** in `GET /api/servers/{path}/connect-config` (`server_info.<field> ?? settings.<default>`);
- default `false`/null → **behavior unchanged** (placeholder shown for every IDE, exactly as today);
- set `true` only for servers whose credential the gateway manages → the Connect config omits the client server-auth header (still keeping the `X-Authorization` gateway token for the non-OAuth IDEs, as in this PR).

That keeps `main` byte-for-byte identical unless an operator opts in, and when your first-class egress-injection feature lands this flag is the natural signal to flip on (or drop).

Want me to update the PR to this opt-in form? Happy to match your naming preference too — e.g. `suppress_server_auth_header`, `egress_auth_managed`, or `gateway_injects_server_auth`.

aarora79:
@go-faustino are you using a downstream/custom egress-auth implementation today? Because on main this injection path does not exist. I am leaning towards just merging your PR as is without any new config param simply because we will very soon have egress auth functionality so this credentials hint will not be needed anyway and adding a new param across all deployment surfaces for a functionality that would not be there seems too much work for little gain. The only clarification I need is about your egress-auth implementation.

go-faustino:
Good question — to be precise, **no, we're not running any custom egress-credential injection** today. We're in the #1224 OAuth-login setup, and the upstream MCP servers are themselves OAuth-protected (enterprise Keycloak in front of them).

The flow today:
1. The IDE does the #1224 OAuth/PKCE login to the gateway (`auth.CLIENT_ID`).
2. The gateway forwards that **user token** (plus `X-User`/`X-Username`) to the upstream MCP server.
3. The upstream validates the user's token itself and authenticates the end user — verified end-to-end: the upstream records the real caller (`user_id=<logged-in user's email>`), not a service account.

So there's no separate "server auth token" the user holds. The hard-coded `Authorization: Bearer [YOUR_SERVER_AUTH_TOKEN]` in the Connect config is therefore (a) unfillable, and (b) it overwrites/collides with the OAuth-obtained token the client uses, breaking the connection. Dropping it under `useOAuthLogin` is exactly what makes the generated config work; non-OAuth IDEs keep their static `X-Authorization` gateway token unchanged.

That's the entire benefit for us today — it's independent of your upcoming egress-injection feature; it just stops the Connect dialog from emitting a placeholder header that conflicts with the #1224 OAuth-login flow. Happy to have it merged as-is, and it stays compatible whenever first-class egress-auth lands (still no client-side server-auth header in OAuth-login mode).

aarora79:
thanks for explanation, will merge this.

PR #1321, opened 2026-06-25
feat(mcp-proxy): make upstream timeout configurable (MCP_PROXY_UPSTREAM_TIMEOUT_SECONDS)

### LINKED ISSUE #1320: MCP proxy hop upstream timeout is hardcoded to 30s (long-running tool calls fail with 504)
## Problem

The auth-server `/mcp-proxy` hop opens its upstream connection with a **hardcoded** 30s timeout:

```python
# auth_server/server.py (mcp_proxy)
async with httpx.AsyncClient(timeout=30.0) as client:
    async with client.stream("POST", upstream_url, ...) as upstream_response:
```

Any upstream MCP tool call that legitimately runs longer than 30s is killed and the hop returns `504 Upstream MCP server timed out`, even when the upstream is healthy and still working.

## Real-world impact

We front kagent multi-cluster A2A coordinator agents through the gateway. A coordinator `invoke_agent` call fans out to many spoke clusters before responding, which routinely exceeds 30s. Simple calls (`list_agents`, `initialize`, `tools/list`) succeed; the heavier delegations time out at exactly 30s.

There is no way to raise this without rebuilding the image, because the value is not configurable.

## Proposed fix

Make the proxy-hop upstream timeout configurable, following the existing `mcp_proxy_max_body_bytes` pattern (settings field + `MCP_PROXY_*` env var fallback + admin-config exposure). Default stays `30.0` so behavior is unchanged unless an operator opts in.

Proposed knob: `MCP_PROXY_UPSTREAM_TIMEOUT_SECONDS` / `settings.mcp_proxy_upstream_timeout_seconds` (float, minimum 1, default 30).

I have a focused PR ready that mirrors the `mcp_proxy_max_body_bytes` implementation (config field, `_read_*` helper with env fallback, admin-config entry, docs, unit tests).

go-faustino:
## Description

The auth-server `/mcp-proxy` hop opens its upstream connection with a **hardcoded** 30s timeout:

```python
# auth_server/server.py (mcp_proxy)
async with httpx.AsyncClient(timeout=30.0) as client:
```

Any upstream MCP tool call that legitimately runs longer than 30s is killed and the hop returns `504 Upstream MCP server timed out`, even when the upstream is healthy and still working. There is no way to raise it without rebuilding the image.

Fixes #1320.

## Root cause

`mcp_proxy()` in `auth_server/server.py` constructs `httpx.AsyncClient(timeout=30.0)` with a literal. Unlike the sibling `mcp_proxy_max_body_bytes` knob, there is no settings field or env var to override it, so operators fronting long-running upstreams (e.g. multi-cluster A2A coordinator agents whose `invoke_agent` fans out to many spoke clusters before responding) get hard 504s at 30s while lighter calls — `initialize`, `tools/list`, `list_agents` — succeed.

## Fix

Make the proxy-hop upstream timeout configurable, mirroring the existing `mcp_proxy_max_body_bytes` implementation exactly:

- `registry/core/config.py` — new `mcp_proxy_upstream_timeout_seconds: float` `Field` (default `30.0`, `gt=0`).
- `auth_server/server.py` — new `_read_mcp_proxy_upstream_timeout()` helper (reads `settings` first, falls back to `MCP_PROXY_UPSTREAM_TIMEOUT_SECONDS`, clamps to a 1s floor, invalid values fall back to the default); `mcp_proxy()` now uses the resolved value instead of the literal.
- `registry/api/config_routes.py` — expose the field in the admin config view (non-sensitive), next to `mcp_proxy_max_body_bytes`.
- `docs/unified-parameter-reference.md` — documented across Docker / Terraform / Helm.

Default stays `30.0`, so behavior is **byte-for-byte unchanged** unless an operator opts in.

## Testing

- `python -m py_compile` on all four changed Python files — passes.
- New unit tests `tests/auth_server/unit/test_server.py::TestMcpProxyUpstreamTimeout` (5 cases): settings value wins, env fallback when settings unset, default when unset, invalid env falls back to default, sub-second value clamped to the 1s floor. Validated the resolution logic standalone (all 5 pass).
- `ruff format --check` clean on the added lines; remaining `ruff check` findings in the touched files are pre-existing `UP042` enum warnings unrelated to this change.

Note: I couldn't run the full suite locally — `uv sync` pulls `torch` (a transitive dep) which fails to download in my sandboxed network. Happy to iterate on any CI findings.

## Checklist

- [x] Working against the latest `main`
- [x] Change is focused on the specific fix (no unrelated reformatting)
- [x] Opened an issue first (#1320)
- [x] Clear commit message
- [x] Added unit tests
- [ ] ~~Full integration test suite run locally~~ (requires `uv sync`, which fails on the `torch` dep in my sandbox; validated via py_compile + standalone logic tests)


go-faustino:
Pushed a follow-up addressing two review points:

### 1. Codecov patch coverage (was 90.9%, 2 missing lines)

The uncovered lines were the `except (TypeError, ValueError)` branch in `_read_mcp_proxy_upstream_timeout()` (a settings value that fails `float()`). Added `test_invalid_settings_value_falls_back_to_default` to exercise it, so the helper is now fully covered.

### 2. Other timeout layers in the proxy chain

I traced the full MCP proxy chain to make sure no other layer silently caps long calls:

| Hop | Layer | Before |
|-----|-------|--------|
| client -> registry nginx (MCP location block) | nginx `proxy_read_timeout` | **unset -> 60s implicit default** |
| registry nginx -> `auth-server:8888/mcp-proxy/` | same block | 60s |
| auth-server uvicorn | request timeout | none (uvicorn has no request timeout) |
| auth-server `mcp_proxy` -> upstream | `httpx.AsyncClient(timeout=...)` | 30s (the original fix) |

The generated MCP location blocks (`nginx_service._create_location_block`) set **no** `proxy_read_timeout`, so they fell back to nginx's 60s default. That made 60s the *next* ceiling: raising `MCP_PROXY_UPSTREAM_TIMEOUT_SECONDS` above 60 would have had no effect because nginx would sever the connection first.

Fix: the MCP location blocks now derive `proxy_read_timeout`/`proxy_send_timeout` from the **same** `MCP_PROXY_UPSTREAM_TIMEOUT_SECONDS` knob plus a fixed 30s headroom buffer (default 30s -> 60s, i.e. byte-for-byte unchanged unless raised). One knob now raises the entire chain, and the inner auth-server hop always times out first so clients get the clean `504 Upstream MCP server timed out` rather than nginx's generic cutoff. `proxy_connect_timeout` stays at 10s (the hop is in-cluster).

Note: set the env on **both** the registry and auth-server deployments (docs updated) — auth-server enforces the inner httpx timeout, registry derives the nginx read timeout. Anything in front of the gateway (e.g. an ALB/ingress idle timeout) is deployment-specific and out of scope for this repo.

Added unit tests for the new `_resolve_mcp_proxy_read_timeout_seconds()` helper (default, scaled, env-fallback, invalid) and extended `test_create_location_block_streamable_http` to assert the timeout directives are emitted.

aarora79:
Thank you for this, @go-faustino, and for the careful writeup and tests.

While this was in flight, the core of this feature landed on `main` via #1356 and #1358, which make the auth-server mcp-proxy hop upstream timeout configurable. The one difference is the knob name: `main` ships it as **`MCP_PROXY_TIMEOUT`** (`settings.mcp_proxy_timeout`), wired across Docker / Terraform / Helm and exposed in the admin config view. Because that already exists, merging this as-is would introduce a second, competing env var for the same behavior, so we're going to close this PR as superseded.

That said, your PR does one thing `main` does not, and it's the better design: deriving nginx's `proxy_read_timeout` for the `/mcp-proxy/` location blocks from the upstream timeout (`ceil(upstream) + 30s` buffer) so a single knob raises the whole proxy chain and the inner hop always times out first. On `main` today that nginx adjustment is still manual (documented, but on the operator). We're going to take that piece from your work and add it to `main` on top of the existing `mcp_proxy_timeout` field, with credit to you.

Closing as superseded by #1356 / #1358. Genuinely appreciate the contribution, and thank you for surfacing the nginx side of this.

aarora79:
Following up: the nginx piece from this PR has now landed on `main` via #1361 (merged in `57664932`). The generated `/mcp-proxy/` location blocks now derive their `proxy_read_timeout`/`proxy_send_timeout` from `MCP_PROXY_TIMEOUT` (plus a 30s buffer), so raising the timeout lifts the whole proxy chain without any manual nginx change, exactly as you designed it here. Credited to you in the code and PR. Thanks again for surfacing this, @go-faustino.

go-faustino:
Thanks so much, @aarora79 — this is a genuinely great outcome, and I really appreciate the care you took closing it out.

Consolidating on a single `MCP_PROXY_TIMEOUT` knob rather than landing a second, competing env var is absolutely the right call — glad the core already landed on `main`. And thank you for taking the nginx piece from here and shipping it in #1361: one knob raising the whole proxy chain so the inner hop always times out first (and clients get the clean 504) was the part I most wanted to see upstream, so I'm really happy it made it in.

No concerns at all with closing as superseded — the end state on `main` is cleaner than this PR. Thanks again for the thoughtful review and the credit; it was a pleasure contributing.

Issue #1323, opened 2026-06-25
registry_management.py: toggle and server-rate fail with HTTP 422 (CLI/API payload mismatch)


aarora79:
## Summary
Two `registry_management.py` subcommands are broken end-to-end: the CLI sends request payloads that the registry API rejects with **HTTP 422**, so `toggle` and `server-rate` never succeed.

### 1. `toggle` -> 422
`RegistryClient.toggle_service` posts `{"service_path": ...}` to `POST /api/servers/toggle`, but the endpoint requires form fields **`path`** plus an explicit boolean **`new_state`** (it does not flip the current state itself). The response also returns `service_path`/`new_enabled_state`, which the client did not map onto `ToggleResponse`.

### 2. `server-rate` -> 422
`RegistryClient.rate_server` posts to `POST /api/servers/{path}/rate`, which expects a **JSON** `RatingRequest` body, but the client's `_make_request` routed `/rate` through the **form-encoded** branch.

## Reproduction
```bash
uv run python api/registry_management.py --registry-url http://localhost \
    --token-file .token toggle --path /currenttime/
# -> 422 Unprocessable Content (field 'new_state' required)

uv run python api/registry_management.py --registry-url http://localhost \
    --token-file .token server-rate --path /currenttime/ --rating 5
# -> 422 Unprocessable Content (body should be a JSON object)
```

## Fix
- `toggle_service`: read current state via `get_server`, send `path` + opposite `new_state` as form fields, map the response keys onto `ToggleResponse`.
- `rate_server`: route `/rate` endpoints through the JSON branch in `_make_request`.
- Add unit tests pinning both request shapes.

## Status
Fixed on the issue-1316 branch / PR (commit `eaae2289`, with `tests/unit/api/test_registry_client_toggle_rate.py`). This issue tracks the standalone bug for visibility.

Issue #1322, opened 2026-06-25
Grafana analytics dashboard panels show 'No data' (success label case mismatch)


aarora79:
## Summary
Several panels in the **MCP Gateway - Analytics Dashboard** render **"No data"** even when metrics are flowing, because the dashboard queries filter on the wrong label-value case.

## Root cause
The auth and tool middleware label metrics with a Python boolean rendered via `str(bool)`, so the emitted label is `success="True"` / `success="False"` (capital first letter). The dashboard panels query `success="true"` / `success="false"` (lowercase), which matches **zero** series in Prometheus.

Affected panels:
- Authentication Success Rate (shows red "No data")
- Authentication Flow Analysis
- Real-time Protocol Activity (auth portion)
- Error-rate panels

## Evidence
```
# emitted (capital):
mcpgw_registry_auth_request_total{success="True",method="session_cookie",...} 174
# dashboard query (lowercase) -> 0 series
mcpgw_registry_auth_request_total{success="true"}
```

## Fix
Update the dashboard JSON queries to use `success="True"`/`"False"` to match what is emitted. Both copies need the change:
- `config/grafana/dashboards/mcp-analytics-comprehensive.json`
- `terraform/aws-ecs/grafana/dashboards/mcp-analytics-comprehensive.json`

(Fixing the dashboard rather than the metric avoids breaking any existing query/recording rule already using `"True"`.)

## Status
Fixed on the issue-1316 branch / PR (commit `90c0297d`). This issue tracks the standalone bug for visibility.

PR #1324, opened 2026-06-25
feat(auth): /validate hot-path optimization + free-threaded Python 3.14t trial (#1316)

### LINKED ISSUE #1316: auth_server /validate: trim hot-path logging + trial free-threaded Python 3.14t
## Summary

The `auth_server` `/validate` endpoint is on the per-request hot path: nginx fires it as an `auth_request` subrequest on **every** proxied request (every MCP call and every `/api/` call) — see [docker/nginx_rev_proxy_http_and_https.conf#L147](https://github.com/agentic-community/mcp-gateway-registry/blob/506aa7fee9b5dbc0e680704e31e92092e746a1ae/docker/nginx_rev_proxy_http_and_https.conf#L147). Unlike the I/O-bound tool servers, `/validate` does real CPU work per call (JWT signature verification, scope-subset math, tool-access resolution), so it is a latency and throughput chokepoint under concurrency.

This issue tracks two independent improvements, cheapest first. All line references are pinned to commit `506aa7fe`.

---

## Track 1 — Trim per-request logging on the hot path (cheap win, do regardless)

`validate_request` emits ~40 log statements per call, and at the default `INFO` level **three of them pretty-print JSON with `indent=2` on every single request**:

| File:line (@506aa7fe) | Statement | Level | Cost |
|---|---|---|---|
| `auth_server/server.py:1947` | `json.dumps(request_payload, indent=2)` | INFO | every request with a body |
| `auth_server/server.py:2653` | `json.dumps(_mask_sensitive_dict(validation_result), indent=2)` | INFO | every request |
| `auth_server/server.py:2655` | `json.dumps(response_data, indent=2)` | INFO | every request |
| `auth_server/server.py:1965` | `json.dumps(masked_headers, indent=2)` | DEBUG | only if DEBUG |

The full enumeration of ~40 hot-path log statements (with line numbers) is captured in the LLD. Proposal: demote the high-cost/verbose ones to `DEBUG`, drop `indent=2`, and gate payload/response dumps behind a debug check so they cost nothing at `INFO`. This is free latency with no rewrite and should be measured first — it may recover most of the per-request overhead on its own.

## Track 2 — Trial free-threaded Python 3.14 (`3.14t`) on the auth_server container

`/validate` is CPU-bound and GIL-bound: under concurrency, the per-request Python work serializes through one core. Python 3.14 ships a free-threaded ("no-GIL") build that lets this work run truly in parallel across cores — potentially the same win we'd otherwise reach for Go to get, without leaving Python.

Scope is **one container**: `auth_server` has its own `pyproject.toml` + `uv.lock` and builds its own image ([docker/Dockerfile.auth#L26](https://github.com/agentic-community/mcp-gateway-registry/blob/506aa7fee9b5dbc0e680704e31e92092e746a1ae/docker/Dockerfile.auth#L26)), independent of the torch-heavy registry. The experiment cannot leak into other services.

### Step zero (prerequisite): swap the OTLP exporter from gRPC to HTTP

In free-threaded CPython, importing a C extension that has **not** declared free-threading support causes the interpreter to **silently re-enable the GIL** for the whole process — defeating the experiment. Every compiled dependency in `auth_server` already ships free-threaded wheels **except `grpcio`**, which is pulled in only as the OTLP **gRPC** telemetry transport (not on the request path).

Fix: switch OTLP export to the pure-Python `opentelemetry-exporter-otlp-proto-http` variant and drop `grpcio` from `auth_server/pyproject.toml`. This is the first task of Track 2 and is detailed in the LLD; its verification (GIL actually disabled, no re-enable warning) is in the testing plan.

### Dependency free-threading support (verified against PyPI latest)

| Dependency | Status |
|---|---|
| pydantic-core, cryptography, pyyaml, aiohttp, pymongo, uvloop, httptools, frozenlist, multidict, yarl, charset-normalizer | free-threaded wheels available |
| pyjwt, python-jose, boto3, httpx, requests, pydantic, pydantic-settings, itsdangerous, motor, aiofiles, opentelemetry-* | pure Python (compatible) |
| **grpcio / grpcio-tools** | **no free-threaded wheel — must be removed (see step zero)** |

### Approach

1. Step zero: OTLP gRPC → HTTP, regenerate `auth_server/uv.lock`.
2. Switch `docker/Dockerfile.auth` base image to the free-threaded 3.14 build.
3. Verify the GIL is actually disabled at runtime.
4. Benchmark `/validate` before/after under concurrency.

## Success metrics

- `/validate` p50/p95/p99 latency under concurrent load (before vs. after each track).
- auth_server process RSS.
- Confirmation that free-threaded mode is genuinely active (`sys._is_gil_enabled()` is `False`, no GIL re-enable warning at startup).

Detailed design: `.scratchpad/issue-1316/lld.md`. Verification plan: `.scratchpad/issue-1316/testing.md`.

aarora79:
## Overview
Implements the issue #1316 LLD: optimize the auth-server `/validate` hot path and trial free-threaded ("no-GIL") Python 3.14t on the auth-server container. `/validate` is the nginx `auth_request` subrequest that runs on **every** proxied request, so its latency and throughput are a gateway-wide chokepoint.

Closes #1316.

## Track 1 - hot-path logging cleanup
- Demote the per-request `INFO` log statements on the `/validate` happy path to `DEBUG`, and gate the structured `json.dumps(...)` dumps behind `logger.isEnabledFor(DEBUG)` with lazy `%s` args. Removes three `json.dumps(indent=2)` serializations plus ~10 string-formats per request at the default level. JWT/scope/tool-access semantics are unchanged.

## Track 2 - free-threaded Python 3.14t
- **Remove grpcio** (no free-threaded `cp314t` wheel; would silently re-enable the GIL at import). It entered the auth tree via two paths, both removed: the gRPC OTLP exporter (swapped for `opentelemetry-exporter-otlp-proto-http`) and the unused `opensearch-py` (which now pulls `opensearch-protobufs` -> grpcio). `uv.lock` regenerated; `grpcio`/`grpcio-tools` no longer present.
- **OTLP push moved to HTTP/protobuf (port 4318)** for auth-server only, across docker-compose, the auth-server + stack Helm charts (+ reserved-env-names, unittests), ECS terraform, and `.env.example`. Registry and mcpgw stay on grpc/4317. The optional OTLP push path is off by default; the always-on Prometheus `:9464` exporter is untouched.
- **`docker/Dockerfile.auth`** keeps the `python:3.14.6-slim` base (for security patching) and has `uv` install the managed free-threaded interpreter (`3.14t`) into a fixed path shared by both build stages. The entrypoint reports free-threading status at startup and can fail (`AUTH_REQUIRE_FREE_THREADING=true`) if the GIL is unexpectedly re-enabled.

Verified in a built image: the GIL stays disabled after importing cryptography, pydantic_core, pymongo, aiohttp, and the HTTP OTLP exporter; grpc is absent.

## Benchmark harness (decision gate)
- `tests/stress/measure_validate_latency.py`: concurrent load test for `/validate` (success + error paths), reading server-side latency from Prometheus via a **before/after histogram snapshot-delta** (isolates each run, immune to rate-window bleed). Supports `--sweep` (concurrency levels x `--repeats`, median-aggregated) and `--compare-to`.
- `tests/stress/run_validate_gil_ab.sh`: A/B orchestrator that runs the sweep twice on the **same** free-threaded image, flipping the GIL via `PYTHON_GIL=1` (no rebuild needed), and prints the comparison.

**Preliminary result:** at low concurrency the GIL is neutral; the free-threading win appears on the **success-path p95/p99 under concurrency** (~40% at c=100 in an early single-run pass). The error path is I/O-bound (blocking Keycloak calls) and unaffected. A full 7-level x 3-repeat sweep (c=50..200, GIL on vs off) is running; the aggregated table will be added to this PR before merge to settle the promotion decision.

## Drive-by fixes (tracked as separate issues)
While verifying telemetry and APIs end-to-end, two pre-existing, unrelated bugs surfaced and are fixed here:
- **#1322** - Grafana analytics dashboard panels showed "No data" due to a `success` label case mismatch (`"true"` vs emitted `"True"`). Both the compose and terraform dashboard copies fixed.
- **#1323** - `registry_management.py` `toggle` and `server-rate` failed with HTTP 422 (CLI/API payload mismatch). Fixed in `api/registry_client.py` with unit tests.

## Testing
- `auth_server` unit tests: pass (348). Helm unittests (auth-server + stack): pass (53). New CLI tests: pass (4). Full suite shows no new regressions.
- Live stack: auth-server runs free-threaded (GIL off, no errors), `/validate` healthy, metrics flow on both the `:9464` scrape and OTLP-push paths, dashboard panels populate, CLI read+write verified.

## Rollout note
The free-threaded image is a trial behind this branch. The sweep result is the decision gate for promoting it as the default; until then the change is isolated to the auth-server image and is functionally identical (same wheels/code, GIL toggle only).


aarora79:
## Closing: free-threading experiment concluded — not promoting

This PR trialed free-threaded ("no-GIL") Python 3.14t on the auth-server to speed up the CPU-bound `/validate` hot path under concurrency (issue #1316). **We are closing it without merging: the experiment showed no meaningful benefit, and the analysis explains why the approach was mismatched to the actual bottleneck.**

### What the A/B benchmark showed
Using a same-image GIL toggle (`PYTHON_GIL=1` vs free-threaded), a concurrency sweep (50–200, 3 repeats each) measuring server-side `/validate` p95/p99 from Prometheus: the GIL-on vs GIL-off deltas flipped sign across concurrency levels (e.g. c=50 regressed, c=100 improved, c=125 regressed) with no monotonic trend — i.e. **within run-to-run noise, no consistent win.**

### Why (the mechanistic reason, not just noisy numbers)
- The auth-server runs a **single uvicorn process / single asyncio event loop**, and `validate_jwt_token` (the CPU work) is called **inline** in the async handler. So concurrent requests are serialized by the one event-loop thread — there is **no thread-level CPU parallelism for no-GIL to unlock**. The GIL was never the bottleneck.
- The real saturation risk at high concurrency is **blocking I/O** (synchronous Keycloak/JWKS calls in `requests`) stalling the event loop — which is **GIL-independent** (blocking I/O releases the GIL anyway).
- Even the CPU part (RSA/HMAC signature verification via `cryptography`/`hashlib`) is **native code that already releases the GIL**, so free-threading adds parallelism only for cheap pure-Python leftovers.

### Correct scaling path (if/when `/validate` load is a concern)
For a **stateless** validation endpoint, the high-leverage, GIL-independent levers are, in order: (1) horizontal scaling — more uvicorn `--workers` and/or ECS replicas; (2) async-ify the Keycloak/JWKS I/O; (3) threadpool-offload the CPU verify; free-threading is a distant fourth. None of this is needed at current production load.

### What we keep (moved to a clean PR off main)
The investigation surfaced unrelated, pre-existing issues worth fixing on their own:
- `/validate` hot-path logging cleanup (demote per-request INFO/json.dumps to gated DEBUG) — pure perf hygiene, independent of the interpreter.
- #1322 — Grafana dashboard `success` label case mismatch ("No data" panels).
- #1323 — `registry_management.py` toggle/rate HTTP 422 (CLI/API payload mismatch).

These will land via a new branch from `main` (clean history, no free-threading machinery). The free-threading-specific changes here (grpcio removal, OTLP gRPC→HTTP swap, free-threaded Dockerfile, the auth exporter-selector patch that only existed because grpcio was dropped) are **not** carried over — they were artifacts of the experiment.

Separately tracked, also as their own PRs (pre-existing, not free-threading-related):
- #1326 — ECS registry/mcpgw metrics land in AMP as `job=unknown_service` (missing `service.name`; traces UNIMPLEMENTED).
- #1325 — ECS Grafana admin password stored plaintext in task-def env; move to Secrets Manager.

The benchmark harness (`tests/stress/measure_validate_latency.py`) and this experiment record are preserved in this branch's history for future reference.

PR #1328, opened 2026-06-25
fix(ecs): registry/mcpgw metric labels (#1326) + Grafana password to Secrets Manager (#1325)

### LINKED ISSUE #1325: Store ECS Grafana admin password in Secrets Manager (currently plaintext in task-def env)
## Summary
The Grafana admin password on the ECS/Terraform deployment is injected as a **plaintext environment value** in the ECS task definition, instead of being stored in AWS Secrets Manager and referenced via `secrets` / `valueFrom` like every other secret in the stack. This is the odd one out and should be brought into line.

## Current behavior
`var.grafana_admin_password` is placed directly into the container `environment` block:

```hcl
# terraform/aws-ecs/modules/mcp-gateway/observability.tf
{ name = "GF_SECURITY_ADMIN_PASSWORD", value = var.grafana_admin_password }   # line ~645
# and the grafana service container ~line 583
```

Consequences:
- The password is visible in plaintext via `aws ecs describe-task-definition` to anyone with that IAM permission.
- It is stored in plaintext at rest in the Terraform state file.
- It diverges from the stack's established pattern for sensitive values.

## Established pattern to follow
Other secrets in this module use a Secrets Manager secret + a `secrets`/`valueFrom` reference, e.g.:

```hcl
# terraform/aws-ecs/modules/mcp-gateway/secrets.tf
resource "aws_secretsmanager_secret"        "secret_key" { ... }
resource "aws_secretsmanager_secret_version" "secret_key" { secret_string = ... }

# referenced in the task def (ecs-services.tf) as:
secrets = [
  { name = "SECRET_KEY", valueFrom = aws_secretsmanager_secret.secret_key.arn },
  { name = "KEYCLOAK_CLIENT_SECRET", valueFrom = "${aws_secretsmanager_secret.keycloak_client_secret.arn}:client_secret::" },
  ...
]
```

## Proposed fix
1. Add `aws_secretsmanager_secret` + `aws_secretsmanager_secret_version` for the Grafana admin password (generate via `random_password` when not provided, mirroring `secret_key`).
2. Move `GF_SECURITY_ADMIN_PASSWORD` from the container `environment` block to the `secrets` block using `valueFrom = <arn>`.
3. Grant the Grafana task execution role `secretsmanager:GetSecretValue` on the new ARN (mirror the existing IAM grants for the other secrets).
4. Check the grafana-config sidecar that builds `GURL="http://admin:$${GF_SECURITY_ADMIN_PASSWORD}@..."` still receives the value (it reads the same env var, so it works once the secret is wired as an env var via `secrets`).
5. Keep the CDK (`infra/`) path in sync if it has an equivalent.

## Scope / notes
- ECS/Terraform deployment only. Docker Compose reads `GRAFANA_ADMIN_PASSWORD` from `.env` (separate concern).
- Rotate the current Grafana admin password as part of this change, since it has been exposed in task-def env and local state.
- To be fixed in a separate PR.

### LINKED ISSUE #1326: ECS: registry/mcpgw metrics land in AMP as job=unknown_service (OTLP-push lacks service.name; traces UNIMPLEMENTED)
## Summary
On ECS, the `mcp-registry` and `mcpgw` services emit OpenTelemetry metrics that **do** reach Amazon Managed Prometheus, but under the wrong identity: they land with **`job="unknown_service"`** instead of `job="mcp-registry"` / `job="mcpgw"`. The ADOT collectors for these services also log continuous `Failed to export traces ... UNIMPLEMENTED` errors. This is a pre-existing inconsistency, separate from the auth-server regression fixed in #1316 (PR #1324).

## Evidence (live AMP, 2026-06-25)
After driving traffic to the ECS gateway:
```
mcpgw_registry_operation_total -> 8 series, all job="unknown_service", otel_scope_name="mcp-gateway-registry"
up{}                           -> only job="mcp-metrics-service"   (never mcp-registry / mcp-auth-server)
```
Registry ADOT app logs:
```
ERROR opentelemetry.exporter.otlp.proto.grpc.exporter: Failed to export traces to localhost:4317, error code: StatusCode.UNIMPLEMENTED   (x200+)
```

## Root cause
Two related issues in the OTLP-**push** path used by registry/mcpgw (`terraform/aws-ecs/modules/mcp-gateway/observability.tf`, `local.adot_otlp_to_amp_config`):

1. **No `service.name` resource attribute.** The push path does not set `OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES`, so AMP's remote-write labels every series `job="unknown_service"`. (Compose sets `OTEL_SERVICE_NAME`; the ECS registry/mcpgw task defs do not.)
2. **Traces sent to a metrics-only pipeline.** The collector config defines only a `metrics` pipeline, so the app's exported traces are rejected `UNIMPLEMENTED`. Harmless but noisy (200+ errors/run).

Contrast: `mcp-metrics-service` uses a **Prometheus scrape** receiver (`local.adot_config`, scrapes `localhost:9465`) and works correctly with the right `job` label. The same scrape pattern is what local Docker Compose and the Helm charts use for registry/auth (`OTEL_EXPORTER_PROMETHEUS_HOST` -> PrometheusMetricReader on :9464). ECS registry/mcpgw are the only surface on the OTLP-push path.

## Proposed fix (option B from the #1316 investigation)
Converge registry and mcpgw onto the **Prometheus-scrape** pattern that already works for metrics-service and on every other deployment surface:
- Give registry/mcpgw an ADOT config with a `prometheus` receiver scraping `localhost:9464` (distinct `job_name` per service).
- Set `OTEL_EXPORTER_PROMETHEUS_HOST=0.0.0.0` / `PORT=9464` on the containers and remove `OTEL_EXPORTER_OTLP_ENDPOINT` so the entrypoint stops wrapping with `opentelemetry-instrument` and the app installs its PrometheusMetricReader.

Verified safe in #1316 investigation: the `:9464` Prometheus reader exposes every metric the Grafana dashboard queries (`mcpgw_registry_operation_total`, `tool_discovery_total`, `http_server_*`, etc.) — switching loses nothing, and the collector has no traces pipeline to lose.

### Alternative (smaller, keeps push path)
Set `OTEL_SERVICE_NAME` (fixes the label) + `OTEL_TRACES_EXPORTER=none` (stops the UNIMPLEMENTED spam) on registry/mcpgw, leaving the OTLP-push architecture in place. Lower blast radius but keeps ECS as the odd-one-out vs other surfaces.

## Scope
- ECS/Terraform deployment only (registry + mcpgw). auth-server is handled separately in #1316.
- Roll out registry first as a canary; confirm `up{job="mcp-registry"}` appears in AMP and the Grafana dashboard populates, then mcpgw.
- Verify against the live AMP workspace `ws-f38f8821-9e58-421b-a5bf-581865c18acc`.

aarora79:
## Overview
Two pre-existing ECS observability fixes surfaced while investigating issue #1316 (the free-threading experiment, since closed). Both are infrastructure-only and verified by `terraform validate` + CDK `tsc` build.

Fixes #1326
Fixes #1325

## #1326 — registry/mcpgw metrics land in AMP as `job=unknown_service`
On the live ECS (Terraform) deployment, registry and mcpgw push metrics to the ADOT sidecar over OTLP, but never set `service.name`. AMP's `prometheusremotewrite` therefore labels every series `job="unknown_service"`, so dashboard/`job` filters miss them, and the apps also spam `Failed to export traces ... UNIMPLEMENTED` (the ADOT pipeline is metrics-only).

Fix (registry + mcpgw task defs):
- `OTEL_SERVICE_NAME` = `mcp-gateway-registry` / `mcp-mcpgw` — matches the names already used on Docker Compose and Helm, so all surfaces agree and AMP gets the correct `job` label.
- `OTEL_TRACES_EXPORTER=none` — stops the metrics-only collector rejecting exported traces with UNIMPLEMENTED.

**CDK note:** not applicable to the CDK stack. Its ADOT sidecar uses a Prometheus *scrape* receiver on metrics-service only and never sets `OTEL_EXPORTER_OTLP_ENDPOINT` on the app containers, so there is no OTLP push to mislabel.

## #1325 — Grafana admin password plaintext in task-def env
The Grafana admin password was injected as a plaintext container `environment` value (visible via `aws ecs describe-task-definition` and in Terraform/CDK state), unlike every other secret in the stack.

Fix (both Terraform and CDK, for parity):
- Store the password in Secrets Manager (KMS-encrypted), reference via `valueFrom` (Terraform) / `ecs.Secret.fromSecretsManager` (CDK).
- Applied to **both** the Grafana container and the post-install config sidecar that builds the Grafana API URL.
- Grant the Grafana execution role `GetSecretValue` + `kms:Decrypt` on the new secret.

**Operational note:** rotate the currently-exposed Grafana password as part of applying this change (it was previously readable in plaintext).

## Testing
- `terraform validate`: Success (pre-existing `data.aws_region.name` deprecation warnings only).
- `terraform fmt`: clean.
- CDK `npm run build` (tsc): passes.
- Both changes are infra-only; verification of the live effect (correct `job` label in AMP, password no longer in task-def env) happens on the next ECS deploy.

### DIFF
--- a/infra/lib/registry/constructs/observability-pipeline.ts
+++ b/infra/lib/registry/constructs/observability-pipeline.ts
@@ -50,6 +50,8 @@ export interface ObservabilityPipelineProps {
   readonly metricsApiKeySecret?: secretsmanager.ISecret;
   /** Secrets Manager secret holding the OTLP exporter headers */
   readonly otlpExporterHeadersSecret?: secretsmanager.ISecret;
+  /** Secrets Manager secret holding the Grafana admin password (issue #1325) */
+  readonly grafanaAdminPasswordSecret?: secretsmanager.ISecret;
   /** Additional IAM policy statements for Secrets Manager + KMS access */
   readonly secretsAccessStatements: iam.PolicyStatement[];
   /** Security group of the registry ECS service (for metrics ingress) */
@@ -409,7 +411,9 @@ export class ObservabilityPipeline extends Construct {
       'GrafanaExecRole',
       `${namePrefix}-grafana-task-exec-${region}`,
       grafanaLogGroup,
-      [],
+      // Issue #1325: the execution role must read the Grafana admin password
+      // secret (and KMS-decrypt it) for the container `secrets` valueFrom.
+      props.secretsAccessStatements,
     );
 
     const grafanaTaskRole = _createTaskRole(
@@ -458,11 +462,23 @@ export class ObservabilityPipeline extends Construct {
       GF_AUTH_ANONYMOUS_ENABLED: 'false',
       GF_AUTH_ANONYMOUS_ORG_ROLE: 'Viewer',
       GF_AUTH_DISABLE_LOGIN_FORM: 'false',
-      GF_SECURITY_ADMIN_PASSWORD: config.grafanaAdminPassword,
       GF_LOG_MODE: 'console',
       GF_LOG_LEVEL: 'info',
       GF_DASHBOARDS_MIN_REFRESH_INTERVAL: '10s',
     };
+
+    // Issue #1325: source GF_SECURITY_ADMIN_PASSWORD from Secrets Manager via
+    // the container `secrets` map instead of a plaintext env value, so it does
+    // not appear in the rendered task definition. Falls back to the plaintext
+    // env only if the secret was not provided (observability disabled paths).
+    const grafanaSecrets: Record<string, ecs.Secret> = {};
+    if (props.grafanaAdminPasswordSecret) {
+      grafanaSecrets['GF_SECURITY_ADMIN_PASSWORD'] = ecs.Secret.fromSecretsManager(
+        props.grafanaAdminPasswordSecret,
+      );
+    } else {
+      grafanaEnv['GF_SECURITY_ADMIN_PASSWORD'] = config.grafanaAdminPassword;
+    }
 
     const grafanaContainer = grafanaTaskDef.addContainer('grafana', {
       containerName: 'grafana',
@@ -470,6 +486,7 @@ export class ObservabilityPipeline extends Construct {
       essential: true,
       cpu: 512,
       memoryLimitMiB: 1024,
+      secrets: grafanaSecrets,
       environment: grafanaEnv,
       logging: ecs.LogDrivers.awsLogs({
         logGroup: grafanaLogGroup,
--- a/infra/lib/registry/constructs/registry-secrets.ts
+++ b/infra/lib/registry/constructs/registry-secrets.ts
@@ -33,6 +33,7 @@ export class RegistrySecrets extends Construct {
   public readonly auth0M2mClientSecret?: secretsmanager.Secret;
   public readonly metricsApiKey?: secretsmanager.Secret;
   public readonly otlpExporterHeaders?: secretsmanager.Secret;
+  public readonly grafanaAdminPassword?: secretsmanager.Secret;
 
   /** IAM statements granting GetSecretValue + KMS decrypt for all secrets above */
   public readonly accessStatements: iam.PolicyStatement[];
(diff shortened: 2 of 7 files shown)


aarora79:
Note on CI: the Checkov check flags one new advisory alert, CKV2_AWS_57 ("enable automatic rotation") on the new `grafana_admin_password` secret. This is an expected false-positive for a static, startup-consumed credential (auto-rotation would require a coordinated Grafana restart), and it carries the same `#checkov:skip` comment the four sibling static secrets in secrets.tf already use. Checkov runs soft_fail/continue-on-error in terraform-test.yml, so it is advisory and non-blocking. CodeQL passed.

Issue #847, opened 2026-04-17
feat: Add reverse proxy gateway support for A2A agents


aarora79:
### Problem Statement

The MCP Gateway Registry currently acts as a reverse proxy for MCP servers -- clients connect to `https://gateway/server-path/mcp` and nginx proxies requests to the backend MCP server with authentication, metrics, and session management. However, registered A2A agents do not get the same treatment. Agents are registered for **discovery only** -- once an agent is discovered, the calling agent must connect directly to the backend agent URL, bypassing the gateway entirely.

This means agent-to-agent traffic misses out on:
- Centralized authentication and scope enforcement
- Request metrics and observability
- Network isolation (agents behind private networks can't be reached)
- Consistent URL patterns for enterprise deployments
- Rate limiting and access control
- Fine-grained access control at per-agent and per-method level

### Proposed Solution

Extend the nginx reverse proxy configuration generation to include A2A agents alongside MCP servers. When an A2A agent is registered and enabled, the gateway should generate nginx location blocks that proxy traffic to the agent backend, using the same auth, metrics, and observability pipeline already in place for MCP servers.

**URL Pattern:**
```
https://mcpgateway.internal.acme.corp/agent/{agent-path}/                                --> POST (JSON-RPC)
https://mcpgateway.internal.acme.corp/agent/{agent-path}/.well-known/agent-card.json     --> GET (discovery)
```

Since A2A is JSON-RPC 2.0 over HTTP (same transport as MCP streamable-http), the existing proxy configuration handles it natively. No protocol translation is needed.

**Agent card URL rewriting:** When serving an agent card through the gateway, the `url` field must be rewritten from the backend's internal URL to the gateway's external URL, so that callers route through the gateway rather than bypassing it.

**Fine-Grained Access Control (FGAC):** Two levels of FGAC are supported:

| FGAC Level | Scope Pattern | Example | Enforcement Point |
|------------|---------------|---------|-------------------|
| **Per-agent** | `a2a-agents:{path}:invoke` | Allow calling `/travel-agent` but not `/hr-agent` | Auth subrequest (`/validate`) |
| **Per-method** | `a2a-agents:{path}:{method}` | Allow `message/send` but block `tasks/cancel` | Lua script parses JSON-RPC `method` field |

[...]

PR #1355, opened 2026-07-01
feat(a2a): add reverse-proxy gateway support for A2A agents

anjosluc:
## Summary

Adds reverse-proxy gateway support for A2A agents, so A2A **agent-card discovery** and **JSON-RPC traffic** are proxied through the nginx gateway (not just discovered). Implements the feature requested upstream in [agentic-community/mcp-gateway-registry#847](https://github.com/agentic-community/mcp-gateway-registry/issues/847).

The mode is **opt-in** via `A2A_REVERSE_PROXY_ENABLED` (default `false`) — registry-only discovery is unchanged for existing deployments.

## What changed

- **Opt-in feature flag** (`registry/core/config.py`, `registry/core/nginx_service.py`): new `a2a_reverse_proxy_enabled` setting (env `A2A_REVERSE_PROXY_ENABLED`, default `false`). Agent location blocks are only generated when it is enabled (and in `with-gateway` deployment mode). When off, `/agent/*` keeps the existing registry-only behavior.
- **nginx config generation** (`registry/core/nginx_service.py`): generate per-agent `location` blocks — one for the agent card (`/agent/<path>/.well-known/agent-card.json`) and one for the JSON-RPC endpoint (`/agent/<path>/`) — wired through the `auth_request /validate` subrequest (same hop as MCP servers). `path`/`url` are input-validated before interpolation. Templates gain a `{{AGENT_LOCATION_BLOCKS}}` placeholder.
- **Agent lifecycle → nginx refresh** (`registry/services/agent_service.py`): mark nginx dirty on enable/disable/update/delete so proxy routes track agent state.
- **Per-agent FGAC** (`auth_server/server.py`): `/validate` enforces invocation access via `validate_a2a_agent_access()`, which mirrors `validate_server_tool_access` — it resolves each caller scope through the scope repository and matches an `invoke_agent` action whose `resources` cover the agent path (`all`/`*` or exact path). Adds `_get_a2a_agent_path()`.
- **Scope seed** (`scripts/registry-admins.json`): adds an `invoke_agent` action (`resources: ["all"]`) to the admin scope; finer-grained per-agent grants are created as additional scope documents via the scope-management API.
- **Agent-card rewrite** (`docker/lua/agent_card_rewrite.lua`): rewrites agent-card URLs from the backend origin to the gateway origin (top-level `url` plus `additionalInterfaces`/`supportedInterfaces`) so clients route JSON-RPC back through the proxy.
- **Settings surface**: flag wired into `.env.example`, the Settings → System Config UI (`registry/api/config_routes.py` `CONFIG_GROUPS`), and `docs/unified-parameter-reference.md`.
- **Docs** (`docs/design/a2a-protocol-integration.md`): "Reverse-Proxy Mode" section, incl. how to enable the mode.

Per-method / per-skill FGAC is intentionally out of scope for this iteration: the structured model is action + resources, and the gateway can't resolve structured permissions in the request path.

## Tests

- `tests/unit/test_agent_nginx.py` (new): 25 tests for agent location-block generation, incl. the opt-in default (`A2A_REVERSE_PROXY_ENABLED=false`) — **all pass**.
- `tests/unit/core/test_nginx_service.py` (extended): render-level gate tests — blocks emitted when the flag is on, skipped when off — **all pass**.
- `tests/auth_server/unit/test_server.py` (extended): 16 A2A tests (`TestGetA2AAgentPath`, `TestValidateA2AAgentAccess`, structured/async) — **all pass**.
- `ruff check` clean on the added code.

## Notes

- No new Python dependencies.
- Changes are additive and opt-in (feature off by default; only `invoke_agent` appended to an existing scope seed).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1078,6 +1078,18 @@ AGENT_SECURITY_ADD_PENDING_TAG=true
 A2A_SCANNER_LLM_API_KEY=your_azure_openai_api_key_here
 
 # =============================================================================
+# A2A REVERSE-PROXY MODE
+# =============================================================================
+
+# Enable reverse-proxy mode for A2A agents (opt-in, default: false)
+# When true, each enabled agent gets nginx location blocks that proxy its A2A
+# traffic (agent card + JSON-RPC) through the gateway for centralized auth and
+# metrics, instead of clients connecting directly to the agent backend.
+# When false, /agent/* paths keep the registry-only discovery behavior.
+# Also requires with-gateway deployment mode.
+A2A_REVERSE_PROXY_ENABLED=false
+
+# =============================================================================
 # CONTAINER REGISTRY CREDENTIALS (for CI/CD and local builds)
 # =============================================================================
 
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -1981,6 +1981,97 @@ def _is_registry_api_request(
     return False
 
 
+def _get_a2a_agent_path(
+    original_url: str | None,
+) -> str | None:
+    """Return the agent path for an A2A reverse-proxy request, or None.
+
+    Recognizes URLs of the form ``{root}/agent/{agent_path}/...``
+    and returns the agent path with a leading slash (e.g. "/flight-booking-agent"),
+    matching the agent's registered path and the ``invoke_agent`` scope
+    resources. Agent paths may be multi-segment (e.g. "/lob1/travel"); the trailing
+    agent-card discovery suffix is stripped so the card and JSON-RPC requests
+    resolve to the same agent path. Returns None for any non-agent request so the
+    caller falls back to MCP handling.
+
+    Args:
+        original_url: The X-Original-URL header value from nginx.
+
+    Returns:
+        The agent path (leading slash, one or more segments) or None.
+    """
+    if not original_url:
+        return None
+
+    parsed = urlparse(original_url)
+    path = parsed.path.strip("/")
+
+    registry_prefix = REGISTRY_ROOT_PATH.strip("/")
+    if registry_prefix and path.startswith(registry_prefix):
+        path = path[len(registry_prefix) :].lstrip("/")
+
+    parts = path.split("/") if path else []
+    if len(parts) < 2 or parts[0] != "agent":
+        return None
+
+    agent_segments = parts[1:]
+    # Drop the agent-card discovery suffix so /agent/x/.well-known/agent-card.json
+    # and /agent/x/ both resolve to the same agent path.
+    if agent_segments[-2:] == [".well-known", "agent-card.json"]:
+        agent_segments = agent_segments[:-2]
+
+    if not agent_segments or not all(agent_segments):
+        return None
+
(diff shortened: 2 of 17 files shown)


aarora79:
Thanks @anjosluc. This implements #847 and matches the architecture we want. Three things line up with our design:

- Agent location blocks are generated inside `_render_config_impl` behind an `{{AGENT_LOCATION_BLOCKS}}` placeholder, so the reload scheduler's hash detects agent changes.
- Invocation FGAC is on the `/validate` hop via `validate_a2a_agent_access`, matching `validate_server_tool_access`.
- The path/url validation regexes and the exact-match card location prevent config injection and suffix smuggling.

We want to merge this as the foundation and add the rest on top. Notes below.

Keep as-is:
- The default-off flag `A2A_REVERSE_PROXY_ENABLED=false` is what we want. It lets us merge this dormant and enable it once the surrounding pieces land. Do not change it.

Changes needed before we can test and merge:

1. Rebase onto latest `main`. It conflicts with #1358 and #1361 (MCP proxy timeout), which changed the same files in different location blocks. The resolve is mechanical.

2. Gate the generator on health and protocol. `_generate_agent_location_blocks` currently emits a block for every enabled agent with a URL. Add two skips in the loop:
   - skip when the agent is not healthy, so we do not generate a proxy route to a dead backend.
   - skip when `supported_protocol != "a2a"`, so a non-A2A agent with a URL does not get a JSON-RPC proxy block.

3. Add one render-level test. The current tests mock `_generate_agent_location_blocks`, so the real render-to-hash path is never run. Add a test that enables an agent and asserts the rendered config hash changes. This guards against a future change breaking agent reloads.

Out of scope for this PR. We will do these in follow-ups. No action needed from you:
- Per-method and per-skill FGAC. The per-agent model extends to it; the JSON-RPC body is already at the `/validate` hop via `capture_body` and `X-Body`.
- Per-agent metrics attribution. `emit_metrics.lua` keys the metric name off the first URI segment, which is always `agent` for `/agent/<name>/` paths. We will fix that separately.
- `flush_now()` for immediate reload on enable, disable, and delete.
- Integration and E2E proxy tests.

After the three items are in and CI is green on current `main`, we will run a live test (enable the flag and one healthy A2A agent, confirm the card and JSON-RPC proxy through the gateway with an `invoke_agent` scope, and confirm a 403 without it) and merge.

anjosluc:
Thanks for the review @aarora79. All three items are in and pushed (`7dcd64b`); the default-off `A2A_REVERSE_PROXY_ENABLED` flag is untouched.

Added unit tests for both skips (unhealthy, unknown-health, `other` protocol, and missing protocol) in `tests/unit/test_agent_nginx.py`.

aarora79:
Thanks for this — merged as the A2A reverse-proxy foundation. We built on top of it in #1434, which includes your changes plus the egress trust model (X-Authorization/Authorization split), the symmetric {agent, actions} scope schema, gateway-aware discovery (url/proxy_pass_url split), and full deployment wiring. Both are now in main.

PR #1356, opened 2026-07-02
fix: make MCP proxy upstream timeout configurable (#1320)

### LINKED ISSUE #1320: MCP proxy hop upstream timeout is hardcoded to 30s (long-running tool calls fail with 504)

zsxh1990:
Fixes #1320. Added _read_mcp_proxy_timeout() following existing config pattern. MCP_PROXY_TIMEOUT env var, default 30s, min 1s. Tested: default unchanged, custom values work, invalid falls back with warning.

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -252,6 +252,36 @@ def _read_mcp_proxy_max_body_bytes() -> int:
     return max(candidate, minimum_bytes)
 
 
+def _read_mcp_proxy_timeout() -> float:
+    """Read the upstream MCP proxy timeout in seconds.
+
+    Resolution order:
+    1. ``settings.mcp_proxy_timeout`` (if the attribute exists)
+    2. ``MCP_PROXY_TIMEOUT`` environment variable
+    3. Default: 30.0 seconds
+
+    The minimum is 1 second to prevent accidental zero/negative values.
+    """
+    default_timeout = 30.0
+    minimum_timeout = 1.0
+    try:
+        value = getattr(settings, "mcp_proxy_timeout", None)
+        if value is not None:
+            candidate = float(value)
+            return max(candidate, minimum_timeout)
+    except (TypeError, ValueError) as e:
+        logger.debug(f"settings.mcp_proxy_timeout parse failed, falling back to env: {e}")
+    raw = os.getenv("MCP_PROXY_TIMEOUT")
+    if not raw:
+        return default_timeout
+    try:
+        candidate = float(raw)
+    except ValueError:
+        logging.warning(f"Invalid MCP_PROXY_TIMEOUT={raw!r}; using default {default_timeout}")
+        return default_timeout
+    return max(candidate, minimum_timeout)
+
+
 # Global scopes configuration (will be loaded during FastAPI startup)
 SCOPES_CONFIG = {}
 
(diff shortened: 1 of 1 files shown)


aarora79:
Follow-up #1358 (now merged) wires `MCP_PROXY_TIMEOUT` into all deployment surfaces so it is usable and discoverable beyond a hand-set env var:

- Added a `mcp_proxy_timeout` field to the registry `Settings` model (which also activates the reader's `settings.mcp_proxy_timeout` resolution branch).
- Surfaced it in `.env.example`, all three Docker Compose files, the Helm auth-server chart (values + secret + reserved-env-names), and Terraform/ECS (env var + module and root variables with a `>= 1` validation + `terraform.tfvars.example`).
- Added it to `config_routes.py` `CONFIG_GROUPS` so it appears in `GET /api/config/full` and the System Config UI.
- Documented it in `docs/unified-parameter-reference.md`.
- Noted the nginx ceiling: the generated `/mcp-proxy/` blocks inherit nginx's 60s `proxy_read_timeout`, so values above 60s also require raising that (documented; not plumbed).
- Minor: `logging.warning` -> `logger.warning` in both `_read_mcp_proxy_*` readers.

Verified end-to-end on all three surfaces: Docker Compose (`timeout=1.0` in the auth-server log after an `.env` override), the System Config UI, and Terraform/ECS (task-def env `MCP_PROXY_TIMEOUT=10` plus a live `timeout=10.0` log line on the deployed gateway).

One clarification for future readers: the value is passed to `httpx.AsyncClient(timeout=...)` as a single float, which httpx applies as per-operation timeouts (connect/read/write/pool), not a total wall-clock deadline. So a steadily-streaming upstream can run longer than the configured value without tripping it; the timeout fires on a stalled connect or a read gap that exceeds it. This matches the original 30s behavior; #1358 only made the value configurable.

Thanks @zsxh1990 for the fix.

PR #1362, opened 2026-07-03
fix: constrain secret characters to URI-safe, AWS-safe set (#1354)

### LINKED ISSUE #1354: Auto-generated secrets with special characters break deployment (RDS rejects `@`, DocumentDB/URI breaks on `+`, Keycloak init breaks on `%`)
## Summary

Three separate first-time-deployment failures in the Terraform `aws-ecs` stack share a single
root cause: **passwords are generated and consumed without a consistent, safe character set or
proper encoding.** Depending on which special character a randomly generated password happens to
contain, one of three things breaks:

| Char | Where it breaks | Failure |
|------|-----------------|---------|
| `@`  | `terraform apply` creating the Keycloak RDS Aurora cluster | RDS rejects the master password; apply fails after ~100 resources |
| `+`  | DocumentDB auth from the app containers | `+` decodes to a space in the Mongo connection URI; all services crash-loop on `Authentication failed` |
| `%`  | `init-keycloak.sh` during post-deploy | `curl -d` mis-parses `%` as a percent-escape; admin auth returns HTTP 500, Keycloak init fails |

Individually each one can block a deployment, and because whether they trigger depends on which
characters a randomly generated password happens to contain, a clean non-interactive deployment
can succeed or fail from one run to the next. Because the failures surface at different stages
(plan/apply, container runtime, and post-deploy scripting), they look unrelated but are the same
defect: **no single generator constrains the character set, and consumers that embed passwords
into URIs or form bodies do not encode them.**

## Severity

High. Any of the three can block a first-time deployment, and which one occurs depends on the
characters in the generated password. The RDS failure happens partway through a long apply; the
DocumentDB failure presents as a crash-loop that is not obviously credential-related; the Keycloak
failure stops post-deploy setup.

## Affected component

Terraform `aws-ecs` stack (initial-deploy password generation, the rotation Lambdas, and the
post-deploy `init-keycloak.sh` script).

---

### Error
```
Error: creating RDS Cluster (keycloak): operation error RDS: CreateDBCluster,
api error InvalidParameterValue: The parameter MasterUserPassword is not a valid password.
Only printable ASCII characters besides '/', '@', '"', ' ' may be used.

  with aws_rds_cluster.keycloak,
  on keycloak-database.tf line 48, in resource "aws_rds_cluster" "keycloak":
```

### Root cause
The generated password included `@`. RDS Aurora rejects `/`, `@`, `"`, and spaces in
`MasterUserPassword`. The password was produced from a charset that included `@`
(e.g. `string.ascii_letters + string.digits + '!@#%^&*'`).

---

### Symptoms
CloudWatch logs for `/ecs/mcp-gateway-v2-auth-server` and `/ecs/mcp-gateway-v2-registry` show a
continuous loop:
```
Waiting for MongoDB replica set at <cluster>.docdb.amazonaws.com:27017...
MongoDB not ready yet: Authentication failed., full error: {'ok': 0, 'errmsg': 'Authentication failed.', 'code': 18, ...}
```
ECS service events show tasks repeatedly failing container health checks and being replaced.

### Root cause
The generated DocumentDB admin password contained `+`. In a MongoDB/DocumentDB connection URI
(`mongodb://user:pass@host`), `+` is reserved and URL-decodes to a space, so the password the
driver sends differs from the one set on the cluster. Any password containing a URI-reserved
character (`+ @ / : ? # %` and space) must be percent-encoded, or it must be excluded at
generation time.

Note this is reachable via **two** independent paths, neither of which excludes `+`:

1. **Initial deploy:** the `terraform-setup` guidance at
   `.claude/skills/terraform-setup/SKILL.md:426` tells the operator to avoid `/ " @` and spaces
   for DocumentDB - but omits `+` (and `: ? # %`). The value that broke this deployment was
   generated from a charset that explicitly included `+`.
2. **Rotation:** the DocumentDB and RDS rotation Lambdas generate passwords via
   `get_random_password(ExcludeCharacters=...)`, and the exclusion set is `/@"'\` - which does
   **not** exclude `+`:
   - `terraform/aws-ecs/lambda/rotate-documentdb/index.py:98` (default `"/@\"'\\"`)
   - `terraform/aws-ecs/lambda/rotate-rds/index.py:98` (default `"/@\"'\\"`)
   - `terraform/aws-ecs/secret-rotation.tf:337` and `:384` (`EXCLUDE_CHARACTERS = "/@\"'\\"`)
   - documented in `terraform/aws-ecs/lambda/README.md:26` and `:55`

   So even after a manual fix, a future rotation can regenerate a `+`-containing password and
   reproduce the outage.

---

### Symptom
Post-deployment Step 5 (Initialize Keycloak) fails with `Error: Failed to authenticate with
Keycloak`, even though the admin password stored in SSM and Secrets Manager is correct. The token
request returns **HTTP 500**.

### Reproduction
```bash
KC=https://<keycloak-cloudfront-domain>
PW='nk3z0O-G%U(KBDOrd4cTZaON'   # example generated password containing %

# Fails - HTTP 500 (the script's current behavior)
curl -s -o /dev/null -w "%{http_code}\n" -X POST "$KC/realms/master/protocol/openid-connect/token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "username=admin" -d "password=$PW" -d "grant_type=password" -d "client_id=admin-cli"

# Succeeds - HTTP 200
curl -s -o /dev/null -w "%{http_code}\n" -X POST "$KC/realms/master/protocol/openid-connect/token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  --data-urlencode "username=admin" --data-urlencode "password=$PW" \
  --data-urlencode "grant_type=password" --data-urlencode "client_id=admin-cli"
```

### Root cause
`curl -d "field=value"` sends the body verbatim and assumes it is already URL-encoded. A password
containing `%` (here `%U`) is mis-parsed as a malformed percent-escape, so the literal password
never reaches Keycloak. The `get_admin_token` function in
`terraform/aws-ecs/scripts/init-keycloak.sh` used `-d`. Switching those form fields to
`--data-urlencode` fixes it. (Other `curl -d` calls in that script send JSON bodies with
`Content-Type: application/json` and are not affected.)

---

## Proposed Fix

Fix at every point where a password is generated or embedded, so no single character can break a
deploy:

1. **Constrain generated passwords to a URI-safe, AWS-safe character set** (the simplest reliable
   choice is alphanumeric plus a few safe symbols). This resolves Failures 1 and 2 at the source:
   ```python
   allowed = string.ascii_letters + string.digits + "!#%^*()-_=."
   # Excludes: + / @ " : ? # & spaces and other URI-reserved / RDS-rejected characters
   ```
   Apply this to:
   - The `terraform-setup` guidance (`.claude/skills/terraform-setup/SKILL.md:426`) - update the
     note from "avoid `/ " @` and spaces" to also list `+ : ? # &`.
   - The rotation Lambdas' `EXCLUDE_CHARACTERS`: extend from `/@"'\` to also exclude
     `+ : ? # & = %` and space, in both `secret-rotation.tf:337`/`:384` and the Lambda defaults at
     `rotate-documentdb/index.py:98` and `rotate-rds/index.py:98` (and update `lambda/README.md`).

2. **Encode passwords at every consumer that embeds them into a URI or form body**, as
   defense-in-depth (so a stray character can never silently corrupt the credential):
   - `init-keycloak.sh` `get_admin_token`: use `curl --data-urlencode` instead of `-d`
     (resolves Failure 3).
   - Anywhere a password is placed into a Mongo/DocumentDB connection URI: percent-encode it, e.g.
     `urllib.parse.quote_plus(password, safe="")`.

3. **Add lightweight validation so bad values fail fast**, before `terraform apply` provisions
   ~100 resources. Example on the DB password variables (`keycloak_database_password`,
   `documentdb_admin_password`):
   ```hcl
   validation {
     condition     = !can(regex("[/@\" +]", var.keycloak_database_password))
     error_message = "Password cannot contain: / @ \" + or spaces (RDS/URI restriction)."
   }
   ```

## Workarounds used during this deployment

- **RDS/DocumentDB (Failures 1 & 2):** generate an alphanumeric-only password and set it in
  `terraform.tfvars` before apply / reset the DocumentDB cluster password and secret to an
  alphanumeric value:
  ```bash
  python3 -c "import secrets,string; print(''.join(secrets.choice(string.ascii_letters+string.digits) for _ in range(24)))"
  ```
- **Keycloak (Failure 3):** patched `init-keycloak.sh`'s `get_admin_token` to use
  `--data-urlencode` (verified: HTTP 500 -> 200).

## Environment

- Deployment: Terraform `aws-ecs`
- Storage backend: `documentdb`
- Auth provider: `keycloak`

zsxh1990:
Fixes #1354. Three deployment failures from unsafe password characters.

Changes:
1. Rotation Lambdas: EXCLUDE_CHARACTERS now excludes URI-reserved and RDS-rejected chars
2. Terraform variables: validation on keycloak_database_password and documentdb_admin_password
3. init-keycloak.sh: --data-urlencode instead of -d for form fields

Tested: validation blocks passwords with @ + % / etc at plan time.


aarora79 (changes requested):
Thanks @zsxh1990. The direction is right and the Keycloak fix is correct, but there seem to be a couple of issues, and the PR only partially addresses #1354. Requesting changes.

**Blocker 1: `rotate-rds/index.py:98` does not compile.**

```python
exclude_chars = os.environ.get("EXCLUDE_CHARACTERS", "/@"'+:?#&!=% ")
```

The inner `"` is not escaped, so Python parses this as the string `"/@"` followed by an unterminated `'+:?#&!=% "`. `python -m py_compile terraform/aws-ecs/lambda/rotate-rds/index.py` fails with `SyntaxError: unterminated string literal`. The Lambda will not import, so RDS rotation is broken. The `rotate-documentdb` file has it right. Fix:

```python
exclude_chars = os.environ.get("EXCLUDE_CHARACTERS", "/@\"'+:?#&!=% ")
```

**Blocker 2: `secret-rotation.tf` lines 337 and 384 exclude the wrong characters.**

```hcl
EXCLUDE_CHARACTERS = "/@\"\\x27+:?#&!=% "
```

In HCL, `\\` is a literal backslash and `x27` is not an escape sequence, so the rendered env value is `/@"\x27+:?#&!=% ` (literal backslash, x, 2, 7). Confirmed with `terraform console`. The deployed Lambda therefore excludes `\ x 2 7` and does not exclude the apostrophe. Since this env value overrides the Lambda default, the apostrophe is not excluded at runtime. HCL double-quoted strings accept a literal apostrophe, so use:

```hcl
EXCLUDE_CHARACTERS = "/@\"'+:?#&!=% "
```

Please keep this value byte-for-byte identical to the Lambda default in both `index.py` files.

**Scope: this does not fully close #1354.**

Complete and correct in this PR:
- `init-keycloak.sh` `get_admin_token` switched to `--data-urlencode` (Failure 3).
- `rotate-documentdb` exclusion set.
- Terraform `validation{}` blocks on `keycloak_database_password` and `documentdb_admin_password`.

Not yet addressed (called out in #1354):
- `.claude/skills/terraform-setup/SKILL.md:426` initial-deploy guidance still omits `+ : ? # &`.
- DocumentDB connection-URI percent-encoding (`urllib.parse.quote_plus`) at the app consumer, the issue's defense-in-depth item.
- `terraform/aws-ecs/lambda/README.md` still documents the old `/@"'\` exclusion set.

Either complete these in this PR, or narrow the PR description so #1354 stays open for a follow-up.

**Requested before merge:**
1. Fix Blocker 1, then run `uv run python -m py_compile` on both Lambda files.
2. Fix Blocker 2 and confirm the rendered env value contains a real apostrophe.
3. Add a small unit test asserting both Lambdas' `exclude_chars` contains `+ @ / % : ? # & ! space '` (this would have caught both blockers).
4. Complete or explicitly defer the three remaining #1354 items above.

zsxh1990:
Thanks for the thorough review @aarora79! All blockers addressed:

**Blocker 1 — RDS SyntaxError:** Fixed the unescaped `"` in `rotate-rds/index.py:98`. Both Lambda files now compile cleanly (`python -m py_compile` passes).

**Blocker 2 — Terraform `\\x27`:** Replaced `\\x27` with literal `'` in both `secret-rotation.tf` entries (lines 337, 384). HCL double-quoted strings accept `'` directly — confirmed the rendered value is now byte-identical to the Lambda defaults.

**Unit test:** Added `tests/test_secret_rotation_chars.py`:
- Compile check for both Lambda `.py` files
- Asserts all 12 required chars (`/ @ " ' + : ? # & ! = % space`) present in defaults
- Asserts both Lambdas use byte-identical defaults

**Remaining #1354 items (deferred):**
- `SKILL.md` and `README.md` — ✅ updated with full exclusion set in this PR
- DocumentDB connection-URI percent-encoding (`urllib.parse.quote_plus`) — deferring to a follow-up PR; this is the app consumer layer, not the rotation/init scripts. Will keep #1354 open for it.

aarora79:
Pushed a small follow-up commit (fc227284) to this branch closing the two test-coverage gaps from review, so the guard test fully covers the exclusion-set contract:

- **Added `#` to `REQUIRED_EXCLUDED`.** The actual exclusion set and the `variables.tf` regex both include `#`, but the test's required set omitted it, so a regression dropping `#` from a Lambda default would not have been caught.
- **Added `test_terraform_env_matches_lambda_defaults`.** `secret-rotation.tf`'s env value is what actually runs in AWS; the Lambda `EXCLUDE_CHARACTERS` default only applies when the env var is unset. The suite cross-checked the two Lambda defaults against each other but not against the `.tf` env, so the two could silently drift. Both `.tf` occurrences are now asserted equal to the Lambda default.

All 6 tests pass, and I mutation-checked the new assertion (dropping `#` from one `.tf` occurrence fails it; restoring passes). Feel free to squash into your own history if you prefer.

PR #1411, opened 2026-07-06
remove releases notes from main repo README and add to mkdocs

### LINKED ISSUE #1410: Request - Remove "What's New" from README.md
Hi –

Would it be possible to remove the what's new from the main repository README.md makes it very difficult to get a brief overview of the repository structure when this is growing indefinitely with each release.

Release notes are already pushed to docs, and surfaced when creating GitHub releases which I think is enough.

ceng-p:
## Move `release-notes/` into `docs/` for MkDocs inclusion

Closes #1410 

Moves all release note files from the repository root into `docs/release-notes/` so they are served as part of the MkDocs documentation site. Trims the verbose inline release history from `README.md` and fixes every relative link inside the release note files.

**`mkdocs.yml`**
- Added `Release Notes` nav section with all versions listed in explicit descending order (replaces `include_dir_to_nav` auto-discovery, which gave unusable alphabetical ordering)
- Added `enable_git_follow: false` to `git-revision-date-localized` to suppress timestamp inversion warnings from the rename

**`docs/llms.txt`**
- Updated path label from `release-notes/` to `docs/release-notes/`

**`.claude/skills/release-notes/SKILL.md`**
- Updated all output path references to `docs/release-notes/`
- Step 7 now prepends the new version entry to the `mkdocs.yml` nav and includes it in the release commit, keeping the descending sort order intact


aarora79:
Follow-up #1421 adds redirect stubs at the old `release-notes/` paths so historical `blob/main/release-notes/<file>.md` links (old release notifications, emails, posts) don't 404 after this move, plus a small stale-path cleanup. The old directory becomes a frozen redirect archive. Flagging so it's clear the old-link gap was handled deliberately, not missed.

PR #1412, opened 2026-07-06
skill credential handling

omrishiv:
# Keep skill-parse credentials out of URLs and logs

## What changed

The `POST /api/skills/parse-skill-md` helper previously accepted the plaintext
credential used to fetch a private repository's `SKILL.md` as a **query
parameter** (`auth_credential`). Query strings are captured in web-server access
logs, proxy logs, browser history, and the registry's own audit log, so any
credential supplied through this endpoint was persisted in cleartext in several
places.

This change:

1. **Moves the credential to a request header.** `parse-skill-md` now reads the
   credential from the `X-Auth-Credential` request header instead of a query
   parameter. Headers are not recorded in access logs or the audit log's
   captured query parameters. The frontend registration form was updated to send
   the credential in this header. A credential passed the old way (in the URL) is
   simply ignored, so it can no longer be honored *or* logged.

2. **Hardens audit-log query-parameter masking (defense in depth).** The audit
   model's sensitive-parameter masking now combines an exact-match list with
   case-insensitive **substring** matching (`token`, `secret`, `credential`,
   `auth`, `api_key`, `key`, `password`, ...). Any current or future
   credential-shaped query parameter is masked by default (fail closed) rather
   than only the handful previously enumerated.

3. **Routes every diagnostic header dump through one shared, fail-closed
   redactor.** Several diagnostic dumps (the auth dependency's DEBUG header dump,
   an agents-endpoint header dump, the health monitor's outbound-probe header
   trace, the MCP client's server-header trace) previously used narrow exact-match
   allowlists — or none at all — and could log an `Authorization`, `Cookie`, or
   the new credential header in plaintext. They now route through the shared
   `redact_headers()` helper. That helper is strengthened to match a credential
   header both by exact name and by a case-insensitive substring in the name
   (`auth`, `token`, `secret`, `credential`, `cookie`, `jwt`, `bearer`,
   `session`, `key`, ...), so a variant credential header such as
   `X-Auth-Credential` — or an unenumerated future one — is masked by default.
   The skill scanner also no longer echoes a raw headers string (which can carry a
   bearer token) into an error message.

4. **Stops the auth server leaking a token prefix.** The auth server's
   `mask_token` previously emitted the first four characters of a masked token.
   For an opaque token (API key, PAT, session id) those characters are real
   key-space, which contradicts the "even a prefix is sensitive" rule the shared
   redactor follows, so `mask_token` now emits only a fixed marker and its
   header-name substring set is widened to match the shared helper.

## Why it is safer / fails closed

- Credential source removed from the URL: the endpoint only binds the credential
  from the header, so it cannot be smuggled through — or logged from — the query
  string.
- Masking is additive and fail-closed: a new sensitive parameter or header name
  is masked by default via substring matching; the trade-off is over-redaction of
  benign names (e.g. `primary_key`), never under-redaction of a secret.
- No behavior change to the parsing itself: the credential still reaches the
  fetch logic; only its transport and its treatment in logs changed.

## How it was tested

- Endpoint: the credential supplied via `X-Auth-Credential` is forwarded to the
  parsing service; a credential placed in the query string is NOT consumed as the
  credential.
- Audit masking: `auth_credential` and a range of variant names
  (`client_secret`, `x_api_key`, `session_token`, ...) are masked; ordinary
  params (`url`, `page`, `limit`) are logged verbatim.
- Header redaction: `Authorization`, `Cookie`, `X-Auth-Credential`, `X-Api-Key`,
  and exotic variants (`X-Jwt`, `X-Bearer`, `X-Session`, `X-Signing-Key`) are
  redacted by the shared helper, the health monitor's masker, and the auth
  server's `mask_headers`; non-sensitive headers pass through. `mask_token`
  returns a fixed marker with no substring of the token value.
- Existing unit suites for skills, audit, health, mcp-client, and the auth server
  remain green.


aarora79:
Pushed a follow-up commit (dc654459) to this branch addressing the review Should-Fix / Consider items, so the duplicated redaction controls can't silently drift:

- **B1 (Should-Fix): pinned the duplicated header-substring lists.** Added `test_header_substrings_match_shared_redactor` asserting `auth_server.server._SENSITIVE_HEADER_SUBSTRINGS` == `registry.common.log_redaction.SENSITIVE_HEADER_SUBSTRINGS`. They're duplicated because the two are separate deployables that can't import each other, and nothing caught drift; weakening one copy now fails CI. Mutation-checked (dropping `jwt` from one list fails the test; restoring passes).
- **B2/B3 (Consider): documented intent.** Comment on the audit `SENSITIVE_QUERY_PARAM_SUBSTRINGS` noting it deliberately differs from the header lists (header-only markers like `cookie`/`jwt`/`bearer`/`session`) and that `key` is intentionally broad (fail closed) so it isn't narrowed later. `KEEP IN SYNC` cross-reference comments on both header lists pointing at the guard test.

Full affected suites green (323 passed, +1 for the new test). No behavior change — comments + one test only.

One item left for the release notes, not code: the credential move from `?auth_credential=` query param to the `X-Auth-Credential` header is a **breaking change for direct/non-browser API callers** (a credential in the old query position is now ignored). Intended, but worth calling out.

PR #1427, opened 2026-07-08
feat(egress): add RFC 8707 resource indicator for custom OAuth providers

go-faustino:
## Summary

Adds optional support for an [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707) **resource indicator** on egress OAuth for `custom` providers. When a server's egress config sets `custom_resource`, the gateway sends it as the `resource` parameter on the authorize request **and** on both token grants (code exchange + refresh), binding the minted per-user token to a specific protected resource. Built-in providers carry no resource, so their flows are unchanged.

## Motivation

Some MCP resource servers mint **per-resource** tokens and reject a flow that omits the resource indicator. The concrete driver is **Atlassian's Rovo MCP** (`https://mcp.atlassian.com/v1/mcp/authv2`):

- Its Authorization Server fails the authorize step with *"Invalid context provided"* when `resource` is absent.
- Its MCP endpoint rejects a token whose audience is the generic `api.atlassian.com` REST API (what the built-in `atlassian` 3LO provider mints) rather than the MCP resource itself.

The existing engine had no way to emit `resource`, so Rovo MCP could not be wired as an egress target. This closes that gap generically (any RFC 8707 provider), without special-casing Atlassian.

## What changed

- `registry/egress_auth/schemas.py` — `OAuthProviderConfig.resource` (optional).
- `registry/egress_auth/oauth_engine.py` — emit `resource` on `build_authorize_url`, `exchange_code`, and `refresh_token` when set.
- `registry/egress_auth/providers.py` — `resolve_provider` threads `custom_resource` onto the assembled custom config.
- `registry/api/egress_auth_routes.py` — `EgressConfigRequest.custom_resource`, persisted and exposed in the `GET` view. Validated as an absolute `https` URI without a fragment. Unlike `custom_authorize_url`/`custom_token_url` it is **not** a request target (it is reflected verbatim into the authorize 302 and the token POST body), so it gets a structural check rather than the SSRF guard, and only for the `custom` provider.
- `docs/egress-credential-vault.md` — documents `custom_resource` with an Atlassian Rovo example.

## Testing

- New unit tests cover all three legs (authorize URL, exchange, refresh) both with and without a resource, provider resolution, and the config-route validation/persistence/exposure.
- `pytest tests/unit/egress_auth/ tests/unit/core/test_egress_config_validation.py tests/auth_server/unit/test_egress_mint.py tests/unit/test_egress_oauth_no_stdout_logging.py` → **184 passed**.
- `ruff format --check` clean on changed files; no new `ruff check` findings introduced.

## Follow-up

- [ ] ~~Built-in `atlassian-rovo` provider~~ — intentionally left out; the generic `custom_resource` path covers it and keeps the provider table minimal.
- [ ] Downstream chart wiring (egress-sync) to push `custom_resource` — tracked separately.


aarora79:
Hi @go-faustino thanks for this contribution, I will merge this but could say if you are able to invoke the tools provided by the Atlassian Rovo MCP server (not just list them). I think there is still some missing piece but would love to understand if you were already able to get this to work and how. Thanks.

aarora79:
As a follow-up I guess my specific question is did you implement **egress-side Dynamic Client Registration** to get Rovo to work wherein we have the gateway registering itself as a client at the resource server's AS (once per server, cached on the `egress_oauth` sub-doc), then running the existing per-user consent/vend flow against the discovered AS. This owuld be complementary to `custom_resource`, not a replacement.

go-faustino:
Thanks for merging @aarora79, and good instinct on both counts.

On invocation: honest status — I validated the flow end-to-end manually (DCR-registered client + PKCE + the resource indicator + full Rovo scopes): token exchange returned 200 (bearer + refresh), initialize succeeded, and tools/list returned the Rovo toolset. I have not yet confirmed a successful tools/call over that path, so I can't claim invocation works end-to-end — which matches your "still a missing piece" read.

On DCR: you're right that it's the complementary piece, and no — this PR intentionally doesn't implement it; it only threads the RFC 8707 resource param through the existing custom provider. To validate, I registered a client manually at Rovo's DCR endpoint (AS discovered from /.well-known/oauth-protected-resource on …/v1/mcp/authv2), then drove the custom provider with that client_id/secret + custom_resource.

Worth stressing why DCR is needed alongside this change: a pre-registered classic 3LO app (audience=api.atlassian.com) is not accepted by Rovo's authv2 — its tokens are minted for the generic REST API audience and get rejected, and the classic /authorize refuses the resource param (Invalid context provided). Only a client registered at Rovo's own AS via DCR, plus the resource indicator, produced a token Rovo would take. So custom_resource binds the token to the right resource, but a client the Rovo AS recognizes is still required — and today that registration is out-of-band.

First-class egress-side DCR — gateway self-registers once per server against the discovered AS, caches the client on the egress_oauth sub-doc, then runs the normal per-user consent/vend — would remove that manual step and pair naturally with this change. I scoped this PR to the resource indicator to keep it small; happy to follow up with a DCR design, and if you've gotten tools/call working I'd love to compare the client registration + scopes you used.

PR #1490, opened 2026-07-15
fix(egress): retry transient Vault/OpenBao failures on token save; surface hard save failures to the user

go-faustino:
## Problem

The per-user egress token store (`OpenBaoStore`) only retries on **expired-token** errors (`_is_auth_expiry_error`: Forbidden / 403 / "permission denied"). Transient **Vault/OpenBao unavailability** is not retried:

- `hvac.InternalServerError` 500 with `local node not active but active cluster node not found` (a standby that doesn't yet know the leader),
- `ConnectionError` / `connection refused` / `Max retries exceeded` (Service still routing to a terminating pod, or an HA 307 redirect to a now-dead leader `api_addr`),
- read timeouts.

All of these occur for a few seconds while an HA cluster **re-elects a leader** after a pod restart (eviction / rollout / spot reclaim). Today they fall straight through as `SecretStoreError`.

The worst case is the OAuth **consent callback**: `handle_callback` exchanges the code successfully and then calls `put_token`. If the write hits that window it raises, and `egress_callback` only catches `EgressAuthError` — so the request 500s and, critically, the token is **never vaulted**. The user is left with a silent **"0 tools"** and no signal that they should retry.

## Change

- **`registry/secrets/openbao/store.py`** — add `_is_transient_error` and extend `_run` with a bounded exponential backoff retry (`0.5/1/2/4s`, module constants so tests can shrink it). It is deliberately **disjoint** from the existing single re-auth path (403 is not transient), so the two recovery paths never overlap. Reads and writes now ride out a leader election instead of failing.
- **`registry/api/egress_auth_routes.py`** — catch `SecretStoreError` in `egress_callback` separately from `EgressAuthError` and return a clear, retryable **"Connection not saved"** page (`503`) instead of a raw `500`. It never falls through to the success page, so a failed write can't masquerade as a connected account.
- **`tests/unit/secrets/test_stores.py`** — new `TestOpenBaoTransientRetry`: retry-then-persist on write, retry-then-succeed on read, bounded give-up after N retries (no infinite loop), and that non-transient errors are not retried.

## Test plan

- [x] `pytest tests/unit/secrets/test_stores.py` — 22 passed (incl. 4 new).
- [x] `pytest tests/unit/egress_auth/test_public_routes.py` — 16 passed (callback route).
- [x] `py_compile` + line-length (≤100) clean on changed files.

### DIFF
--- a/registry/api/egress_auth_routes.py
+++ b/registry/api/egress_auth_routes.py
@@ -40,6 +40,7 @@ from registry.egress_auth.providers import list_provider_names, resolve_provider
 from registry.egress_auth.service import EgressAuthError, is_per_user_auth_method
 from registry.exceptions import UrlValidationError
 from registry.repositories.factory import get_server_repository
+from registry.secrets.interfaces import SecretStoreError
 from registry.services.server_service import server_service
 from registry.utils.credential_encryption import encrypt_credential
 from registry.utils.url_guard import PROXY_PROFILE, validate_url
@@ -706,6 +707,24 @@ async def egress_callback(
         return HTMLResponse(
             "<h3>Connection failed. Please close this tab and try connecting again.</h3>",
             status_code=400,
+        )
+    except SecretStoreError as exc:
+        # The code exchange SUCCEEDED but persisting the token to the secret store
+        # (Vault/OpenBao) failed — the store already retried transient blips (HA
+        # leader election / pod restart), so landing here means the write did not
+        # persist. Critically, DO NOT fall through to the success page: that would
+        # tell the user they are "Connected" while no token was vaulted, leaving
+        # them with a silent "0 tools" and no signal to retry. Surface a clear,
+        # retryable error instead. Detail to logs only (may wrap internal store
+        # addresses). 503 == transient/backing-store issue, please retry.
+        logger.error(
+            "egress callback: code exchange ok but secret store write failed: %s", exc
+        )
+        return HTMLResponse(
+            "<h3>Connection not saved.</h3>"
+            "<p>We couldn't store your connection because of a temporary storage "
+            "issue. Please close this tab and try connecting again in a minute.</p>",
+            status_code=503,
         )
 
     # The egress consent is the web Connected-Accounts / MCP URL-mode elicitation
--- a/registry/api/egress_auth_routes.py
+++ b/registry/api/egress_auth_routes.py
@@ -717,9 +717,7 @@ async def egress_callback(
         # them with a silent "0 tools" and no signal to retry. Surface a clear,
         # retryable error instead. Detail to logs only (may wrap internal store
         # addresses). 503 == transient/backing-store issue, please retry.
-        logger.error(
-            "egress callback: code exchange ok but secret store write failed: %s", exc
-        )
+        logger.error("egress callback: code exchange ok but secret store write failed: %s", exc)
         return HTMLResponse(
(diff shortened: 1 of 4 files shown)


aarora79:
Thanks for this. Reviewed and tested it live; merging.

I verified the retry against a real OpenBao outage (`docker stop`, which drops the container so the registry gets a genuine `ConnectionError` rather than a hung socket). The store logged the exact bounded backoff and then failed closed:

```
OpenBao request failed transiently (...); retry 1/4 in 0.5s
retry 2/4 in 1.0s
retry 3/4 in 2.0s
retry 4/4 in 4.0s
-> SecretStoreError: OpenBao get failed
```

Confirmed: transient classification catches `ConnectionError` / `Max retries exceeded`, the 0.5/1/2/4s backoff is bounded (no infinite loop), it fails closed after give-up, and reads recover to 200 once OpenBao returns. The consent-callback write path and its 503 "Connection not saved" page are covered by your unit tests; the change is clean and does what it says.

One follow-up worth considering, not blocking this merge:

The retry hardens the store, and the callback write now surfaces a clear 503. The **token-vend read path** (`auth-server -> /_egress_internal/egress-token -> registry -> OpenBao`) does not get the same graceful failure. When I forced the outage during a tool call, the registry retried correctly, but after give-up the auth-server logged `egress vend: registry unreachable` and proxied to the upstream **without a token, returning a 401** to the caller. That is arguably the same silent-failure class this PR set out to remove, just on the read side rather than the write side.

Two things to weigh for a follow-up:

1. The auth-server vend call uses a fixed `timeout=10.0` to the registry, while the registry retry budget is about 7.5s. It works today, but the two are coupled: if the retry budget ever grows past the vend timeout, the auth-server would give up before the retry finishes. Worth making that relationship explicit (a comment, or deriving one from the other).
2. On vend failure the auth-server returns `None` and falls through to a tokenless upstream call. Consider surfacing a clearer "connection temporarily unavailable, retry" signal to the caller instead of a raw upstream 401, mirroring the 503 you added on the write path.

Neither blocks this PR. The retry-and-fail-closed behavior is solid and tested.

go-faustino:
Great catch @aarora79, and thanks for testing it live against a real OpenBao outage — that's exactly the failure mode. You're right that the read (vend) path had the same silent-failure gap as the write path, just mirrored. I've opened #1529 to close it:

- **Explicit timeout ↔ retry-budget coupling.** The store now exposes `transient_retry_budget_seconds()` (the `0.5+1+2+4 = 7.5s` sum) as the single source of truth, and the auth-server derives its vend HTTP timeout from it (+ headroom) instead of the hard-coded `10.0`. A drift-guard test pins the relationship so the two can't silently decouple if the backoff constants ever change.
- **Retryable read-path failure, mirroring your write-path 503.** `vend_egress_token` now catches `SecretStoreError` and returns `503` (never a misleading `consent_required`, which would wrongly tell the user to reconnect). The auth-server classifies transport failures/timeouts and `5xx` as transient (`EgressVendUnavailable`) and the `mcp_proxy` hop returns a retryable `503` (JSON-RPC error + `Retry-After`) instead of forwarding tokenless → the raw upstream `401` you saw. Terminal `401/403/404` keep their existing fall-through behavior.

Tested the affected suites (`414 passed`) and added coverage for the transient classification, the 503, and the timeout coupling. Would appreciate a look when you have a moment.

PR #1491, opened 2026-07-15
fix(egress): key per-user egress vault on OIDC sub across consent + vend

go-faustino:
## Problem

The egress vault keys each per-user connection on a principal id resolved by `_canonical_egress_user`. The intended key is the OIDC `sub` (stable across an IdP's id_tokens **and** access tokens, unlike `preferred_username`, which some IdPs — e.g. Entra — omit from access tokens). In practice two paths disagreed on that id for the **same human**, so a user's token was written under one key and looked up under another → **"0 tools" despite holding a valid, non-expired token**.

Root cause: `generate_user_token` mints the gateway self-signed bearer that Cursor/Claude present on every MCP call, and its `sub` claim is the **login username/email**, not the OIDC `sub`. So at vend time `_canonical_egress_user` keyed the vault on the **email**, while the **browser-consent** path (registry resolves `session.subject`) keyed on the **OIDC `sub` UUID**. Same person, two buckets → whatever a user consented via the Connected-Accounts portal (UUID-keyed) was invisible to their bearer vend (email-keyed).

## Fix

**Align the primary key to the OIDC `sub` (auth_server).** Thread the session's persisted OIDC `sub` out of `_reconcile_context_against_session` and stamp it as an explicit `egress_user` claim on the self-signed USER token. `_canonical_egress_user` now prefers `egress_user` over the token's username `sub`, so the bearer-vend and browser-consent paths key the vault on the **same** id. Omitted for non-session / legacy callers (no OIDC sub) — those fall back to username unchanged.

## Transition (per review)

An earlier revision also carried a legacy-key fallback with migrate-on-read (vend under the old email key, copy to the canonical key, delete the old one). Per review discussion (thanks @omrishiv), that has been **dropped** (reverted in the last two commits, kept as reverts so it's easy to restore if we change our minds):

- This is a bug fix, not a long-lived schema migration — carrying stateful migrate-on-read code that must later be removed isn't worth it.
- The delete-on-migrate could leave a non-expiring orphan on failure; re-consent leaves an orphan anyway, so orphans want a separate cleanup (TTL / admin sweep) regardless.
- Re-consent is quick and self-service.

Net: **new** connections key on the OIDC `sub` consistently across consent and vend. The pre-cutover email-keyed cohort re-consents once (ideally paired with a clearer "reconnect this account" prompt when a server returns 0 tools). Smallest surface, nothing to remove later.

## Scope

This closes the deterministic key-mismatch cohort (consistent 0 tools). It is orthogonal to intermittent "0 tools" caused by transient Vault unavailability at vend time (leader election / restart windows), which is hardened separately (see #1490) and is not this PR.

## Tests

- `TestCanonicalEgressUser` (`tests/auth_server/unit/test_egress_mint.py`): `egress_user` wins over username `sub`; cookie `subject` and bearer `sub` fallbacks; username fallback preserved.
- `test_token_mint_session_reconcile.py`: reconcile now returns the session `subject`; body-asserted egress id and subject-less-session (empty) paths.
- `test_resource_bound_tokens.py`: session-backed `/internal/tokens` mint stamps the session OIDC `sub` as `egress_user` while keeping `sub`=username.
- `pytest tests/unit/egress_auth/test_service.py tests/auth_server/unit` — 102 passed locally after the reverts; `ruff format` + `check` clean.


omrishiv:
Hi @go-faustino, thanks for this PR. I was thinking about this issue and wondering if it's better to have to reconsent than add this fallback:

- if the delete fails, you have an orphaned record with a non-expiring key in your example case. I guess re-consenting would leave an orphaned record anyways
- the code could eventually be removed as we won't have to migrate. Is it worth pulling in for the transition period or is it enough to re-consent?

What do you think?

go-faustino:
Good question — agreed the migrate-on-read is the weakest part here, happy to reshape it. Splitting the two pieces:

The `sub`-alignment (stamping `egress_user`) I'd keep regardless — that's what stops *new* connections from splitting buckets.

The fallback is only about the pre-cutover cohort, with one nuance worth flagging: today the vend keys on the username/email, so the connections that *currently work* are the email-keyed ones. Once vend keys on `sub`, those session-minted bearers flip to a miss — so dropping the fallback means the users who work today are the ones re-consenting at cutover. Bounded and one-time, but a visible regression rather than a no-op.

On orphans — agreed, and re-consent doesn't really avoid it either: the old email-keyed entry is left behind regardless, so we want a cleanup (TTL or a one-off admin sweep) either way rather than leaning on delete-on-migrate.

So maybe three options:

- **A (this PR):** migrate-on-read — self-healing, no re-consent, but stateful + the delete-failure orphan you called out.
- **B:** pure read-through fallback — try `sub`, fall back to the legacy id, serve, no copy/delete. No orphan risk, no re-consent, and still removable once the legacy set ages out.
- **C:** drop the fallback, require re-consent for the pre-cutover email-keyed cohort. Smallest surface, nothing to remove later; tradeoff is those currently-working users re-consent once.

I lean **B** (addresses the orphan directly without regressing anyone), but I'm happy with **C** if you'd rather keep the surface minimal and treat re-consent as the transition story — ideally paired with a clear "reconnect this account" prompt when a server returns 0 tools. I'll trim the PR to the `sub` alignment + whichever fallback you prefer. What's your call?

omrishiv:
I'd lean towards C. This is a bug and adding code to mitigate it very briefly doesn't seem worth it. Reconsent is quick. @aarora79 what do you think?

go-faustino:
Went with **C**. Reverted the legacy-principal fallback + migrate-on-read in the last two commits (left them as `Revert` commits rather than squashing, so it's trivial to restore if we ever want the read-through variant instead).

The PR is now just the `sub`-alignment: `generate_user_token` stamps the OIDC `sub` as `egress_user` and `_canonical_egress_user` prefers it, so consent and vend key the same bucket. The registry service/route are back to baseline (no fallback) — the pre-cutover email-keyed connections re-consent once. 102 tests pass locally after the reverts; description updated.

Agreed on your orphan point regardless of direction — the stale email-keyed entries want a separate cleanup (TTL or an admin sweep) rather than delete-on-migrate; happy to follow up with that separately if useful.

auth_server/server.py
```diff
@@ -281,16 +281,25 @@ def _canonical_egress_user(validation_result: dict) -> str:
...
          at login (see create_session ``subject``).
-      2. top-level ``sub`` -- direct-token paths that surface it there.
-      3. ``username`` -- fallback for callers with no sub (keeps pre-existing
+      3. ``data.subject`` -- the cookie session's persisted OIDC sub.
+      4. top-level ``sub`` -- direct-token paths that surface it there.
+      5. ``username`` -- fallback for callers with no sub (keeps pre-existing
          non-OIDC behavior unchanged; only OIDC callers change bucket).
     """
     data = validation_result.get("data") or {}
     return (
```
omrishiv, line 300: Can we update to
```
# Only trust egress_user from tokens we minted ourselves
if validation_result.get("method") == "self_signed":
    egress_user = data.get("egress_user")
else:
    egress_user = None

return (
    egress_user
    or data.get("sub")
    or data.get("subject")
    or validation_result.get("sub")
    or validation_result.get("username")
    or ""
```

omrishiv (changes requested):
Hi @go-faustino sorry for the delay. I've reviewed and left a comment. Please also update the branch to the latest

go-faustino:
Heads-up: this merged while @omrishiv's requested change (gate `egress_user` to `self_signed` tokens) was still open, and it didn't make it into the merge commit — `_canonical_egress_user` on `main` still trusts `data.get("egress_user")` from any validated token.

I've applied it exactly as suggested in #1530: `egress_user` is now only honored when `method == self_signed` (a token this gateway minted), so a `jwt`/M2M/other-issuer token can't inject the vault key and vend another user's egress token. No behavior change for the vend path (the Cursor/Claude bearer is `self_signed`) or the consent-write path (keys on `sub`/`subject`).

While there I also found `test_egress_mint.py` had two classes both named `TestCanonicalEgressUser` — the first was shadowed and never collected; renamed it so both suites run. @omrishiv mind taking a look at #1530?

PR #1493, opened 2026-07-16
Skill discovery parity

omrishiv:
Builds off of #1492; merge first. 

# Skill discovery parity, a canonical asset-permission model, and per-record custom-entity grants

## Summary

Brings **skills** to authorization parity with servers/agents, unifies the
per-asset permission model behind a single canonical helper, fixes a latent
cross-asset scope bug, and extends **custom-entity** discovery grants to the
per-record level. Net effect: every first-class asset family (servers, agents,
skills, custom entities) is discovered and mutated through one consistent,
fail-closed model, and admins can grant a group either a whole family/type or
specific resources.

## Problem

- **Skills had no discovery gate.** `list_skills_for_user` and search filtered
  skills only by per-record visibility, so *every* user (including anonymous)
  could discover public skills. There was no `list_skills` analogue to
  `list_service` / `list_agents`.
- **Per-asset scope checks were open-coded and drifting.** Each family spelled
  its scope names differently (`list_service` singular, `list_agents` plural,
  `register_`/`publish_`/`create_` for "create") and re-implemented the
  `"all"`-or-named check inline. That drift hid a real bug: **agent modify/toggle
  enforced the SERVER scopes** `modify_service`/`toggle_service`, so a server
  grant leaked agent control while the intended `modify_agent`/`toggle_agent`
  grant was ignored (admins masked it, so it went unnoticed).
- **Custom-entity discovery was all-or-nothing per type.** `list_<type>_entity`
  could only open a whole type; there was no way to grant a group specific
  records, unlike `list_agents` which can hold specific agent paths.

### Canonical asset-permission model
- New `registry/auth/asset_permissions.py`: the single `(family, action) ->
  on-disk scope-name` map plus `user_has_asset_permission` /
  `accessible_resources_for`. The irregular persisted names live in one place
  (deliberately **not** normalized — renaming a persisted scope key is a breaking
  data migration, not a refactor). Discovery projections and the
  server/agent/custom-entity mutation gates route through it.
- **Cross-asset bug fixed:** agent modify/toggle now resolve the agent family's
  scopes via the map (call sites pass a logical action, not a scope string), so a
  sibling family's scope can never be substituted again. `toggle_agent` and the
  skill scopes are added to the seed groups and the IAM UI so the
  newly-enforced scopes are grantable.

### Skill discovery gate
- `list_skills` now gates skill discovery in `list_skills_for_user` and search,
  layered on top of (not replacing) the per-record visibility check; a caller
  with no `list_skills` grant sees zero skills — including public ones — matching
  `list_service`. `accessible_skills` is wired into the derived user context.
  Fail-closed.

### Skill mutation enforcement (was seeded/grantable but inert)
- `modify_skill` / `delete_skill` / `toggle_skill` were in the canonical map,
  seeded to the admin groups, and shown in the IAM UI, but the routes only
  checked owner-or-admin — the grants did nothing. The mutation routes
  (`update`, `delete`, `toggle`, plus `refresh_resources`) now enforce the
  canonical scope via `_user_can_modify_skill(skill, ctx, action=...)`: a
  **dual gate** of `scope AND (admin OR owner)`, matching server/custom-entity
  mutation. This also completes canonical-map adoption (skills no longer
  open-code their checks).
- **Skill-management scopes are excluded from admin derivation.**
  `publish_/modify_/delete_/toggle_skill` are now excluded in
  `is_admin_conferring_action` (same treatment as per-type entity scopes), so an
  admin can grant a non-admin "skill-manager" group the full skill scope set
  without silently promoting it to full registry admin. `publish_skill` is
  included in the exclusion so the full seeded set can't re-acquire admin via the
  create scope.

### Agent delete tightened to match the other families
- Agent delete was the lone outlier: `admin OR delete_agent-scope OR owner` — an
  owner could delete **without** the `delete_agent` scope. It now requires
  `delete_agent` scope AND (admin OR owner), uniform with server/skill/
  custom-entity delete. Fail-closed tightening; see upgrade notes.
- **Batch path brought in line.** `POST /agents/batch`'s per-item authorizer
  (`agent_batch_item_processor._authorize`) previously enforced the OLD logic —
  delete on owner-or-admin with **no** `delete_agent` scope, and patch/replace on
  the cross-asset `modify_service` scope — so a bulk op bypassed both the
  agent-delete tightening and the cross-asset fix. It now routes delete/modify
  through the same `user_has_asset_permission("agent", ...)` helper as the
  single-card handlers (packaging the submitter's persisted `is_admin` +
  `ui_permissions` snapshot into a minimal context), so batch and single-card can
  never diverge again.
- **Single-card delete unified on the canonical helper.** The delete endpoint now
  calls `_check_agent_permission("delete", agent.name, ...)` like modify/toggle,
  and the legacy `_has_delete_agent_permission` helper is removed. Previously that
  helper matched a narrow `delete_agent` grant against the agent PATH while the
  canonical helper (modify/toggle/batch) matched the NAME; all agent surfaces now
  resolve the grant identically (by name). Identical for `["all"]` grants.

### Per-record custom-entity discovery
- `list_<type>_entity` is now three-tier (parity with `list_agents`): `"all"` or
  the bare type name open the whole type (backward compatible); a record path
  `/<type>/<uuid>` opens just that record. Predicates `list_grant_allows_type` /
  `list_grant_record_paths` in `custom_entity_scopes.py` are the shared naming
  SSOT used by every enforcement site. The collection list intersects granted
  paths with the visibility filter, single-record GET/rate check the specific
  path, and search checks each hit's own path. **Granting one record never
  exposes other (even public) records of the type**, and the discovery grant and
  the record's own visibility must both pass (defense-in-depth). Existing
  `"all"` grants and the backfill are unchanged.

### Frontend
- Shared `UiPermissionEditor`: checkboxes for mutation scopes, an "All" toggle +
  multi-select for discovery scopes. Custom-type discovery uses
  `CustomTypeListPicker` — an "All" toggle plus a searchable picker of the type's
  actual records (record **name** shown, record **path** stored).
- Entity tabs (Servers / Agents / Skills / each custom type) are gated on the
  matching `list_` scope (admin bypass), redirecting to Discover when a hidden
  tab is reached.

## Authorization model after this PR (for reference / follow-up discussion)

Current mutation gates across all four families, non-admin caller (admin bypasses
every check). Ownership is `registered_by` for server/agent, `owner` for
skill/custom-entity. `scope` = the canonical per-resource scope (`"all"` or the
named resource). `accessible` = the `list_<family>` discovery projection cached in
`user_context`.

| Operation | Server | Agent | Skill | Custom entity |
|-----------|--------|-------|-------|---------------|
| **modify** | `scope AND owner` | `scope AND owner` | `scope AND owner` | `scope AND owner` |
| **delete** | `scope AND owner` | `scope AND owner` ⬅ tightened here | `scope AND owner` | `scope AND owner` |
| **toggle** | `scope AND (accessible OR —)` | `scope AND (accessible OR owner)` | `scope AND owner` | *no toggle* |

The `POST /agents/batch` path enforces the SAME gates as the single-card agent
endpoints (its per-item authorizer routes through the same canonical helper), so
a bulk op can't bypass what the single-card endpoint enforces. The matrix applies
to both surfaces.

Notes / known non-uniformities (deliberately **not** changed in this PR, tracked
for a follow-up):
- **Toggle's second gate is inconsistent.** Server/agent toggle admit a non-owner
  who holds an `accessible_*` (i.e. `list_*` discovery) grant; skill toggle is
  owner-only; custom entities have no toggle. This couples a mutation to a *read*
  scope and differs per family. Skills have no `accessible_skills` toggle path
  today. Deferred — needs a design decision on whether toggle should be owner-only
  everywhere (kill the read-scope coupling) or admit `accessible` everywhere.
- After this PR, **modify and delete are uniform (`scope AND owner`) across all
  four families**; only toggle remains non-uniform.
- **Narrow grants key on resource NAME, not path.** The canonical helper matches
  a narrow per-resource grant (e.g. `delete_agent: ["foo"]`) against the agent
  name. Agent names are not unique (only paths are), so a narrow name grant can
  match two same-named agents at different paths. This is pre-existing behavior
  for modify/toggle; delete now shares it (previously delete alone keyed on path).
  Bounded by the ownership AND-half and only relevant for narrow grants
  (`["all"]` unaffected). A future follow-up could key all asset-permission checks
  on the unique path family-wide — a cross-family design change, out of scope here.

## Upgrade notes (behavior changes)

- **Skills become gated (discovery).** Run
  `scripts/backfill-skill-list-scope.py --apply` once after upgrade (dry-run by
  default) to grant `list_skills: ["all"]` to `mcp-registry-admin`. Until an admin
  grants `list_skills` to other groups, skills are admin-only — including public
  ones. Intended parity with `list_service`.
- **Skill mutation now requires the skill scopes (owner self-service change).** A
  non-admin who could previously modify/delete/toggle a skill they **own** now
  also needs `modify_skill`/`delete_skill`/`toggle_skill` (grantable via the IAM
  UI, seeded to the admin groups). This is NOT covered by the discovery backfill
  and admins are unaffected (they bypass scope checks); grant the mutation scopes
  to whichever non-admin group is expected to manage skills.
- **Agent delete now requires `delete_agent` (owner self-service change).** A
  non-admin who owned an agent could previously delete it without the
  `delete_agent` scope; now the scope is required (AND ownership). No data
  migration — every admin surface (both seed groups, Helm configmap) already
  carries `delete_agent`. Grant `delete_agent` to any non-admin group that should
  delete agents it owns.
- **Agent modify/toggle now require the agent scopes.** A non-admin who managed
  agents via `modify_service`/`toggle_service` must be granted
  `modify_agent`/`toggle_agent` instead (both now grantable via the IAM UI and
  seeded to the admin groups).
- **Migration scripts unchanged.** Both backfills
  (`backfill-custom-entity-scopes.py`, `backfill-skill-list-scope.py`) are correct
  as-is: the mutation and admin-derivation-exclusion changes need no data
  migration (admins bypass; non-admin grants are an explicit IAM-UI decision per
  the discovery model).
- Custom entities: no migration needed — `"all"`/type-name grants keep their
  meaning; the per-record tier is additive.

## Tests

- Canonical map (family/action name resolution, custom-entity delegation,
  fail-closed).
- Skill discovery gate (no grant sees nothing incl. public; named/`all` grants;
  admin bypass; layered on visibility).
- Agent-scope fix (a server scope no longer satisfies an agent action).
- Skill mutation dual gate: admin bypass, owner+scope allow, owner-without-scope
  deny, scope-without-ownership deny, per-action scope resolution, per-resource
  grant, and endpoint-level regression tests proving `delete`/`toggle` wire their
  own action (a modify-only grant is denied on delete/toggle end-to-end).
- Skill-management scopes non-admin-conferring: verified through the real
  derivers (`_user_is_admin`, repo `_grants_admin`), a mixed-grant case (the
  exclusion doesn't poison a context that also holds a genuine admin scope), and
  that genuine admin actions still confer.
- Agent delete tightening: owner-without-scope now denied, scope-without-ownership
  denied, owner+scope succeeds, admin bypass intact.
- Per-record custom entities: the three-tier predicates, the `record_path` gate
  (one grant allows only that record, never a sibling), a search-route
  regression test guarding the whole-type-cache short-circuit against leaking
  sibling public records, and MongoDB integration tests for the `$and`/`$in`
  path+visibility filter (including that a granted-but-not-visible record is
  still excluded).
- Tab-gating and backfill coverage.


aarora79:
Reviewed the authorization model changes and they align with the direction I want: every mutation gate across servers, agents, skills, and custom entities is now `scope AND (admin OR owner)`, with the agent-delete outlier (previously `scope OR owner`, so a `delete_agent: ["all"]` grant could delete agents the caller never created) tightened to match, and the cross-asset bug fixed (agent modify/toggle were enforcing `modify_service`/`toggle_service`). Confirmed the single-card delete and the batch item authorizer both land on the same dual gate through the canonical helper, so bulk ops cannot be used as a bypass. This is the model I want: keep ownership, enforce it uniformly. Approving on that axis.

One naming cleanup I would like as a follow-up (separate PR, not blocking this one). The discovery scopes are inconsistent: `list_agents` and `list_skills` are plural, but `list_service` is singular. Since this PR just created the single chokepoint (`asset_permissions.py`) that every discovery/enforcement site routes through, we can canonicalize to `list_services` and accept the legacy `list_service` as an alias in the helper, so existing stored grants keep working with no data migration. We would update the JSON examples, seed groups, Helm configmap, and IAM editor to emit the new name. The mutation verbs are already consistent (`register/modify/delete/toggle_service` line up with the agent/skill verbs), so this is a one-scope change.

Note for whoever picks up that follow-up: the alias is only safe if every site reads the scope through the canonical helper and nothing calls `ui_permissions.get("list_service")` directly, so that needs verifying first. There is also a deeper naming question (the server family uses "service" in scope names while the UI calls it "server") that I would keep out of scope here.

Remaining items unchanged from the earlier pass: patch coverage is 90.8% with the uncovered lines concentrated in the enforcement paths themselves (server_routes, search_routes, skill_routes), which for a security change are the branches most worth covering before merge.

aarora79:
Merging this as-is. Follow-up (separate small PR): add a UI hint on the Skills empty state so a user without `list_skills` sees "ask your registry admin to grant skills access" instead of the current misleading "No skills are registered yet." Will pair it with a release-notes breaking-change entry noting skills stay hidden until `backfill-skill-list-scope.py --apply` runs.

aarora79:
Tracking the two follow-up items (UI access hint for empty Skills/Agents tabs + upgrade doc) via #1511.

PR #1497, opened 2026-07-16
fix(mcpgw): allow reverse-proxy Host so airegistry-tools stays healthy (FastMCP 3.x 421)

aarora79:
## Symptom

Connecting to the built-in **airegistry-tools** MCP server (e.g. from Claude Code) fails with **HTTP 405** at `/airegistry-tools/mcp`, and the server shows **Unhealthy** in the UI.

## Root cause

**PR #1021** ("multi-replica airegistry-tools support") moved mcpgw to **FastMCP 3.x** (`fastmcp>=3.2.0,<4.0.0`) and switched to `stateless_http=True`. FastMCP 3.x enables **host/origin DNS-rebinding protection by default**, allowing only `127.0.0.1` / `localhost` as the request `Host`.

But mcpgw is always reached through the registry's nginx/auth front door via `proxy_pass http://mcpgw-server:8003/`, so the `Host` is `mcpgw-server`. FastMCP rejects that with **421 Misdirected Request**, which cascades:

```
Host: mcpgw-server:8003 → 421
  → registry health check fails ("session initialization failed")
    → server marked Unhealthy
      → nginx drops the /airegistry-tools/ location block (only healthy servers get one)
        → request falls through → 405 Method Not Allowed
```

Latent since the **1.26.0** cycle (the change is in both 1.26.0 and 1.27.0). It only surfaces when a client actually connects, since it's health-check-gated.

Reproduced directly: `POST /mcp` with `Host: localhost` → 200, with `Host: mcpgw-server:8003` → 421.

## Fix

Configure FastMCP's transport security explicitly at startup in `servers/mcpgw/server.py`. **Protection stays ON**; the default allowlists `mcpgw-server` — the service name used **identically across all three surfaces**:

| Surface | How registry reaches mcpgw |
|---|---|
| Docker Compose | service `mcpgw-server` |
| Kubernetes/Helm | Service `mcpgw-server` (`charts/mcpgw` `app.name`) |
| ECS | Service Connect `dns_name = "mcpgw-server"` |

This matches the value the registry's SSRF guard already trusts as the built-in proxy target (`registry/utils/url_guard.py::_BUILTIN_PROXY_ALLOWED_HOSTS = {"mcpgw-server"}`), so **one code default fixes every deployment surface with zero per-surface config**.

- **Default** (`MCPGW_HTTP_ALLOWED_HOSTS=mcpgw-server`): `host_origin_protection=True`, `allowed_hosts=["mcpgw-server"]` (FastMCP always also allows loopback; the port is stripped before matching).
- **Override**: operators fronting mcpgw under a different name set the comma-separated `MCPGW_HTTP_ALLOWED_HOSTS` (e.g. `my-mcpgw.svc`). It's an optional extra-env var, deliberately NOT added to the reserved-env lists so it stays overridable.
- **Escape hatch**: `MCPGW_HTTP_ALLOWED_HOSTS=*` disables protection for the discouraged unpredictable-Host case.

## Verification (end-to-end, local stack)

```
POST /mcp  Host: mcpgw-server:8003   → 200   (was 421)
POST /mcp  Host: evil.example.com    → 421   (protection still enforced)
airegistry-tools health              → HEALTHY   (was unhealthy)
nginx /airegistry-tools/ location    → present   (was commented out)
front-door /airegistry-tools/mcp     → route live (was 405)
```

## Tests

7 unit tests for `_resolve_allowed_hosts` (default / `*` / empty / whitespace / comma-list / `*`-wins / custom host). Isolated via a fixture that snapshots + restores `sys.modules` so the shared `fastmcp` stub cannot leak across sibling mcpgw test files (verified in both collection orders). Full `tests/unit/servers/mcpgw/` suite: **47 passed**.

## Surface footprint

Server-side only — one change fixes Compose, Helm, and ECS because they all use the `mcpgw-server` name. No chart/Terraform edits required; the override env rides the existing `extraEnv` / `extra_env` / `mcpgw_extra_env` mechanisms.

### DIFF
--- a/servers/mcpgw/server.py
+++ b/servers/mcpgw/server.py
@@ -39,6 +39,27 @@ logger.info(
 REGISTRY_URL = os.getenv("REGISTRY_BASE_URL", "http://localhost")
 REGISTRY_EXTERNAL_URL = os.getenv("REGISTRY_EXTERNAL_URL", "")
 
+# Host allowlist for FastMCP's DNS-rebinding protection (streamable-http).
+# FastMCP 3.x enables host/origin protection by default, allowing only
+# 127.0.0.1 / localhost. But mcpgw is always reached through the registry's
+# nginx/auth front door (proxy_pass http://<service-name>:8003/), so the request
+# Host is the container/service name, which the default rejects with 421 -> the
+# registry health check fails -> nginx drops the /airegistry-tools/ location ->
+# clients get 405.
+#
+# The fix keeps protection ON and allowlists the exact service name the gateway
+# uses to reach mcpgw. That name is "mcpgw-server" across every surface (Docker
+# Compose service name, ECS Service Connect alias, Kubernetes Service name) --
+# the same canonical value the registry's SSRF guard trusts as the built-in
+# proxy target (registry/utils/url_guard.py::_BUILTIN_PROXY_ALLOWED_HOSTS). The
+# port is stripped by FastMCP before matching, and 127.0.0.1/localhost are
+# always allowed by FastMCP's built-in defaults, so those are not listed here.
+# Operators who front mcpgw under a different name override the comma-separated
+# MCPGW_HTTP_ALLOWED_HOSTS (e.g. "my-mcpgw.svc"); "*" disables protection
+# entirely for the (discouraged) case where the Host is unpredictable.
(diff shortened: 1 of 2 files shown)


aarora79:
Fixes #1498

Issue #1502, opened 2026-07-16
bug: Entra logout fails with AADSTS90015 (query string too long) for users with many group memberships


zoltan-fedor:
## Summary

Signing out when authenticated via Entra ID fails with Microsoft's error page:

> AADSTS90015: Requested query string is too long.

The `oauth2_logout` handler builds the Entra logout URL and unconditionally appends `id_token_hint=<JWT>` to the query string. For enterprise users who belong to many Entra groups, the ID token can exceed 1500–3000+ characters.
Combined with `post_logout_redirect_uri`, the total query string breaches Entra's length limit and the logout request is rejected. Because the local session is cleared before the IdP redirect, the user is left in a broken state: locally logged out but unable to complete the Entra OIDC logout.

## Steps to reproduce

1. Deploy with Entra ID as the auth provider.
2. Authenticate as a user who is a member of many Entra groups (≥ ~15 groups    is typically sufficient to produce a token that exceeds the limit; the exact threshold depends on group name lengths and other claims in the
   token).
3. Click Sign Out.
4. The browser reaches `https://login.microsoftonline.com/<tenant>/oauth2/v2.0/logout?post_logout_redirect_uri=...&id_token_hint=<very long JWT>` and Entra returns the AADSTS90015 error page instead of processing the logout.

## Root cause

`auth_server/server.py::oauth2_logout` unconditionally adds `id_token_hint` to the Entra logout query string without checking the resulting length:

```python
elif logout_hostname == "login.microsoftonline.com" or "entra" in provider.lower():
    logout_params = {
        "post_logout_redirect_uri": full_redirect_uri,
    }
    if id_token_hint:
        logout_params["id_token_hint"] = id_token_hint   # no length guard
```

Entra ID tokens are JWTs whose payload contains user claims including group memberships. In enterprise environments where users belong to many Entra groups, these tokens routinely exceed 2000 characters. The `urlencode` of the full logout query string then breaches Entra's query string length limit, causing AADSTS90015.

## Expected behaviour

Logout completes successfully regardless of token size. If `id_token_hint` would push the query string over a safe threshold it should be omitted — Entra still processes the logout correctly without it. The only user-visible difference is that multi-account users may see a brief account-selection prompt from Entra rather than a silent logout.

## Proposed fix

Add a query-string length guard in the Entra branch before adding `id_token_hint` to the logout params:

```python
elif logout_hostname == "login.microsoftonline.com" or "entra" in provider.lower():
[...]

PR #1509, opened 2026-07-17
fix: internalize OAuth2 logout id_token_hint (#1503)

### LINKED ISSUE #1503: enhancement: Logout id_token_hint JWT passed in browser-facing URL is fragile and an information disclosure risk
## Summary

During OAuth2 logout, `registry/auth/routes.py::logout_handler` redirects the browser to:

```
https://<domain>/oauth2/logout/<provider>?redirect_uri=<uri>&id_token_hint=<JWT>
```

The `id_token_hint` is the full OIDC ID token — a JWT typically 1000–3000+ characters long. Routing it through a browser redirect means the token:

1. **Passes through any WAF or reverse proxy** sitting in front of the application. WAFs with SQL-injection, XSS, or query-string size-limit rules commonly block this request with 403 before it reaches the auth server, causing logout to fail silently (the local session has already been cleared at this point).
2. **Appears in browser history**, server access logs, and any proxy or CDN logs — a mild information disclosure for a credential that is being actively revoked.

## Steps to reproduce (WAF scenario)

1. Deploy behind a WAF that inspects query strings (e.g. AWS WAF with `AWSManagedRulesCommonRuleSet` or `AWSManagedRulesSQLiRuleSet`, ModSecurity,  Cloudflare, etc.).
2. Authenticate via any OAuth2 provider.
3. Click Sign Out.
4. The browser is redirected to `/<domain>/oauth2/logout/<provider>?...&id_token_hint=<JWT>`.
5. The WAF blocks the request — 403 is returned before the auth server is reached. The local session was already cleared, so the user cannot log back in without manually clearing cookies, and the IdP OIDC session is never   terminated.

## Root cause

`logout_handler` uses `settings.auth_server_external_url` (the public-facing hostname, behind the WAF) to build the browser redirect, embedding the JWT directly in the query string:

```python
# registry/auth/routes.py
auth_external_url = settings.auth_server_external_url
logout_url = f"{auth_external_url}/oauth2/logout/{provider}?redirect_uri={redirect_uri}"
encoded_token = urllib.parse.quote(id_token, safe="")
logout_url = f"{logout_url}&id_token_hint={encoded_token}"
response = RedirectResponse(url=logout_url, ...)
```

The `id_token_hint` only needs to reach `oauth2_logout` in `auth_server` so it can be forwarded to the IdP. There is no reason for it to travel through the browser or any public-facing infrastructure.

## Expected behaviour

The `id_token_hint` JWT should never appear in a browser-facing URL. It should travel directly from the registry to the auth server over the internal network, and the browser should be redirected straight to the IdP logout URL (e.g. `https://login.microsoftonline.com/...`), bypassing any WAF entirely for that hop.

## Proposed fix

Replace the browser redirect with a server-to-server call from the registry to the auth server using `settings.auth_server_url` (the internal cluster URL, e.g. `http://auth-server.<namespace>.svc.cluster.local:8888`). The registry reads the `Location` header from auth-server's `302` response and redirects the browser directly to the IdP:

```python
# registry/auth/routes.py
if provider:
    redirect_uri = _build_external_url(request, "/logout")
    logout_params: dict[str, str] = {"redirect_uri": redirect_uri}
    if id_token and _validate_jwt_format(id_token):
        logout_params["id_token_hint"] = id_token

    # Server-to-server call — never crosses the public WAF
    s2s_url = f"{settings.auth_server_url}/oauth2/logout/{provider}"
    trusted_host = _resolve_trusted_host(request)
    x_forwarded_proto = "https" if redirect_uri.startswith("https") else "http"
    idp_redirect_url: str | None = None
    try:
        async with httpx.AsyncClient(follow_redirects=False) as client:
            s2s_resp = await client.get(
                s2s_url,
                params=logout_params,
                headers={
                    "X-Forwarded-Host": trusted_host,
                    "X-Forwarded-Proto": x_forwarded_proto,
                },
                timeout=10.0,
            )
        if s2s_resp.status_code in (301, 302, 303, 307, 308):
            idp_redirect_url = s2s_resp.headers.get("location")
    except Exception as exc:
        logger.warning("S2S auth-server logout failed: %s; falling back to local clear", exc)

    if idp_redirect_url:
        response = RedirectResponse(url=idp_redirect_url, status_code=status.HTTP_303_SEE_OTHER)
    else:
        response = RedirectResponse(url="/login", status_code=status.HTTP_303_SEE_OTHER)
    response.delete_cookie(
        settings.session_cookie_name, path="/", domain=settings.session_cookie_domain
    )
```

`X-Forwarded-Host` and `X-Forwarded-Proto` are forwarded so auth-server's `redirect_uri` same-origin validation (`_is_redirect_within_cookie_domain`) sees the real public hostname and approves the URI. The fallback to `/login` ensures logout always completes locally even if the S2S call fails.

## Notes

- The `id_token_hint` only appears in the final IdP URL (e.g. `https://login.microsoftonline.com/...?id_token_hint=...`), which goes directly from the browser to the identity provider — never through the
  application's WAF.
- This issue was discovered on a deployment behind AWS WAF with managed rule sets. The WAF blocked the JWT in the query string, causing a 403 before the auth server was reached.
- When this issue and the [AADSTS90015 query-string-too-long bug](https://github.com/agentic-community/mcp-gateway-registry/issues/1502) are both present, this one manifests first: the WAF blocks the request before Entra is ever reached. Fix this issue first to expose the AADSTS90015 failure.
- The fix also prevents the JWT from appearing in browser history, access logs, and proxy logs.

omrishiv:
The logout flow redirected the browser to the public auth-server URL with the OIDC id_token_hint JWT embedded in the query string. WAFs and reverse proxies routinely block long/JWT-looking query strings, returning 403 before the auth-server was reached -- logout then failed silently after the local session had already been cleared. The JWT also leaked into browser history and access/proxy logs while it was being actively revoked.

The registry now calls auth-server's /oauth2/logout/ directly over the internal network (settings.auth_server_url, the same URL already used by get_oauth2_providers) and redirects the browser straight to the IdP logout URL returned in the Location header. The JWT only ever appears on the direct browser->IdP hop, bypassing the deployment's public infrastructure. This is deployment-agnostic: auth_server_url is configured and reachable in every mode (EKS, Docker/Compose, ECS, EC2, local).

- routes.py: server-to-server logout with X-Forwarded-Host/Proto forwarded so auth-server's redirect_uri same-origin validation approves the URI; _is_safe_logout_redirect guards the returned Location; fail-closed to /login on any S2S failure, non-3xx, or unsafe Location; 5s timeout. Logout URL length remains owned by auth-server (#1502/#1508); no duplicate guard here.
- auth_server: _is_redirect_within_cookie_domain strips the port from a forwarded Host before the same-origin comparison, so a forwarded "localhost:7860" matches the port-stripped redirect_uri hostname.
- nginx: /oauth2/logout/ is no longer browser-facing; return 404 to close the path that would otherwise let a crafted request reintroduce the JWT into a browser-facing URL and the access/WAF logs.
- docs: session-flow logout section rewritten for the S2S flow.
- tests: S2S logout flow (all failure modes) and the host-with-port same-origin validation.

*Issue #, if available:*

Fixes #1503 

*Description of changes:*

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -953,7 +953,15 @@ def _is_redirect_within_cookie_domain(
         forwarded_proto = request.headers.get("x-forwarded-proto", "").split(",")[0].strip()
         forwarded_host = request.headers.get("x-forwarded-host", "").split(",")[0].strip()
         request_scheme = forwarded_proto or request.url.scheme
-        request_host = (forwarded_host or request.url.hostname or "").lower()
+        # Compare hostnames only, never host:port. urlparse(url).hostname above
+        # already strips the port, and request.url.hostname is port-less too, but
+        # a raw X-Forwarded-Host may carry a port (e.g. "localhost:7860" from the
+        # registry's server-to-server logout hop). Normalize the forwarded value
+        # the same way so a same-origin redirect isn't falsely rejected.
+        if forwarded_host:
+            request_host = (urlparse(f"//{forwarded_host}").hostname or "").lower()
+        else:
+            request_host = (request.url.hostname or "").lower()
         if request_host and parsed.scheme == request_scheme and hostname == request_host:
             return True
 
--- a/docker/nginx_rev_proxy_http_and_https.conf
+++ b/docker/nginx_rev_proxy_http_and_https.conf
@@ -375,15 +375,14 @@ server {
         proxy_pass_request_headers on;
     }
 
-    # OAuth2 Cognito logout endpoint
+    # OAuth2 logout is no longer browser-facing (issue #1503). The registry
+    # calls auth-server's /oauth2/logout/ directly over the internal network
+    # (AUTH_SERVER_URL) and redirects the browser straight to the IdP, so the
+    # id_token_hint JWT never traverses this public hop. Return 404 to close
+    # the path that would otherwise let a crafted request reintroduce the JWT
+    # into a browser-facing URL and any proxy/WAF/access logs.
     location /oauth2/logout/ {
-    proxy_pass http://{{AUTH_SERVER_HOST}}:{{AUTH_SERVER_PORT}}/oauth2/logout/;
-    proxy_http_version 1.1;
-    proxy_set_header Host $host;
-    proxy_set_header X-Real-IP $remote_addr;
-    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-    proxy_set_header X-Forwarded-Proto $real_scheme;
-    proxy_pass_request_headers on;
+        return 404;
     }
 
     # OAuth2 Entra ID callback endpoint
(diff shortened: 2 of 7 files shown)


aarora79:
Heads-up: this change introduced a logout regression on **Cognito behind CloudFront**, fixed in #1545.

Two things the S2S internalization surfaced:
1. On the internal registry→auth-server hop, auth-server reconstructs the request origin as `http`/loopback and the forwarded headers don't survive, so the same-origin check dropped a legitimate `https` public `redirect_uri` → Cognito got no valid `logout_uri` (`error=Required+parameters+missing`). #1545 makes the same-origin check also trust the deployment's configured external host.
2. The `/logout` post-logout redirect must be a registered sign-out URL. Keycloak accepts it via its wildcard (`post.logout.redirect.uris='+'`), but Cognito requires the exact URL — `docs/idp/cognito.md` was also corrected (it said `/login`).

Keycloak and Cognito logout both re-verified end-to-end after #1545.

PR #1515, opened 2026-07-19
fix(compose): pass SSRF_ALLOWED_HOSTS/CIDRS to registry in prebuilt and podman (#1513)

### LINKED ISSUE #1513: SSRF_ALLOWED_HOSTS / SSRF_ALLOWED_CIDRS aren't passed to the registry container in docker-compose
Ran into this while wiring an internal MCP server behind the gateway — seems like a
small wiring gap, wanted to flag it before assuming it's a bug.

`.env` documents `SSRF_ALLOWED_HOSTS` / `SSRF_ALLOWED_CIDRS`, and the registry reads
them (`registry/core/config.py` → `settings.ssrf_allowed_hosts` / `_cidrs`, consumed by
`registry/utils/url_guard.py`). But none of the `docker-compose*.yml` files pass those
two vars into the `registry` service's `environment:` block. So on a compose deploy,
setting them in `.env` has no effect — the value never reaches the container.

Practical effect: you can't allowlist an internal upstream. Register a server that's only
reachable by a docker-network hostname (e.g. `http://my-mcp:8000/`) and the health check
marks it `unhealthy: url blocked by SSRF guard`. Putting `SSRF_ALLOWED_HOSTS=my-mcp` in
`.env` changes nothing, because the registry process never sees it. The only workaround
today is hand-editing the compose file to add the passthrough.

**Repro**
1. `./build_and_run.sh --prebuilt`
2. Register a server whose `proxy_pass_url` is an internal docker-network hostname
3. Set `SSRF_ALLOWED_HOSTS=<that-host>` in `.env`, restart the registry
4. Health check still reports `url blocked by SSRF guard` — the host is never allowlisted

**Environment**
- Prebuilt images `:latest` (commit `20996adc`)
- Docker on WSL2 / Ubuntu 24.04

**Suggested fix**
Add the two vars to the `registry` service `environment:`, same `${VAR:-}` pattern as the
neighbors around it:

\`\`\`yaml
      - SSRF_ALLOWED_HOSTS=${SSRF_ALLOWED_HOSTS:-}
      - SSRF_ALLOWED_CIDRS=${SSRF_ALLOWED_CIDRS:-}
\`\`\`

Happy to open a PR (one line each, across the compose files that define the registry
service) — just wanted to check there wasn't an intentional reason they're left out.

rahul188:
## Summary

Fixes #1513

`.env.example` documents `SSRF_ALLOWED_HOSTS` / `SSRF_ALLOWED_CIDRS`, and the registry reads them (`registry/core/config.py` → `settings.ssrf_allowed_hosts` / `_cidrs`, consumed by `registry/utils/url_guard.py`). Only `docker-compose.yml` passed those two vars into the `registry` service. On a `--prebuilt` (or podman) deploy, setting them in `.env` had no effect because the value never reached the container, so an internal upstream reachable only by a docker-network hostname stayed `unhealthy: url blocked by SSRF guard` with no way to allowlist it short of hand-editing the compose file.

## Change

Add `SSRF_ALLOWED_HOSTS` / `SSRF_ALLOWED_CIDRS` to the `registry` service `environment:` in `docker-compose.prebuilt.yml` and `docker-compose.podman.yml`, using the same `${VAR:-}` pattern as `docker-compose.yml`. `docker-compose.dhi.yml` defines only the infra services (mongodb/prometheus/grafana/keycloak-db) and no `registry` service, so it needs no change.

## Verification

Both files parse as valid YAML and the two vars are now present in `services.registry.environment`. With the passthrough in place, `SSRF_ALLOWED_HOSTS=<internal-host>` in `.env` reaches the registry container, so the health check can allowlist an internal docker-network upstream on a prebuilt deploy.

### DIFF
--- a/docker-compose.podman.yml
+++ b/docker-compose.podman.yml
@@ -65,6 +65,10 @@ services:
       # Comma-separated host allowlist for the RUM snippet (fail-closed when set). Empty disables the check.
       - RUM_ALLOWED_HOSTS=${RUM_ALLOWED_HOSTS:-}
       - GATEWAY_ADDITIONAL_SERVER_NAMES=${GATEWAY_ADDITIONAL_SERVER_NAMES:-}
+      # SSRF guard allowlist for internal MCP-server / A2A-agent upstreams at
+      # private IPs (e.g. docker-network hostnames); mirrors docker-compose.yml.
+      - SSRF_ALLOWED_HOSTS=${SSRF_ALLOWED_HOSTS:-}
+      - SSRF_ALLOWED_CIDRS=${SSRF_ALLOWED_CIDRS:-}
       # Registry Card Configuration
       - REGISTRY_URL=${REGISTRY_URL:-http://localhost}
       # Request-source trust (audit client IP + OAuth external-URL host allowlist)
--- a/docker-compose.prebuilt.yml
+++ b/docker-compose.prebuilt.yml
@@ -71,6 +71,10 @@ services:
       # Comma-separated host allowlist for the RUM snippet (fail-closed when set). Empty disables the check.
       - RUM_ALLOWED_HOSTS=${RUM_ALLOWED_HOSTS:-}
       - GATEWAY_ADDITIONAL_SERVER_NAMES=${GATEWAY_ADDITIONAL_SERVER_NAMES:-}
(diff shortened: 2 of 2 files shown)


omrishiv (approved):
Thank you for adding these. LGTM

PR #1519, opened 2026-07-20
feat(egress): client-side Dynamic Client Registration for Atlassian Rovo authv2

go-faustino:
## What this does

Adds RFC 7591 **client-side Dynamic Client Registration** to the egress OAuth engine, and wires the `atlassian` provider recipe to Atlassian's Rovo MCP `authv2` Authorization Server.

From 2026-05-27 Atlassian Identity only issues MCP-audience tokens to DCR-registered clients. A static classic-3LO app still authenticates, but its token carries the wrong audience and is rejected at the Rovo MCP tool layer. So the gateway has to register its own OAuth client rather than use operator-supplied credentials.

## Verified end to end against live Atlassian

DCR-registered public client → authorize → consent → code → token → MCP session → **real data-tool calls**:

- token exchange succeeds as a **public** client (PKCE only, no `client_secret`), `HTTP 200`, refresh token present
- `aud` binds to the DCR `client_id` **plus the user's site/workspace ARIs** — not to `api.atlassian.com`, and no `resource`/`audience` parameter participates
- `MCP initialize` → `200`, `tools/list` → `200`
- `getVisibleJiraProjects`, `getJiraIssue`, `atlassianUserInfo`, `getAccessibleAtlassianResources` all `200` / `isError: false`, with the site reporting granular `read:jira-work`/`write:jira-work` scopes

## The blocker, for the record

This PR sat conflicting for six weeks because the authorize request kept failing with `invalid_request` / "Incorrect request parameters". The cause turned out to be a **single missing scope: `read:account`**.

Rovo authv2 rejects any authorize request that omits it — and rejects it *after* the user submits consent, naming no scope. Nothing advertises the requirement: RFC 9728 has no required-scopes field, so `read:account` looks exactly as optional as the other 21 entries in `scopes_supported`. Isolated by a controlled sweep, same client and redirect, one variable at a time:

| Requested scopes | Consent screen | Result |
|---|---|---|
| minimal Jira set, no `read:account` | left as-is | fail |
| minimal Jira set, no `read:account` | Search group unticked | fail |
| minimal Jira set **+ `read:account`** | Search group unticked | success |
| minimal Jira set **+ `read:account`** | left as-is | success |
| full 13-scope set (includes `read:account`) | left as-is | success |

`read:account` fully determines the outcome. The consent screen's Read/Write/Search permission groups have no effect in either direction.

Ruled out with evidence, not inference: DCR client propagation/churn, `redirect_uris` not being persisted (the DCR `201` does echo them), `prompt=consent`, RFC 8707 `resource`, classic `audience`, `state` length, org entitlement, callback-domain allowlisting, and site/ARI resolution.

## Scope of the change

Rebased onto current `main`. That rebase showed **most of the original PR had become redundant**: upstream now has `TokenEndpointAuthStyle.NONE`, whose docstring already anticipates *"a client minted by an MCP resource server's Dynamic Client Registration endpoint"*, and `service._client_secret` already returns `None` for that style while fail-closing for confidential ones.

So the original public-client apparatus — the `public_client` field, the `_build_token_request` change, and `_token_leg_secret` — is **deleted** and rebuilt on upstream's `NONE` style. This PR no longer touches `_build_token_request`, and the fail-closed guard plus its two tests are byte-identical to `main`.

What remains is only what upstream lacks:

- `requires_dcr`, `registration_url`, `protected_resource_metadata_url`, `dcr_client_name`, `default_scopes`, `required_scopes` on `OAuthProviderConfig`
- `register_dcr_client()` with registration-endpoint discovery walking RFC 9728 → RFC 8414 (the Rovo AS is path-scoped and its `registration_endpoint` is tenant-scoped, so pinning is not viable)
- discovery and registration through the same `CREDENTIALED_OAUTH_PROFILE` guarded client as the token endpoint
- one-time DCR at config time, persisting and reusing the `client_id` so a re-save does not churn a registration; `502` on failure
- a clear `400` instead of a `KeyError`/`500` at both consent entry points when a `requires_dcr` provider has no `client_id` yet
- `required_scopes`, unioned into every request regardless of operator config — deliberately not just a default, since a default only applies when the operator supplies nothing and an explicit list omitting `read:account` would fail identically
- config-time scope validation against the resource's `scopes_supported`, reusing the PRM the discovery walk already fetches (no extra round-trip) and running *before* registration so a bad config cannot orphan a DCR client at the AS
- the `atlassian` recipe: `requires_dcr` + `NONE` style + PRM discovery, no `audience`, no `resource`

A NONE-style client does not persist the `client_secret` Atlassian returns anyway (it arrives with `client_secret_expires_at: 0` even when `token_endpoint_auth_method` is echoed as `"none"`): the token leg never reads it, so storing it is needless exposure.

## Notes

- `read:jira-user` and the classic `read:confluence-content.all` / `write:confluence-content` are absent from the authv2 resource's `scopes_supported`.
- Confluence scopes are not defaulted. The 5-scope Jira default covers 16 of the 31 tools and narrows `aud` to Jira ARIs; operators wanting Confluence opt in per server.
- The Rovo AS also advertises `pushed_authorization_request_endpoint` and `client_id_metadata_document_supported: true`, so PAR and CIMD are available for later — relevant to #1687.

## Tests

`334 passed` in `tests/unit/egress_auth/`. `ruff check` and `ruff format --check` clean.


aarora79:
Hi @go-faustino — we pulled this PR and tested the full DCR + public-PKCE flow end-to-end against Atlassian's live Rovo authv2 AS. It mostly works; we made one fix on top and hit a wall on the Atlassian side we couldn't get past. Sharing everything so you can take the last mile (you presumably have a working authv2 org).

**What works (verified live):**
1. RFC 9728 → 8414 discovery of the registration endpoint.
2. RFC 7591 DCR registration (`201 Created`, fresh client per config-save).
3. The authorize URL the code builds is RFC-complete and matches Atlassian's own AS metadata (`response_type=code`, `client_id`, `redirect_uri`, `state`, space-joined `scope`, `code_challenge`+`S256`, `prompt=consent`, endpoint `https://auth.atlassian.com/authorize`).
4. Reached the **real consent screen + site picker** with a valid client and valid scopes.

**One fix we'd suggest — Atlassian authv2 DCR returns a _confidential_ client.**
We observed that the DCR endpoint hands back a `client_secret` **even when we register with `token_endpoint_auth_method=none`**. The current code forces `client_secret=""` whenever `cfg.public_client` on the token/refresh legs, which drops the secret Atlassian issued and expects → the token exchange would fail. Our change decides the token-leg auth by **whether a secret is actually held**, not by the declared `public_client` intent (262 egress_auth tests pass). Open question: is the intended model (a) register `none` but send the returned secret if Atlassian gives one — what we did — or (b) register a confidential client for Atlassian (`client_secret_post`) and drop `public_client=True`?

**Config gotchas we hit (not code):**
- `read:jira-user` is **not** in the Rovo authv2 resource's `scopes_supported`; Atlassian rejected consent ("requested scopes that have not been added to the app: read:jira-user") until we removed it. Classic 3LO had it; authv2 doesn't. Might be worth validating requested scopes against the resource metadata at config time.
- Our org blocked the callback domain until an admin allowlisted it.

**The wall — need your input.** After the domain was allowlisted and scopes were valid, the authorize leg still redirects to:
```
https://id.atlassian.com/error?error=invalid_request&error_description=Incorrect%20request%20parameters
```
…for a freshly-minted DCR client — even though an earlier, identical-shape request (different client) had reached the consent screen minutes before. No callback ever reached the gateway; the failure is entirely Atlassian-side of the authorize step. We suspect DCR-client churn/propagation (we created ~6 clients while debugging), an org entitlement nuance, or an authorize detail authv2 needs beyond the standard set.

Questions:
1. Did you get an end-to-end token mint working on your org? What did the final working authorize request look like (anything beyond the params above)?
2. Does authv2 need anything special after DCR — propagation delay, a specific `prompt`, `audience`/`resource`, a particular `token_endpoint_auth_method`, or org-side app approval?

**Minor robustness bug (unfixed, your call):** `GET /oauth2/egress/connect` → `build_consent_url` does `egress_oauth["client_id"]` unguarded (`egress_oauth_facade_routes.py:178` → `service.py:287`); a `requires_dcr` provider with no `client_id` yet raises `KeyError` → HTTP 500. A clean 4xx ("re-save egress config to register") would be friendlier.

<details>
<summary>Full patch (our commit on top of this PR head — <code>git am</code>-able)</summary>

````diff
diff --git a/registry/egress_auth/oauth_engine.py b/registry/egress_auth/oauth_engine.py
index ef413054..94ace6ae 100644
--- a/registry/egress_auth/oauth_engine.py
+++ b/registry/egress_auth/oauth_engine.py
@@ -166,11 +166,14 @@ def _build_token_request(
     """Return (form_data, headers), placing the client secret per the provider's style."""
     headers = {"Accept": "application/json"}
     data = dict(form)
-    if cfg.public_client:
-        # RFC 6749 public client (token_endpoint_auth_method=none): no secret is
-        # sent -- PKCE is the proof-of-possession. DCR providers like Atlassian
-        # authv2 hand back a secret-less client, and their token endpoint rejects
-        # a client_secret. Only the client_id identifies the client.
+    if not client_secret:
+        # No secret to present, so authenticate as an RFC 6749 public client
+        # (token_endpoint_auth_method=none) proven by PKCE; only the client_id
+        # identifies the client. We key off the ABSENCE of a secret rather than
+        # cfg.public_client because some DCR providers (Atlassian authv2) return a
+        # confidential client -- with a client_secret -- even when we register with
+        # token_endpoint_auth_method=none. In that case a secret IS stored and the
+        # branches below send it, matching what the provider actually expects.
         data["client_id"] = client_id
         return data, headers
     if cfg.token_endpoint_auth_style == TokenEndpointAuthStyle.BASIC_HEADER:
diff --git a/registry/egress_auth/service.py b/registry/egress_auth/service.py
index b21b7a65..2c7438ed 100644
--- a/registry/egress_auth/service.py
+++ b/registry/egress_auth/service.py
@@ -24,6 +24,7 @@ from registry.egress_auth import oauth_engine
 from registry.egress_auth.providers import resolve_provider
 from registry.egress_auth.schemas import (
     EgressConnection,
+    OAuthProviderConfig,
     OAuthState,
     StoredToken,
 )
@@ -214,6 +215,24 @@ class EgressAuthService:
             raise EgressAuthError("could not decrypt egress client_secret (SECRET_KEY changed?)")
         return secret
 
+    def _token_leg_secret(
+        self,
+        cfg: OAuthProviderConfig,
+        egress_oauth: dict,
+    ) -> str:
+        """Return the client_secret to present on the token/refresh legs.
+
+        A public client sends no secret (PKCE is the proof-of-possession), so we
+        return "" only when none is stored. We key off the STORED secret rather
+        than cfg.public_client alone because some DCR providers (Atlassian authv2)
+        return a confidential client -- with a secret -- even when we register with
+        token_endpoint_auth_method=none. If a secret was persisted we must send it,
+        or the provider rejects the exchange; if none was stored we stay public.
+        """
+        if not egress_oauth.get("client_secret_encrypted"):
+            return ""
+        return self._client_secret(egress_oauth)
+
     def _is_near_expiry(self, token: StoredToken) -> bool:
         if not token.expires_at:
             return False  # no expiry info -> treat as long-lived; refresh on 401 elsewhere
@@ -316,8 +335,9 @@ class EgressAuthService:
             raise EgressAuthError("state auth_method mismatch")
 
         cfg = resolve_provider(egress_oauth)
-        # Public clients (DCR authv2) have no secret; PKCE is the proof-of-possession.
-        client_secret = "" if cfg.public_client else self._client_secret(egress_oauth)
+        # Send the stored secret if we have one; a genuine public client (no secret
+        # persisted) sends none and relies on PKCE. See _token_leg_secret.
+        client_secret = self._token_leg_secret(cfg, egress_oauth)
         token = await oauth_engine.exchange_code(
             cfg=cfg,
             client_id=egress_oauth["client_id"],
@@ -410,7 +430,7 @@ class EgressAuthService:
             if current is None or not current.refresh_token:
                 return None
             cfg = resolve_provider(egress_oauth)
-            client_secret = "" if cfg.public_client else self._client_secret(egress_oauth)
+            client_secret = self._token_leg_secret(cfg, egress_oauth)
             try:
                 new = await oauth_engine.refresh_token(
                     cfg=cfg,
diff --git a/tests/unit/egress_auth/test_oauth_engine.py b/tests/unit/egress_auth/test_oauth_engine.py
index 6c204d9f..23212222 100644
--- a/tests/unit/egress_auth/test_oauth_engine.py
+++ b/tests/unit/egress_auth/test_oauth_engine.py
@@ -259,10 +259,12 @@ class TestQuirkParsers:
         assert "Authorization" not in headers
         assert headers["Accept"] == "application/json"
 
-    def test_public_client_omits_secret(self):
-        # RFC 6749 public client (token_endpoint_auth_method=none): only client_id
-        # is sent -- no client_secret in the body and no Basic header. Atlassian
-        # authv2's token endpoint rejects a secret on a public DCR client.
+    def test_no_secret_authenticates_as_public_client(self):
+        # With no secret to present, authenticate as an RFC 6749 public client
+        # (token_endpoint_auth_method=none): only client_id is sent -- no
+        # client_secret in the body and no Basic header. This is the genuine
+        # public-client case (a secret-less DCR client), keyed off the ABSENCE of
+        # a secret rather than cfg.public_client.
         cfg = OAuthProviderConfig(
             name="atlassian",
             display_name="Atlassian",
@@ -275,17 +277,21 @@ class TestQuirkParsers:
         assert "client_secret" not in data
         assert "Authorization" not in headers
 
-    def test_public_client_ignores_supplied_secret(self):
-        # Even if a secret is somehow present, a public client never sends it.
+    def test_public_client_with_secret_still_sends_it(self):
+        # Some DCR providers (Atlassian authv2) return a CONFIDENTIAL client -- with
+        # a client_secret -- even when we register token_endpoint_auth_method=none.
+        # When a secret is actually held it MUST be sent, or the token exchange is
+        # rejected. The secret's presence, not cfg.public_client, drives the choice.
         cfg = OAuthProviderConfig(
-            name="p",
-            display_name="P",
-            authorize_url="https://i/a",
-            token_url="https://i/t",
+            name="atlassian",
+            display_name="Atlassian",
+            authorize_url="https://auth.atlassian.com/authorize",
+            token_url="https://auth.atlassian.com/oauth/token",
             public_client=True,
         )
-        data, _ = oauth_engine._build_token_request(cfg, "cid", "leaked", {"grant_type": "x"})
-        assert "client_secret" not in data
+        data, _ = oauth_engine._build_token_request(cfg, "cid", "real-secret", {"grant_type": "x"})
+        assert data["client_id"] == "cid"
+        assert data["client_secret"] == "real-secret"
 
 
 @pytest.mark.unit
````
</details>

go-faustino:
Thanks @aarora79 — this is an incredibly useful review, especially validating the discovery/DCR/authorize legs against the live AS. I've pushed `af8375df` adopting your fixes and answering your questions below.

## Adopted from your patch

**Token-leg auth keyed off secret presence (your option (a)).** You're right — Atlassian's authv2 DCR hands back a `client_secret` even when we register `token_endpoint_auth_method=none`, and my `if cfg.public_client:` short-circuit dropped it. Pushed your exact approach:
- `oauth_engine._build_token_request` now branches on `if not client_secret` (PKCE is always sent, so it's correct whether the DCR client comes back public *or* confidential).
- Added `service._token_leg_secret(cfg, egress_oauth)` (send the stored secret if one was persisted, else stay public) and wired it into both `exchange_code` and `refresh_token`.
- Tests updated (`test_no_secret_authenticates_as_public_client`, `test_public_client_with_secret_still_sends_it`, plus `_token_leg_secret` unit tests).

I kept `public_client=True` on the `atlassian` recipe because it still drives the two things that *are* intent (not observation): the DCR registration request (`token_endpoint_auth_method=none`) and relaxing the operator-`client_secret` requirement at config time. The token-leg behaviour is now purely secret-presence-driven, so option (b) isn't needed.

**Robustness bug (the unguarded `client_id` → KeyError/500).** Fixed. `build_consent_url` now raises a clear `EgressAuthError` ("re-save the server's egress config to trigger dynamic client registration") when a `requires_dcr` provider has no `client_id` yet, and both consent entry points (`/api/egress-auth/initiate` and `/oauth2/egress/connect`) translate it to a clean 400 instead of a 500. Tests added on both routes.

## Scope gotcha — confirmed, independently

Same finding on our side: `read:jira-user` (and the classic `read:confluence-content.all` / `write:confluence-content`) are **not** in the authv2 resource's `scopes_supported`; the AS rejects consent until they're removed. We run the minimal Jira set: `read:me`, `offline_access`, `read:jira-work`, `write:jira-work`. Agree that validating requested scopes against the PRM `scopes_supported` at config time is worth doing — I'd rather land it as a focused follow-up than add another config-time network call to this PR, but happy to include it here if you'd prefer.

## Your questions

**1. Did I get an end-to-end token mint on my org, and what did the final working authorize look like?**

Straight answer: I have **not** reproduced a clean *interactive* mint through the gateway's own callback either — I hit the identical `error=invalid_request&error_description=Incorrect%20request%20parameters` at `/authorize`, and in some attempts a `Redirect URL is missing` variant, for freshly-minted DCR clients. My conclusions about the token shape (`aud = [DCR client_id, <site ARIs>]`, no `resource`/`audience` binding) came from decoding vended tokens + the resource/AS metadata, and from a standards-compliant reference client — **not** from a clean gateway-callback mint. So I can't hand you a "magic" authorize request beyond the RFC-complete set the code builds (which you already confirmed matches the AS metadata). We're standing at the same wall.

**2. What does authv2 need after DCR?** Ranked by how well it fits our shared symptom (identical-shape request reaches consent for one client, fails for the next):

- **DCR client propagation / churn (my top suspect).** "Worked minutes earlier, a fresh client fails" is the classic signature of a just-registered client not yet usable at `/authorize`. We both created several clients while debugging. Suggestion to isolate: register **once**, stop churning, wait ~30–60s, then authorize; and don't re-register on every config save (this PR already reuses a persisted/operator `client_id` for exactly this reason).
- **`redirect_uri` exact-match.** `register_dcr_client` sends `redirect_uris: [<gateway callback>]` and the authorize leg uses the same value, so the code is consistent. But `Redirect URL is missing` specifically means the AS doesn't recognise the redirect *for that client* — which is either propagation or a client that didn't actually persist `redirect_uris`. **Does the DCR `201` response you saw echo back `redirect_uris`?** If it doesn't, that's the smoking gun. (Also worth isolating: mcp-remote uses a loopback `127.0.0.1` redirect; the gateway uses an `https` callback — if a loopback client authorizes but an identical https one doesn't, that narrows it to redirect handling rather than propagation.)
- **Org-side app authorization / allowlist.** Your callback-domain allowlist matches what we saw; some orgs additionally gate the *app* itself. For a per-config DCR client this may need re-approval each time a new client appears — another reason to register once and reuse.
- **No `resource` / `audience`.** Confirmed rejected/ignored on both legs; the recipe keeps them off. Not the cause of the authorize error.

## Net

With your token-leg fix + valid scopes + the domain allowlist, the only thing between this and a working vend is that intermittent authorize-side `invalid_request` — and your own data (one client reached consent) says that's an Atlassian propagation/churn/entitlement issue, not a code gap. If you can confirm whether your DCR `201` echoed `redirect_uris`, and whether a single non-churned client succeeds after a short delay, I think we can close it out. Thanks again for taking it this far.

aarora79:
Hi @go-faustino, will test this next week, so with this do you have Atlassian-Rovo server working?

go-faustino:
Hi @aarora79 — thanks, appreciate you taking another pass. I'll be away until September so won't be able to respond to findings in real time, but to set expectations honestly: as of my last comment, no, I haven't gotten a clean end-to-end token mint through the gateway's own callback. The DCR/PKCE/discovery/authorize-URL shape is validated correct (yours and mine both reach consent), but we're both hitting the same invalid_request wall at /authorize for freshly-minted DCR clients — I think it's Atlassian-side (propagation/churn or entitlement), not something in this code, but I couldn't confirm.

If you get a chance to test: the two things that'd actually move this forward are (1) whether your DCR 201 echoes back redirect_uris, and (2) whether a single non-churned client succeeds after a short delay (~30-60s) rather than registering repeatedly while debugging. If you get a clean mint, that'd basically close this out.

I'll pick this back up in September — feel free to push commits or just leave findings on the thread in the meantime.

zoltan-fedor:
+1

adrianB1996:
+1

go-faustino:
Hi @aarora79 — back as promised. End-to-end mint works. Root cause is one missing scope: **`read:account`**.

Rovo authv2 rejects any authorize request that omits it, **after** the user submits consent, with `invalid_request` / "Incorrect request parameters" and no indication of which scope is at fault. RFC 9728 has no required-scopes field, so `read:account` looks as optional as the other 21 entries in `scopes_supported`. That is why neither of us caught it: every requested scope was individually valid.

Same DCR client, same redirect, one variable: without `read:account` the request fails (consent groups ticked or not); with it, it succeeds (consent groups ticked or not). `mcp-remote` always worked because it requests the full set, which includes `read:account`.

Your question 1: the DCR `201` **does** echo `redirect_uris`, and the consent screen shows the registered callback under "Domains". Also ruled out: propagation/churn, `prompt=consent`, `resource`/`audience`, `state` length, org entitlement.

Token exchange succeeds as a **public** client — PKCE only, no `client_secret`, HTTP 200, refresh token present. `aud` binds to the DCR `client_id` plus the user's site/workspace ARIs. Real data tools (`getVisibleJiraProjects`, `getJiraIssue`) return 200 with granular `read:jira-work`/`write:jira-work`, not `:agent-interface`.

That reverses the token-leg fix you contributed. Atlassian's DCR *issues* a secret even when `token_endpoint_auth_method` is `"none"`, but the token endpoint does not require it. Your patch keyed auth off "is a secret held" — the right call on the evidence then, untestable because neither of us had a code. I have dropped it.

Rebased onto current `main`. `TokenEndpointAuthStyle.NONE` now exists upstream and already covers this case, so the original `public_client` / `_build_token_request` / `_token_leg_secret` apparatus is gone. **This PR no longer touches `_build_token_request`**; your fail-closed guard and its two tests are byte-identical to `main`.

What remains is DCR discovery + registration, the `atlassian` authv2 recipe (`requires_dcr` + `NONE`, no `audience`/`resource`), `required_scopes=["read:account"]` (mandated, not merely defaulted — a default would not catch an operator-supplied list that omits it), a 5-scope Jira `default_scopes`, and config-time validation against `scopes_supported` (the check you asked for). NONE-style clients do not persist the secret Atlassian hands back.

Happy to split the scope work into a follow-up if you would rather land DCR on its own. Thanks for the live testing that narrowed this to the post-consent step.

go-faustino:
Hi @aarora79 gentle bump on this one — no rush if you're heads-down elsewhere.

Since your Aug 2 question, the answer is now yes: end-to-end token mint through the
gateway callback works against live Rovo authv2. Root cause was the missing
`read:account` scope (details in my Sep 2 comment).

Branch is rebased on current main and green.

Thanks!

aarora79:
thanks for the nudge @go-faustino , I have been following your commits and this has been on my list of things to get to, definitely this week and thank you for your work on this.

go-faustino:
Rebased onto current `main` (`13badb74`) — conflicts resolved.

Both conflicts were contextual, so nothing this PR does has changed: the diff is still 1394 insertions across the same 12 files.

- `registry/api/egress_auth_routes.py` — import block only. Upstream's obo work replaced the `_is_gateway_own_audience` call site with `_validate_obo_egress_config`, so I kept upstream's helper alongside this PR's `oauth_engine` import. That call site came from the old base, not from this PR.
- `tests/unit/egress_auth/test_configure_egress_url_validation.py` — upstream's `TestConfigureOboAudience` and this PR's `TestConfigureEgressDcr` were appended at the same place. Both kept, purely additive.

Verification: `tests/unit/egress_auth/` is `383 passed`. A full `tests/unit/` run matches clean `main` exactly — the same failures before and after the rebase, none of them in `egress_auth`. `ruff check` and `ruff format --check` are clean on every file this PR touches.

go-faustino:
Rebased onto current main (b8f3e395) and conflicts resolved; head is now fe377b90.

The one non-trivial conflict came from the backend-auth discovery identity work (7275a29a), which moved provider-config building out of `configure_egress_auth` into a shared `build_oauth_provider_config`, now also called from `server_routes.py`. This PR's additions were inline in the old block, so they're split across the new structure:

- **Scope handling** (`default_scopes`, then union with `required_scopes`) now lives in `build_oauth_provider_config`, so every caller gets the same resolved scope list.
- **DCR** stays specific to per-user egress. It's the same code, moved unchanged into `_register_or_reuse_dcr_client`, which `configure_egress_auth` calls after the builder when the provider has `requires_dcr`.
- **The builder refuses a `requires_dcr` provider with a 400** unless the caller passes `dcr_supported=True`; only `configure_egress_auth` does. Discovery can't register a client, so Atlassian there now gets an explicit "requires dynamic client registration" error instead of persisting a static-app config that Rovo authv2 would reject at the tool layer. Two new tests cover the builder on its own: refused without DCR support, and returned with required scopes applied when the caller will complete it.

The other two conflicts were mechanical. In `oauth_engine.py`, upstream's `client_credentials_token` and this PR's DCR functions were added at the same spot; both are kept. In `egress_oauth_facade_routes.py`, upstream's new `purpose` argument and `oauth_cfg` are kept inside this PR's `EgressAuthError` handler.

Verification: `tests/unit/egress_auth` plus the discovery suites (`test_oauth_backend_routes_authz.py`, `test_backend_oauth.py`, `test_server_delete_revokes_discovery.py`) pass, 510 passed; `tests/unit/api/` has 1218 passed and 8 skipped. ruff 0.11.13 and mypy 1.15.0 (the CI pins) are clean on all changed files.

PR #1520, opened 2026-07-20
Secrets Manager overflow sharding and concurrency hardening

omrishiv:
### Problem

The Secrets Manager egress store packs all of a principal's OAuth connections into a single JSON secret. This worked for typical users with a few connections, but had three robustness gaps:

1. **No size limit enforcement.** A principal with enough connections—or unusually large tokens/scopes—would hit a hard write failure with no graceful handling or advance detection.

2. **Weak concurrency guarantees.** The previous read-merge-write-verify loop could still lose updates: two writers updating different connections for the same principal could each verify their own version before the other's later write overwrote it.

3. **No distributed mutation serialization.** Multi-replica deployments could race on the same principal's shared JSON document without coordination..

### Design

**Transparent overflow via hash shards, coordinated by a principal-level Mongo lease.**

The store preserves the existing bare-map JSON format for principals that fit within a configurable byte target (default 60 KiB). When a mutation would exceed the target, the store automatically promotes to a sharded layout:

```
Root secret (manifest):
  { "_egress": { "schema": 1, "layout": "sharded", "generation": "<uuid>",
                 "bucket_count": N, "hash": "sha256-v1" } }

Shard secrets:
  {root}/overflow/{generation}/0
  {root}/overflow/{generation}/1
  ...
```

Entries are routed to buckets via `SHA-256(map_key) mod bucket_count`. Bucket count doubles until every shard fits under the target, bounded by a configurable max (default 64).

**Commit protocol (manifest-last):**
1. Acquire principal-level mutation lease (Mongo TTL-backed, existing `mcp_egress_operational` collection).
2. Read root → load all entries (inline or across shards).
3. Apply mutation in memory.
4. If result fits inline → write root as bare map, verify, clean up old shards.
5. If overflow needed → write all new-generation shards, verify each, then commit new manifest to root last.
6. Release lease; async-clean old generation.

Readers never hold a lease—they read the root, then fetch only the target shard (for `get_token`) or all shards in bounded parallel (for `list_for_user`). A crash before manifest commit leaves the old layout authoritative. A crash after leaves the new generation readable.

**Compaction:** When a delete brings the entry set back under the target, the store compacts to inline and removes all shards.

### What changed

| File | Change |
|------|--------|
| `registry/secrets/secrets_manager/store.py` | versioned document model, overflow partitioning, manifest-last commit, bounded parallel shard I/O, payload-size validation, in-process local lock + distributed Mongo lease integration. |
| `registry/secrets/factory.py` | Wire `EgressOperationalRepository` into the Secrets Manager store constructor for cross-replica mutation leasing. |
| `registry/repositories/documentdb/egress_operational_repository.py` | Add `renew_lease()` method; handle `DuplicateKeyError` on contended upsert. No schema migration needed—same collection, same TTL index. |
| `tests/integration/test_aws_secrets_manager_live_integration.py` | New live-AWS integration test: gated by explicit env vars + exact account ID confirmation. Exercises full inline → sharded → read → list → compact → cleanup lifecycle against real Secrets Manager. |

### Backward compatibility

- Existing inline secrets (the previous bare-map format) are read transparently without migration. The store recognizes the absence of `_egress` metadata as legacy/inline.
- The `SecretStoreBase` interface is unchanged—callers (`EgressAuthService`, routes, refresh) require no modifications.
- No new environment variables are required for default behavior. New optional tunables: none exposed yet (internal constants; can be promoted to config if needed).
- Mongo operational collection gains `renew_lease` but no new indexes or document kinds.

### Testing

- **Unit (Moto):** All 24 existing contract tests pass (both backends + lease tests).
- **Local smoke:** Fake-client overflow lifecycle (inline → sharded → compact → empty).
- **Live AWS:** Full integration test passed—created 5 secrets (1 root + 4 shard buckets), verified reads from shards, list across shards, compaction on delete, and complete cleanup with zero leaked resources.
- **Deployed:** Registry patched to `SECRET_STORE_BACKEND=secrets-manager`; rolled out successfully with pod identity credentials.


aarora79:
Pushed a follow-up commit (`add5dee1`) to this branch addressing the one concern from review — the reader-vs-cleanup generation race.

**Problem:** `get_token` / `list_for_user` can read the root at generation G1, then a concurrent same-principal commit flips the manifest to G2 and cleans up G1's shards before the reader fetches them. The shard read used `required=True`, so that 404 surfaced as a hard `SecretStoreError` on a read path with no caller retry (confirmed in `egress_auth/service.py` — it gates token vend for a live proxied request).

**Fix:** treat it as the transient stale read it is. `_read_shard` now reads with `required=False` and raises an internal `_StaleShardError` when a required shard is absent (distinct from a corrupt layout); `get_token`/`list_for_user` run through a bounded `_read_with_retry` (3 attempts) that re-reads the root — which now points at the new generation, or an inline/empty layout — logging at INFO when a race is observed, and raising a clear error only if it keeps racing.

**Tests:** `TestOverflowStaleReadRetry` simulates a concurrent commit that flips the root and deletes the old shards mid-read; both `get_token` and `list_for_user` recover. Verified they fail against the pre-fix `required=True` behavior. Full `tests/unit/secrets/` suite: 56 passed, ruff clean. OpenBao path untouched.

Nothing else changed. Feel free to squash it into your history however you prefer.

PR #1523, opened 2026-07-20
Fix dead @app.on_event(startup) legacy-scope-audit handler

soheilkhan:
## Summary

`FastAPI(..., lifespan=lifespan, ...)` in `auth_server/server.py` means Starlette never invokes `@app.on_event("startup")` handlers registered on the same app — silently, with no warning. This file has two such handlers left over from before the `lifespan` migration:

- A scope-reload handler (`startup_event`) — harmless, since `lifespan()` already reloads scopes earlier in the same startup sequence. Dead but not a functional bug.
- `_run_legacy_scope_audit_on_startup` — a real bug. The legacy-scope audit (which warns about scope/server pairs needing attention before an upgrade) never actually runs on boot, silently.

## Changes

Move `_audit_legacy_scopes_on_startup()`'s call into `lifespan()` (right after the static-token-map build, matching where the removed handler sat relative to it in registration order) and remove both dead `@app.on_event` decorators.

## Testing

`python3 -c "import ast; ast.parse(...)"` clean. Found and fixed the identical pattern in a downstream fork, where it also silently broke a static-token-map builder (that particular handler already lives correctly inside `lifespan()` on this `main`, so only the legacy-scope-audit half of the bug applies here).

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -1762,6 +1762,16 @@ async def lifespan(app: FastAPI):
     # Runs after scopes are loaded so map_groups_to_scopes can resolve groups.
     await _build_static_token_map()
 
+    # Run the legacy-scope audit. This used to be registered via the
+    # deprecated @app.on_event("startup") API, but FastAPI/Starlette never
+    # invokes on_event handlers once a custom `lifespan` is passed to
+    # FastAPI() (as above) -- it was silently dead code, so this audit never
+    # actually ran on boot.
+    try:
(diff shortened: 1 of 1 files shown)


aarora79:
Pushed a small follow-up commit (`abc0ccf9`) to fix the failing `pre-commit` CI job. Removing the trailing `_run_legacy_scope_audit_on_startup` handler left two blank lines at EOF, which tripped the `ruff-format` hook. The commit just trims those; no logic change. CI should be green on re-run.

PR #1526, opened 2026-07-20
Fix virtual tools/list endpoint resolution and partial caching

soheilkhan:
## Summary

- Reuse the centralized endpoint resolver when generating virtual backend routes, so nested transport paths such as `/mcp/http` are not rewritten as `/mcp/http/mcp`.
- Track backend discovery success separately from the returned tool count, including valid empty tool lists.
- Fall back to mapping metadata per failed backend instead of dropping that backend whenever another backend succeeds.
- Cache enriched tools only after every backend succeeds, so fallback responses are retried on the next request rather than cached for 60 seconds.
- Treat truncated or malformed backend responses as discovery failures.

## Problem

A virtual server backed by multiple MCP servers could return only the tools from whichever backend happened to respond successfully. Because any non-empty backend response marked the whole aggregation successful, that partial list was then cached for 60 seconds. A separate endpoint-building path also appended `/mcp` to valid nested endpoints such as `/mcp/http`, causing a 404 and triggering the partial-list behavior.

## Verification

- `pytest tests/unit/test_virtual_server_nginx.py -q --no-cov`: 28 passed.
- Full upstream unit suite: 5,417 passed and 24 skipped; the sole environment failure was rerun successfully with the upstream-locked `hvac` dependency present.
- Ruff passed for the changed Python files (excluding the repository's existing F541 findings).
- The modified router compiled successfully with OpenResty LuaJIT.
- An OpenResty harness with one healthy and one failing backend verified that the first response included per-backend fallback metadata without populating the enriched cache; the next request retried the recovered backend, used its live metadata, and then cached the complete result.

## Compatibility and security

Explicit MCP endpoint paths still use the configured proxy host for internal routing. Existing registration-time URL validation and render-time nginx sanitization remain in place.

### DIFF
--- a/docker/lua/virtual_router.lua
+++ b/docker/lua/virtual_router.lua
@@ -336,8 +336,27 @@ local function _collect_backend_locations(mapping)
 end
 
 
+-- Append mapping-file metadata for one backend when live discovery fails.
+local function _append_mapping_tools_for_backend(enriched_tools, mapping, backend_location)
+    if not mapping.tools then
+        return
+    end
+
+    for _, tool in ipairs(mapping.tools) do
+        if tool.backend_location == backend_location then
+            enriched_tools[#enriched_tools + 1] = {
+                name = tool.name,
+                description = tool.description or "",
+                inputSchema = _ensure_mcp_schema(tool.inputSchema),
+                required_scopes = tool.required_scopes,
+            }
+        end
+    end
+end
+
+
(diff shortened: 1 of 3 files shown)


aarora79:
@soheilkhan please resolve https://github.com/agentic-community/mcp-gateway-registry/actions/runs/29777990651/job/88471971846?pr=1526, we will include this in 1.29.0. Thanks for the contribution.

PR #1527, opened 2026-07-21
Harden the AgentCore sync adapter's egress and secret handling

omrishiv:
# Harden the AgentCore sync adapter's egress and secret handling

## Summary

The AgentCore sync adapter (`cli/agentcore/`) is a standalone CLI/sidecar that
discovers Bedrock AgentCore gateways, registers them with the registry, and
refreshes upstream OAuth2 credentials. This PR adds:

1. **Credentials never travel over cleartext.** The registry hop (which carries
   the registry Bearer JWT and a minted upstream access token) refuses plain
   `http://` to any non-loopback host. The OIDC hops (which carry the OAuth2
   client secret) require HTTPS.

2. **Registrant/manifest-supplied OIDC URLs are SSRF-guarded.** The
   `discovery_url` (from Bedrock gateway metadata) and the `token_endpoint`
   derived from it are validated through the registry's canonical URL guard —
   default-deny for private/loopback/link-local/reserved/cloud-metadata targets,
   with the resolved IP pinned at connect time (rebinding-safe). Validation runs
   at three points: when a registration is built, when the refresh manifest is
   written, and again when the manifest is read at refresh time (a tampered
   manifest cannot drive a credential fetch to an internal or non-HTTPS target).

3. **IdP vendor detection is host-anchored, not substring-based.** Vendor
   classification now matches the parsed hostname on registered-domain
   boundaries, so a look-alike host such as `myorg.auth0.com.attacker.example`
   is no longer classified as the real vendor

4. **The Cognito client-secret lookup validates before calling AWS.** The
   `region`/`pool_id` spliced out of a discovery URL are shape-checked before a
   boto3 client is built.

5. **On-disk secrets are written `0o600` atomically.** The refresh manifest is
   written owner-only via a temp-file + atomic-rename helper (no world-readable
   or partial-file window). Token files that the adapter reads are checked and a
   warning is logged if they are group/other-readable.

6. **Cross-account AssumeRole fails closed.** Account IDs are validated to the
   12-digit shape, and cross-account scanning is refused unless an explicit
   `AGENTCORE_ALLOWED_ACCOUNTS` allowlist authorizes the requested accounts (or
   the operator sets the separate, discouraged `AGENTCORE_ALLOW_ANY_ACCOUNT`
   opt-in). This prevents an `AGENTCORE_ACCOUNTS` value from a lower-trust source
   from silently driving AssumeRole into arbitrary accounts.

## Why it's safer / how it fails closed

- Every guard denies on error, missing config, or ambiguity.
- The OIDC egress uses the canonical guard's federation profile, which carries an
  empty bypass allowlist — no operator config can re-permit an internal target on
  this path, and the cloud-metadata endpoint is never reachable.
- The canonical URL guard is reused rather than reimplemented. To let this
  standalone tool import it without the registry's server configuration
  (SECRET_KEY, DocumentDB, etc.), the guard's settings access was made lazy — a
  pure import-decoupling that does not change validation behavior for existing
  registry consumers (verified: full existing url_guard/consumer suites pass).

## Tests

- New `tests/unit/cli/test_agentcore_security.py` covers each guard's
  fail-closed behavior (cleartext-to-remote rejected, HTTPS/loopback accepted,
  SSRF/metadata/private/non-HTTPS URLs rejected, `0o600` atomic write incl.
  overwrite of a loose-mode file, account-id shape + allowlist).
- `test_agentcore_token_refresher.py`: OIDC fetches go through the guarded
  helpers and fail closed when blocked; Cognito host-steering region and
  malformed pool-id are rejected before boto3; manifest read drops unsafe /
  missing-field / non-dict entries; look-alike hosts classify `unknown`.
- `test_agentcore_cross_account.py`: cross-account fails closed without an
  allowlist or the explicit opt-in; allowlist permits/rejects correctly.
- `test_agentcore_registration.py`: an unsafe `discovery_url` is not persisted;
  the manifest is written `0o600`; look-alike vendor hosts are not classified.
- Full existing agentcore + url_guard + downstream consumer suites remain green.

## Documentation

`docs/agentcore.md` documents the two new environment variables
(`AGENTCORE_ALLOWED_ACCOUNTS`, `AGENTCORE_ALLOW_ANY_ACCOUNT`) and the
fail-closed cross-account behavior; the CLI `--help` epilog lists them too.

### DIFF
--- a/cli/agentcore/models.py
+++ b/cli/agentcore/models.py
@@ -201,6 +201,27 @@ def _get_auth_scheme(authorizer_type: str) -> str:
     return mapping.get(authorizer_type, "none")
 
 
+def _warn_if_world_readable(abs_path: str) -> None:
+    """Log a warning if a credential file is group/other-readable.
+
+    The registry JWT is a long-lived write credential. This adapter only reads
+    the token file (it is written by the credential provider), so it cannot fix
+    the mode at write time here, but it surfaces an over-permissive file so an
+    operator notices a 0o644 token on a shared host. Best-effort: never fails.
+    """
+    try:
+        mode = os.stat(abs_path).st_mode
+        if mode & 0o077:
+            logger.warning(
+                "Token file %s is group/other-accessible (mode %o); a registry JWT "
+                "should be 0o600. Restrict it (chmod 600) to avoid local disclosure.",
+                abs_path,
+                mode & 0o777,
+            )
+    except OSError:
+        pass
+
+
 def _load_token(token_file: str) -> str:
     """Load JWT token from a JSON file.
 
@@ -212,6 +233,7 @@ def _load_token(token_file: str) -> str:
     or missing token field.
     """
     abs_path = os.path.abspath(token_file)
+    _warn_if_world_readable(abs_path)
     try:
         with open(abs_path) as f:
             data = json.load(f)
--- a/cli/agentcore/registration.py
+++ b/cli/agentcore/registration.py
@@ -29,6 +29,7 @@ from .models import (
     _slugify,
     _validate_https_url,
 )
+from .security import is_safe_egress_url, write_secret_file
 
 # Add parent directory to path for api imports
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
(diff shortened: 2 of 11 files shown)


aarora79:
Pushed a small follow-up (`2195a75f`) — docstring correction only, no logic change.

`_parse_account_ids`'s docstring said the no-allowlist case applies a shape-only check "preserving backward compatibility," but the code actually **raises** (fails closed) for a cross-account request unless `AGENTCORE_ALLOWED_ACCOUNTS` or the explicit `AGENTCORE_ALLOW_ANY_ACCOUNT` opt-in is set. Rewrote the docstring to describe the real fail-closed behavior and to call out that this is a **behavior change** for cross-account operators (the single-account default — empty `AGENTCORE_ACCOUNTS` → `[]` — is unchanged).

Worth mirroring that note in the release/upgrade notes so anyone using `AGENTCORE_ACCOUNTS` for cross-account sync sets `AGENTCORE_ALLOWED_ACCOUNTS` before upgrading.

For what it's worth, I reviewed the whole PR and verified all six guards against the code (cleartext-registry refusal, OIDC SSRF at all three points, host-anchored vendor detection defeating the look-alike, Cognito region/pool shape-check before boto3, `0o600` atomic writes, cross-account fail-closed) plus the `url_guard` lazy-settings refactor being behavior-preserving for existing consumers — 225 tests pass, ruff + bandit clean. Solid hardening; the docstring was the only real nit.

PR #1528, opened 2026-07-21
Harden ARD catalog / .well-known discovery endpoints

omrishiv:
# Harden ARD catalog / .well-known discovery endpoints

## What changed

The `.well-known` discovery surface (ARD catalog publisher, registry card,
per-server PRM) is unauthenticated by design — it serves public metadata for
spec-compliant MCP clients. This PR adds four hardening layers:

1. **Nginx rate limiting** — a dedicated `mcp_gateway_wellknown` zone (2 req/s
   per IP, burst=10) applied to `location ^~ /.well-known/`. Tighter than the
   general `mcp_gateway_edge` zone because these endpoints hit the DB on every
   request and are fully anonymous. The existing exact-match
   `/.well-known/openid-configuration` location (IdP proxy) is unaffected
   (exact match takes priority over `^~` prefix match).

2. **Query caps** — `ard_service.build_catalog` now passes
   `limit=settings.ard_catalog_max_entries_per_type` (default 500) to all three
   DB queries (servers, agents, skills). Prevents unbounded result sets from
   exhausting memory or starving connections on large registries.

3. **TOCTOU race prevention** — `_auto_initialize_registry_card` now uses
   double-checked locking (`asyncio.Lock`) so concurrent requests cannot both
   read None and both write, eliminating last-write-wins churn.

4. **Exception detail redaction** — the per-server PRM path no longer reflects
   internal exception text in the HTTP response. Generic message only;
   exception logged server-side.

One new config knob is added (`ard_catalog_max_entries_per_type`) with a safe
default requiring no operator action. Rate limiting is handled at the nginx
layer via the new zone — no application-level config needed.

## Why it is safer

- The nginx rate limit only affects `/.well-known/*` paths. Authenticated
  API traffic (`/api/`) and MCP proxy traffic are on separate zones, unaffected.
- The existing `location = /.well-known/openid-configuration` (exact match for
  the IdP metadata proxy) takes priority over the `^~` prefix match, so the
  PingFederate/IdP passthrough is not rate-limited by this zone.
- Query caps are additive: registries with fewer than 500 public entries per
  type see zero behavior change. Large registries get a bounded (but generous)
  catalog rather than a full dump.
- The asyncio lock serializes only the rare initialization path (first request
  after startup); once the card exists the fast path skips the lock entirely.
- The PRM error message change is cosmetic to legitimate clients (they only see
  502 on backend failure regardless).

## How it was tested

- New unit tests covering TOCTOU double-checked locking, exception detail
  redaction, and query-cap propagation to all three repositories.
- New assertion in `test_nginx_service.py` verifying the `mcp_gateway_wellknown`
  zone is declared in both nginx conf templates.
- Existing `test_wellknown_routes.py` and `test_ard_service.py` suites pass
  unchanged (no regressions).
- Full test suite: 6311 passed, 71 skipped, 0 failures.

### DIFF
--- a/docker/nginx_rev_proxy_http_and_https.conf
+++ b/docker/nginx_rev_proxy_http_and_https.conf
@@ -46,6 +46,10 @@ variables_hash_bucket_size 128;
 limit_req_zone $binary_remote_addr zone=mcp_gateway_edge:10m rate=50r/s;
 limit_req_zone $mcp_gateway_register_key zone=mcp_gateway_register:10m rate=5r/s;
 limit_conn_zone $binary_remote_addr zone=mcp_gateway_conn:10m;
+# - mcp_gateway_wellknown: tighter cap for unauthenticated .well-known discovery
+#   endpoints (ai-catalog.json, registry-card, oauth-protected-resource) which
+#   perform DB queries on every request and are crawled anonymously.
+limit_req_zone $binary_remote_addr zone=mcp_gateway_wellknown:5m rate=2r/s;
 
 # Return 429 (Too Many Requests) rather than the default 503 when a limit trips,
 # so clients can back off correctly.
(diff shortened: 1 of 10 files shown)


aarora79:
Followup #1555 (now merged) addressed the two non-functional review items: surfaced `ard_catalog_max_entries_per_type` in the config API/UI, and corrected the nginx `.well-known` match comment (longest-prefix, not exact-match).
Thanks for the hardening work here.

PR #1529, opened 2026-07-21
fix(egress): apply transient-failure hardening to the token vend read path (follow-up to #1490)

go-faustino:
## Problem

#1490 hardened the consent-callback **write** path against transient Vault/OpenBao unavailability (bounded retry in the store + a clear `503` "Connection not saved" page). As @aarora79 noted in the merge follow-up, the **token-vend read path** did not get the same graceful failure:

`auth-server -> /_egress_internal/egress-token -> registry -> OpenBao`

When OpenBao is briefly down (HA leader election / pod restart), the registry retries correctly, but after give-up:
- `svc.get_valid_token` raises `SecretStoreError`, which was unhandled → the vend endpoint `500`s;
- the auth-server's `_vend_egress_token` returns `None`, which falls through to a **tokenless upstream call → `401`**.

That is the same silent-failure class this line of work set out to remove, just on the **read** side. Two specific points from the follow-up:

1. **Implicit timeout ↔ retry-budget coupling.** The auth-server vend call used a hard-coded `timeout=10.0` while the registry's transient-retry budget is ~7.5s. It works today, but the two were coupled with no guardrail against drift.
2. **No retryable signal.** On vend failure the auth-server returned `None` and forwarded tokenless, surfacing as a raw upstream `401` rather than a "temporarily unavailable, retry" signal like the `503` added on the write path.

## Change

**Make the timeout ↔ retry-budget relationship explicit (single source of truth)**
- `registry/secrets/openbao/store.py` — expose `transient_retry_budget_seconds()` (sum of the bounded backoff, `0.5+1+2+4 = 7.5s` today).
- `auth_server/server.py` — derive the vend HTTP timeout from that budget + headroom in `_egress_vend_timeout_seconds()` (replaces the magic `10.0`), so the vend never abandons a call the registry is still legitimately retrying, and the two cannot silently drift.

**Read-path graceful failure, mirroring the write-path 503**
- `registry/api/egress_auth_routes.py` — `vend_egress_token` catches `SecretStoreError` from `get_valid_token` and returns `503` (not a `500`, and never a misleading `consent_required` that would tell the user to reconnect a token that is actually fine).
- `auth_server/server.py` — `_vend_egress_token` classifies transport failures/timeouts and `502/503/504` as **transient**, raising `EgressVendUnavailable`; the `mcp_proxy` hop returns a retryable `503` (JSON-RPC error `-32001`, `Retry-After`) instead of forwarding tokenless. Terminal denials (`401/403/404`) keep their existing behavior (fall through).

## Test plan

- `tests/unit/egress_auth/test_internal_egress_token_route.py` — vend returns `503` when the store fails transiently (never `consent_required`).
- `tests/unit/secrets/test_stores.py` — `transient_retry_budget_seconds()` matches the backoff sum and equals `7.5s` (drift guard).
- `tests/auth_server/unit/test_server.py` — `_vend_egress_token` transient classification (transport/timeout/`5xx` → raise; `401/403/404` → `None`; `200` → body); vend timeout > retry budget; `mcp_proxy` returns a retryable `503` on transient vend failure.

Ran the affected suites locally: `414 passed, 16 skipped`. `ruff check` + `ruff format` clean on changed files.

Follow-up to #1490 (non-blocking there; opened as a separate PR as suggested).

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -6931,6 +6931,20 @@ async def mcp_proxy(
         relay_authorization=relay_ingress_auth,
     )
 
+    # Re-authorize the caller's scopes against the EXACT body we are about to act
+    # on, BEFORE any outbound work (egress vend or upstream forward). /validate
+    # authorizes on a separately-captured copy (X-Body) that can diverge from this
+    # body (e.g. a large body that spilled to disk, which /validate then treats as
+    # an unprivileged "initialize"), so we authorize the forwarded bytes here and
+    # fail closed on any body we cannot parse (TM-15). Running this FIRST is a
+    # security ordering guarantee: an unscoped caller is denied 403 regardless of
+    # egress-vend / OpenBao availability -- otherwise a transient vend failure
+    # (503) could mask a genuine authorization denial, turning a "denied" into a
+    # "retry later" in responses, logs, and metrics. Authorized-but-not-connected
+    # callers still pass here and reach the egress consent/local-answer paths
+    # below (they already had these methods in scope on the connected path).
+    await _authorize_forwarded_mcp_body(server_name, request_body, user_scopes)
+
     # True once we inject a vaulted egress token below. An egress upstream is
     # itself an OAuth resource server: if it rejects our injected token it 401s
     # with its OWN WWW-Authenticate (resource_metadata pointing at the upstream's
@@ -7184,18 +7198,6 @@ async def mcp_proxy(
                     req_id=req_id,
                     vend=vend,
                 )
-
-    # Re-authorize the caller's scopes against the EXACT body we are about to
-    # forward upstream. /validate authorizes on a separately-captured copy
-    # (X-Body) that can diverge from this body (e.g. a large body that spilled to
-    # disk, which /validate then treats as an unprivileged "initialize"), so we
-    # authorize the forwarded bytes here and fail closed on any body we cannot
-    # parse (TM-15). This runs AFTER the egress-consent block above so methods the
-    # gateway answers LOCALLY for a tokenless egress server (initialize,
-    # notifications/*, tools/list, tools/call consent) are never gated here --
-    # they are not forwarded upstream. Only the request we actually forward is
-    # re-authorized. Raises 403 before the outbound call.
-    await _authorize_forwarded_mcp_body(server_name, request_body, user_scopes)
 
     logger.info(
         f"mcp_proxy: server={server_name} method={incoming_method} filter_enabled={filter_enabled} timeout={proxy_timeout}"
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -6233,6 +6233,50 @@ _EGRESS_STRIP_HEADERS: frozenset[str] = frozenset(
 )
 
 
+class EgressVendUnavailable(Exception):
+    """The registry's egress-token vend failed *transiently* -- the registry was
+    unreachable/timed out, or it answered 5xx after riding out its own bounded
+    Vault-retry budget.
+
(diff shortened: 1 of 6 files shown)


aarora79 (changes requested):
One blocker before merge: a pre-existing security test now fails.

`tests/auth_server/unit/test_server.py::TestMcpProxyEndpointHeaderPassthrough::test_forwarded_body_reauthorized_denies_unscoped_caller` (1 failed, 249 passed in that file).

That test asserts a scopeless caller is denied `403` before any outbound work (the TM-15 guarantee). It targets the egress server `office-docs` and does not mock the vend. With this change, the vend runs first, fails transiently in the unit env (no registry to reach), raises `EgressVendUnavailable`, and the proxy returns a retryable `503`, short-circuiting before the `403` scope re-authorization.

I do not think this is just a test that needs a mock. In `mcp_proxy` the egress vend happens before the forwarded-body re-authorization:

- vend: `auth_server/server.py` ~line 6832
- 403 re-authorization: ~line 7051

So while OpenBao is briefly unavailable, an authorization denial is masked by a transient-availability signal: an unscoped caller gets `503` "retry later" instead of `403` "denied." It is not an auth bypass (no token is vended and no upstream call is made either way), but it does weaken the "authorize before any outbound work" property the TM-15 test exists to protect, and it can make a genuine denial look like a transient blip in logs and metrics.

Suggested fix: run the caller scope re-authorization before the egress vend, so a denial is always `403` regardless of vend or OpenBao availability. The existing test then passes for the right reason rather than by mocking the vend green. Worth adding a regression test too: unscoped caller gets `403` even when the vend would raise `EgressVendUnavailable`.

For context, the rest of the change looks good: the derived timeout removes the drift risk (`transient_retry_budget_seconds()` computed from the same backoff constants, consumed by `_egress_vend_timeout_seconds()`), the transient-vs-terminal classification is careful (transport/timeout + 502/503/504 raise `EgressVendUnavailable`; 401/403/404 still fall through to consent), your new tests pass, and lint/bandit are clean.

go-faustino:
Good catch — you're right that this was an ordering hazard, not a test that needs a mock. Fixed in a2fb0ea by moving the caller scope re-authorization to run **before** the egress vend, exactly as you suggested.

**Change:** `_authorize_forwarded_mcp_body(...)` now runs right after `forward_headers` is built, before the `settings.egress_auth_enabled` block — so an unscoped caller is denied `403` before any outbound work, regardless of egress-vend / OpenBao availability. A transient vend `503` can no longer mask a genuine authorization denial (in the response, or in logs/metrics).

**No regression for authorized callers.** The connected egress and obo paths already fell through to this same re-auth downstream, so anyone with the server's scopes still passes and reaches the vend and the consent/local-answer paths. The only behavior change is that an *unauthorized* caller now gets `403` instead of a local consent nudge on a tokenless egress server — which is the intended hardening.

**Tests:**
- `test_forwarded_body_reauthorized_denies_unscoped_caller` now passes for the right reason (denial precedes the vend), not by mocking the vend green.
- Added the regression you asked for: `test_unscoped_caller_denied_403_before_vend_even_if_vend_would_fail` — an unscoped caller gets `403` even when `_vend_egress_token` would raise `EgressVendUnavailable`, and asserts the vend is **never attempted**.
- The egress-consent / obo-terminal / vend-unavailable tests now represent an authorized caller (they legitimately reach the egress block, which sits behind the scope check) via `_patch_scope_repo_allow_all`.
- `tests/auth_server/unit/test_server.py`: **251 passed**; `ruff check` + `format` clean.

Thanks for the careful review — re-requesting when you have a moment.

PR #1546, opened 2026-07-24
feat(#1494): make UI scope creation authoritative (reload + validated config)

### LINKED ISSUE #1494: Scope level access in UI
Please add the scope cretion in UI level - instead of doing everytime in mongod insert.

aarora79:
Closes #1494

## Summary

Makes UI-driven scope creation authoritative: changes made through the IAM Groups UI now take effect immediately (auth-server reload) and are validated with Pydantic models instead of a free-form dict.

## Changes

- **Reload after mutation:** trigger an auth-server scope reload after IAM group create/update/delete, mirroring the CLI import path, so UI-created scopes apply immediately without an auth-server restart. Reload failure is non-fatal (logged as WARNING; the persisted change is authoritative).
- **Validated config models:** replace the free-form `scope_config` dict with validated `ScopeConfig` / `ServerAccessRule` / `AgentAccessRule` Pydantic models (`extra=forbid`). Malformed input now returns an actionable 422 instead of a silent drop or opaque 500. Access fields default to `None` so PATCH preserves omitted fields (critical for Entra Object IDs in `group_mappings`); the create handler coerces `None` to empty collections, keeping the persisted doc shape byte-identical (`model_dump(exclude_unset=True)`).
- **Guard refusals surfaced:** `import_group` guard refusals (reserved wildcard, privileged write) now return a 400 on create and update instead of a success response that persisted nothing. Added `except HTTPException: raise` to the create handler so the 400 is not swallowed by the IAM error translator.
- **Docs/comments:** fixed the stale "not yet applied server-side" docstring and frontend comment; documented "Managing scopes from the UI (recommended)" in `docs/scopes-mgmt.md`, including the wildcard-server limitation.

## Tests

- Reload fires on all three mutations; reload-failure is non-fatal; refusal-400 skips reload.
- 422 on unknown keys / malformed rules; `scope_config` reaches `import_group`; PATCH preserves omitted fields.
- New `tests/unit/schemas/test_management_models.py` for the models.

## Note

Branch is behind `main` and will need a rebase before merge.

### DIFF
--- a/docs/scopes-mgmt.md
+++ b/docs/scopes-mgmt.md
@@ -403,6 +403,36 @@ Access to specific MCP servers and one agent:
 
 ## Managing Scopes
 
+### Managing scopes from the UI (recommended)
+
+Scopes can be created, edited, and deleted entirely from the registry UI — no
+direct database access or JSON file editing is required:
+
+1. Navigate to **Settings → IAM → Groups** and click **Create Group** (or
+   **Edit** on an existing group).
+2. Fill in the group name, description, and group mappings, then use the
+   pickers to grant **Server Access** (per-server methods and tools),
+   **Agent Access**, and **UI Permissions**. Alternatively, drag and drop an
+   existing scope JSON file onto the form to pre-fill it.
+3. Submit. The scope is validated, persisted, and the auth server is reloaded
+   automatically — **the scope takes effect immediately**, without restarting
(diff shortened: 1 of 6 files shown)


aarora79:
Closing this, we will have a single PR which has all the backend and frontend changes needed for this functionality.

PR #1549, opened 2026-07-25
feat(#1494): make UI scope creation authoritative (reload + validated…

vrindabhandari:
*Issue #, if available:* Part of #1494 — UI implementation per the wireframes posted on the issue. **Stacked on #1546** (backend): the first commit here is the same backend commit as #1546; only the top commit is new. Will rebase to the pure UI delta once #1546 merges.

*Description of changes:*

First slice of the group-form redesign from the #1494 wireframes.

- **New `iam/GroupAccessPanel.tsx`** — a tab rail (Servers / Agents / Permissions, with live counts) and a compact server-access row table, shared by the create and edit forms. Replaces the stacked Server Access / Agent Access / UI Permissions sections that were duplicated across both forms in `IAMGroups.tsx`.
- **Methods as a chip popover** — the 11-checkbox methods grid collapses to a chip summary ("all" / "GET, POST" / "initialize +3") that opens a checkbox popover.
- **Explicit clears on update** — `server_access`, `agent_access`, and `ui_permissions` are now sent unconditionally in the update payload, so an emptied selection is an explicit clear rather than an omitted field (which the PATCH semantics from #1546 deliberately preserve — previously, removing all agents and saving silently kept the old grant).
- **Honest Permissions count** — the tab badge excludes the auto-synced service/agent permission keys (`list_service`, `call_tool`, etc.), which previously made the count include grants the user never made.

Known omitted-field cases left as-is by design: clearing `group_mappings` from the UI preserves existing mappings (protects Entra Object IDs, per #1546), and same for `description`. Can add explicit-clear affordances in a follow-up if desired.

**Next commits on this PR:** summary sidebar + demoting the JSON preview to collapsed/read-only, and mapping the 422 field errors / 400 guard refusals from #1546 to inline row errors and a banner.

Verified locally: production build clean, manual round-trip on create and edit (values populate, save-unchanged preserves, clears persist).


vrindabhandari:
Consolidating into #1546 per Amit's request - that PR now contains the backend, the UI implementation, and the CI fix.

vrindabhandari:
Reopening - single consolidated PR per the discussion: backend, UI implementation, CI fix, merged current with main. #1546 stays closed.

PR #1598, opened 2026-08-04
feat: add static bearer token fast-path for egress auth

hochbit:
Author: Claude Sonnet 4.6

### Problem

MCP servers registered with `auth_scheme: bearer` — i.e. a simple static
token stored encrypted in MongoDB via the registry UI — were not reachable
through the gateway proxy. When a client request hit `/mcp-proxy/<server>/`,
the auth-server correctly validated the inbound user token and forwarded the
request, but the backend received no `Authorization` header and rejected it
with 401.

The root cause was that the egress credential vend endpoint
(`/_egress_internal/egress-token`) had two gaps:

1. When `EGRESS_AUTH_ENABLED=false` (the default), the endpoint immediately
   returned 404, so the auth-server never attempted a vend and forwarded
   tokenless.
2. When `EGRESS_AUTH_ENABLED=true`, the vend only handled `oauth_user`,
   `obo_exchange`, and `pat` modes. Servers with a plain `bearer`
   `auth_credential_encrypted` but no `egress_auth_mode` fell through to
   `consent_required=True`, again resulting in a tokenless forward.

This meant that any MCP server requiring a static bearer token — a common
pattern for in-cluster services like mcpg, custom sidecars, or internal
tooling — was effectively unreachable through the gateway even when the
credential was correctly configured in the UI.

### Fix

Two fast-paths added to `vend_egress_token` in
`registry/api/egress_auth_routes.py`:

**Path 1 — egress disabled, server has bearer credential:**
Before raising 404, check whether the requested server has
`auth_scheme=bearer` and a stored `auth_credential_encrypted`. If so,
decrypt and vend it immediately. This allows operators to use bearer-token
backends without enabling the full per-user vault stack (OpenBao / Secrets
Manager).

**Path 2 — egress enabled, server has bearer credential but no egress mode:**
After the per-user auth check and server lookup, if `egress_auth_mode` is
not set to one of the vault-backed modes but `auth_scheme=bearer` is
present, decrypt the stored credential and vend it. This handles the case
where egress is enabled for OAuth/PAT servers but a particular server only
needs a static token.

In both paths the credential is decrypted using the existing
`decrypt_credential` helper (the same one used by the health checker), and
the inject header/prefix is derived via `_derive_pat_inject_header` — so
the server's `auth_header_name` configuration is honoured consistently.

### Behaviour

- No vault required for static bearer servers — no OpenBao or Secrets
  Manager configuration needed.
- Fail-closed: if decryption fails or the credential is absent, the
  existing `consent_required` / 404 response is preserved.
- No change to PAT, `oauth_user`, or `obo_exchange` flows — those are
  unaffected.
- The `EGRESS_AUTH_ENABLED=false` 404 is preserved for servers that have
  no bearer credential, keeping the existing fail-closed default.

### Testing

Verified end-to-end against a Kubernetes-deployed mcpg (PostgreSQL MCP
server) registered with `auth_scheme: bearer`. After this change, MCP
`initialize` and `tools/call` requests through the gateway correctly
received the backend bearer token and returned valid responses.
*Description of changes:*

### DIFF
--- a/registry/api/egress_auth_routes.py
+++ b/registry/api/egress_auth_routes.py
@@ -356,11 +356,28 @@ async def vend_egress_token(
     missing/invalid; a clean miss (no connection, non-per-user caller, upstream
     mismatch, etc.) returns ``consent_required=True`` with no token.
     """
-    if not settings.egress_auth_enabled:
-        raise HTTPException(status.HTTP_404_NOT_FOUND, detail="egress auth disabled")
-
     if not x_internal_token:
         raise HTTPException(status.HTTP_401_UNAUTHORIZED, detail="missing X-Internal-Token")
+
+    # Static bearer fast-path: when egress_auth is disabled but the server has
+    # a stored bearer credential, vend it without requiring the full vault stack.
+    if not settings.egress_auth_enabled:
+        server_path = body.server_path if body.server_path.startswith("/") else "/" + body.server_path
+        server = await get_server_repository().get(server_path)
+        if server is not None and server.get("auth_scheme") == "bearer":
+            encrypted = server.get("auth_credential_encrypted")
+            if encrypted:
+                from ..utils.credential_encryption import decrypt_credential
+
+                credential = decrypt_credential(encrypted)
+                if credential:
+                    header_name, value_prefix = _derive_pat_inject_header(server)
+                    return EgressTokenResponse(
+                        access_token=credential,
+                        pat_header_name=header_name,
+                        pat_value_prefix=value_prefix,
+                    )
+        raise HTTPException(status.HTTP_404_NOT_FOUND, detail="egress auth disabled")
 
     # Independently re-verify the mcp-proxy token; identity is the verified
     # claim, never an asserted body field.
@@ -392,6 +409,25 @@ async def vend_egress_token(
 
     # Per-server enablement: a misconfigured/half-deleted server never vends.
     egress_mode = server.get("egress_auth_mode")
+
+    # Static bearer: server has a stored encrypted credential (auth_scheme=bearer)
+    # but no per-user egress flow. Decrypt and vend directly — no vault required.
+    auth_scheme = server.get("auth_scheme", "none")
+    if egress_mode not in ("oauth_user", "obo_exchange", "pat") and auth_scheme == "bearer":
+        encrypted = server.get("auth_credential_encrypted")
+        if encrypted:
+            from ..utils.credential_encryption import decrypt_credential
+
+            credential = decrypt_credential(encrypted)
+            if credential:
+                header_name, value_prefix = _derive_pat_inject_header(server)
+                return EgressTokenResponse(
+                    access_token=credential,
+                    pat_header_name=header_name,
+                    pat_value_prefix=value_prefix,
+                )
+        return EgressTokenResponse(consent_required=True)
+
     if egress_mode not in ("oauth_user", "obo_exchange", "pat") or not server.get("egress_oauth"):
         return EgressTokenResponse(consent_required=True)
 


ebarkhordar (commented):
Both new branches return before checks the existing vend applies, and the second one drops an invariant this repo already has a test for.

**Path 2 skips the upstream cross-check.** `test_upstream_mismatch_403`
(`tests/unit/egress_auth/test_internal_egress_token_route.py:138`) pins that a forged
`upstream_url` is refused. The new bearer branch at :413-429 returns before
`_registered_upstreams` at :437, so the same forged upstream vends. Driving this file's own
`make_client` fixture at ce28dd7b and changing only the server's `auth_scheme` to `bearer`:

```
upstream_url = "https://attacker.example/mcp"   (not in the registered set)
  existing oauth_user path -> 403
  new bearer branch        -> 200 {"access_token": "sekrit-static-token", ...}
```

The comment at :435 says the cross-check "applies to BOTH egress modes". The bearer branch is
effectively a third mode that never reaches it.

**Path 1 never verifies the proxy token.** With `egress_auth_enabled=false` the branch at
:362-380 runs ahead of `verify_mcp_proxy_token` at :384. Replacing that function with one that
raises, the request still came back 200 with the credential, so it is not called on this path,
and neither is `is_per_user_auth_method`. That leaves `body.server_path` as the only input
selecting which server's credential is returned, which is what the comment at :383 rules out:
"identity is the verified claim, never an asserted body field."

`validate_internal_auth` still gates the route, so neither branch is reachable unauthenticated,
and I have not worked out how far that narrows the exposure in a real deployment. Calling that
is yours. What I can say is that both branches skip gates the surrounding code applies on
purpose, and one of them is already covered by the suite.

Verified in a clean python:3.14-slim container at ce28dd7b, through the existing fixture with
`decrypt_credential` stubbed to a known string. The file's 13 current tests still pass.

Would moving both branches below `verify_mcp_proxy_token` and the upstream check work, so the
bearer case reuses the same gates and only the credential source differs?

aarora79:
Thanks for this @ebarkhordar, and for the detailed writeup and end-to-end verification. You've identified a real capability gap: a backend registered with `auth_scheme: bearer` isn't reachable for user tool calls through the gateway, and that's worth supporting.

We'd like to treat this as a **feature request for a new egress mode** rather than a direct patch, because static-bearer egress on the user path has a couple of design points we want to make explicit. Some context that shapes the design:

- The vend endpoint (`vend_egress_token`) is on the **per-user data path only** -- its sole caller is the `mcp-proxy` hop in the auth-server. Registry-internal use like health checks doesn't go through it; the health checker already decrypts the stored credential directly, so static-bearer health checks work today.
- On this path the stored token is injected for **every user's** tool call, so the backend sees a single shared identity. That's a reasonable model for a trusted internal service where the gateway is the access-control boundary (per-user authorization still happens via scopes at `/validate`), but for that reason we don't want it to switch on implicitly for any server that merely has `auth_scheme=bearer` set -- it should be a deliberate operator choice.
- One structural note on the vend itself: the registry is the component that holds the backend credential and decrypts it on a caller's behalf, so it's acting as a deputy for the request. In the egress-disabled fast-path the credential is vended based on the body-supplied `server_path` before the internal token is verified (`verify_mcp_proxy_token` runs later in the enabled branch), and the credential isn't re-bound to the resolved destination. To keep the deputy from acting on an unverified/asserted request, the mode should verify the caller's token and bind to the resolved destination before decrypting. This is standard for the endpoint -- the OAuth/PAT paths already do it -- so it's mostly about applying the same discipline to the new mode.

Here's the shape we'd propose, explicit and fail-closed at two layers:

1. **A new egress mode** (e.g. `egress_auth_mode = "static_bearer"`). The shared bearer is vended only when an admin has selected this mode for that server, never inferred from `auth_scheme=bearer` alone.
2. **A global config flag** (e.g. `EGRESS_ALLOW_STATIC_BEARER`, default `false`) that enables the mode at the deployment level. Both the global flag and the per-server mode must be set; otherwise the endpoint keeps its current `consent_required` / 404 behavior.

We'd document the mode's tradeoff -- per-user access control stays at the gateway via scopes, but the backend sees one identity, so it fits trusted internal services rather than backends needing per-user attribution -- and keep `oauth_user` / `obo_exchange` as the default for user access.

Really useful contribution, and it's the right feature. We'd like to land it in this shape and credit your original patch. Would you like to iterate toward the explicit-mode version, or would you prefer we take it from here?

PR #1602, opened 2026-08-05
fix(cli): forward the custom-OIDC fields so egress-configure --provider custom works

### LINKED ISSUE #1601: egress-configure --provider custom is unusable: CLI never forwards the custom_authorize_url/custom_token_url the server requires
## Summary

`egress-configure --provider custom` can never succeed: the CLI accepts the flag but forwards none of the `custom_*` fields, and the server requires two of them for a custom provider. The UI already exposes them, so this is a CLI/UI parity gap.

`resolve_provider` rejects the config server-side:

```
custom requires custom_authorize_url and custom_token_url
```

## Reproduction

```bash
uv run python api/registry_management.py \
  --token-file <admin.json> --registry-url http://localhost \
  egress-configure --path /my-server --mode oauth_user --provider custom \
  --client-id "$CID" --client-secret "$SEC" --scopes 'mcp_api,refresh_token'
```

There is no flag to supply the authorize/token URLs, so the request is always incomplete. The only workaround is calling `POST /api/servers/{path}/egress-auth` directly with curl.

## Why this is a gap rather than a design choice

- The API request model accepts five `custom_*` fields: `custom_authorize_url`, `custom_token_url`, `custom_scope_separator`, `custom_token_auth_style`, `custom_resource` (`registry/api/egress_auth_routes.py`).
- `EgressOAuthConfig` persists four of them (`registry/core/schemas.py`, "Custom-OIDC overrides (only when provider == 'custom')").
- **The frontend already sends them** — `frontend/src/components/entities/forms/ServerEditModal.tsx` references `custom_authorize_url` / `custom_token_url`, with its own test coverage.
- `PROVIDER_REGISTRY`'s docstring frames `custom` as a first-class option: *"A `custom` provider is assembled at runtime from the server's `EgressOAuthConfig.custom_*` fields."*
- `list_providers()` returns `sorted(PROVIDER_REGISTRY.keys()) + ["custom"]`, so `custom` is advertised as selectable.

Nothing is protected by omitting the flags; the CLI is simply an incomplete passthrough.

## Real-world case

Registering **Salesforce Hosted MCP (Headless 360)**, whose OAuth endpoints are per-org:

```
https://<my-domain>.my.salesforce.com/services/oauth2/authorize
https://<my-domain>.my.salesforce.com/services/oauth2/token
```

A per-org provider cannot be a static `PROVIDER_REGISTRY` row, so `custom` is the correct mechanism — and it was unreachable from the CLI.

## Fix

Add the five fields to `RegistryClient.configure_egress_auth` and the matching `--custom-*` flags to `egress-configure`, following the existing `if x is not None` convention so omitted flags send nothing (no null overwrite on edit). `--custom-token-auth-style` is constrained to `post_body|basic_header` to match `TokenEndpointAuthStyle`.

`custom_resource` is included even though the Salesforce case does not need it — it is the RFC 8707 indicator added for Atlassian's Rovo MCP, and leaving it out would reproduce this same gap for the next per-resource provider.

## Verification

Configured a real Salesforce Hosted MCP server end-to-end using **only** the CLI (previously impossible), after first resetting the server to `--mode none` so the write was genuine:

```
egress_auth_mode : oauth_user
egress_provider  : custom
scopes           : ['mcp_api', 'refresh_token']
custom_authorize_url : https://...my.salesforce.com/services/oauth2/authorize
custom_token_url     : https://...my.salesforce.com/services/oauth2/token
```

`POST /api/egress-auth/initiate` then returns HTTP 200 with a correct authorize URL — org host, `scope=mcp_api refresh_token`, `code_challenge_method=S256`, and the registry's own callback as `redirect_uri`. The server also appears in `GET /api/egress-auth/available-servers`, i.e. in the Connected Accounts dropdown.

Adds `tests/unit/api/test_registry_client_egress_custom.py` (6 cases): the two required URLs reach the body, all five fields are forwarded, omitted fields send no `custom_*` keys, the endpoint/method are unchanged, and the CLI command wires its args through. **4 of the 6 fail against current `main`.**

Full suite: **6734 passed, 74 skipped**. `ruff format` clean; `ruff check` reports only 4 pre-existing `UP042` enum warnings in untouched regions of `registry_client.py`.

PR to follow.

## Related, filed separately in spirit

Two other things surfaced while registering that server, if useful:

- `append_mcp_path` is silently dropped when passed in a `register --config` JSON; it only persists via `patch-server`. Salesforce is strict here (root returns 200, `/mcp` returns 404), so the wrong value breaks the server.
- `register --overwrite` resets `is_enabled` to `false`, silently disabling a previously enabled server.

billtarr-aws:
Fixes #1601

## Problem

`egress-configure --provider custom` could never succeed. argparse accepted the flag, but neither the CLI nor `RegistryClient.configure_egress_auth` forwarded any of the five `custom_*` fields, and the server's `resolve_provider` requires two of them:

```
custom requires custom_authorize_url and custom_token_url
```

The only way to configure a custom OIDC provider was `curl` against `POST /api/servers/{path}/egress-auth`.

## Why it's a gap, not a restriction

- The API request model accepts all five `custom_*` fields
- `EgressOAuthConfig` persists four of them
- **The frontend already sends** `custom_authorize_url` / `custom_token_url` (`ServerEditModal.tsx`, with its own tests) — so this is a CLI/UI parity gap
- `list_providers()` advertises `custom` as selectable

Nothing was protected by omitting the flags.

## Change

Five optional params on `configure_egress_auth` + five `--custom-*` flags, following the existing `if x is not None` convention so an omitted flag sends nothing and cannot null out a stored value on edit. `--custom-token-auth-style` is constrained to `post_body|basic_header` to match `TokenEndpointAuthStyle`.

`custom_resource` is included although the motivating case doesn't need it — it's the RFC 8707 indicator added for Atlassian's Rovo MCP, and omitting it would reproduce this gap for the next per-resource provider.

## Motivating case

**Salesforce Hosted MCP (Headless 360)**, whose OAuth endpoints are per-org:

```
https://<my-domain>.my.salesforce.com/services/oauth2/{authorize,token}
```

A per-org provider can't be a static `PROVIDER_REGISTRY` row, so `custom` is the right mechanism — and it was unreachable from the CLI.

## Verification

Configured a live Salesforce org end-to-end using **only** the CLI, after first resetting to `--mode none` so the write was genuine:

```
egress_auth_mode     : oauth_user
egress_provider      : custom
scopes               : ['mcp_api', 'refresh_token']
custom_authorize_url : https://...my.salesforce.com/services/oauth2/authorize
custom_token_url     : https://...my.salesforce.com/services/oauth2/token
```

`POST /api/egress-auth/initiate` then returns **200** with a correct authorize URL — org host, `scope=mcp_api refresh_token`, `code_challenge_method=S256`, and the registry's own `/oauth2/egress/callback` as `redirect_uri`. The server also appears in `GET /api/egress-auth/available-servers`, i.e. the Connected Accounts dropdown.

## Tests

`tests/unit/api/test_registry_client_egress_custom.py`, 6 cases: required URLs reach the body, all five fields forwarded, omitted fields send no `custom_*` keys, endpoint/method unchanged, and the CLI command wires args through. **4 of 6 fail against current `main`** — verified by stashing the fix.

Full suite: **6734 passed, 74 skipped**. `ruff format` clean; `ruff check` shows only 4 pre-existing `UP042` warnings in untouched regions.

### DIFF
--- a/api/registry_client.py
+++ b/api/registry_client.py
@@ -3883,6 +3883,11 @@ class RegistryClient:
         client_secret: str | None = None,
         scopes: list[str] | None = None,
         target_audience: str | None = None,
+        custom_authorize_url: str | None = None,
+        custom_token_url: str | None = None,
+        custom_scope_separator: str | None = None,
+        custom_token_auth_style: str | None = None,
+        custom_resource: str | None = None,
     ) -> dict[str, Any]:
         """Configure per-user egress auth on a server (admin only).
 
@@ -3898,6 +3903,18 @@ class RegistryClient:
             client_secret: OAuth client secret (oauth_user only, write-only).
             scopes: Optional list of OAuth scopes.
             target_audience: Target audience (obo_exchange only).
+            custom_authorize_url: Authorize endpoint. REQUIRED when
+                ``provider="custom"`` (server-side ``resolve_provider`` rejects
+                the config without it).
+            custom_token_url: Token endpoint. REQUIRED when
+                ``provider="custom"``.
+            custom_scope_separator: Scope delimiter when the provider does not
+                use a space (custom only).
+            custom_token_auth_style: Where the client secret goes on the token
+                request -- "post_body" (default) or "basic_header" (custom only).
+            custom_resource: RFC 8707 resource indicator, sent on both the
+                authorize and token requests to bind the token to one protected
+                resource (custom only).
 
         Returns:
             Non-secret egress config view dict.
@@ -3920,6 +3937,19 @@ class RegistryClient:
             body["scopes"] = scopes
         if target_audience is not None:
             body["target_audience"] = target_audience
+        # Custom-OIDC overrides. Only meaningful when provider == "custom"; the
+        # server ignores them for built-in providers, so they are forwarded
+        # unconditionally like the fields above rather than gated here.
+        if custom_authorize_url is not None:
+            body["custom_authorize_url"] = custom_authorize_url
+        if custom_token_url is not None:
+            body["custom_token_url"] = custom_token_url
+        if custom_scope_separator is not None:
+            body["custom_scope_separator"] = custom_scope_separator
+        if custom_token_auth_style is not None:
+            body["custom_token_auth_style"] = custom_token_auth_style
+        if custom_resource is not None:
+            body["custom_resource"] = custom_resource
 
         response = self._make_request(
             method="POST",
--- a/api/registry_management.py
+++ b/api/registry_management.py
@@ -2035,6 +2035,11 @@ def cmd_egress_configure(args: argparse.Namespace) -> int:
             client_secret=args.client_secret,
             scopes=_parse_scopes(args.scopes),
             target_audience=args.target_audience,
+            custom_authorize_url=args.custom_authorize_url,
+            custom_token_url=args.custom_token_url,
+            custom_scope_separator=args.custom_scope_separator,
+            custom_token_auth_style=args.custom_token_auth_style,
+            custom_resource=args.custom_resource,
         )
         print(json.dumps(response, indent=2, default=str))
         return 0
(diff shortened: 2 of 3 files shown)


omrishiv (approved):
LGTM, thanks for updating the client for these

PR #1610, opened 2026-08-06
fix: use relative redirects to prevent internal URL leak (#1606)

### LINKED ISSUE #1606: bug: MCP server endpoint without trailing slash returns a 301 to an internal `http://<host>:8080/...` origin, causing MCP clients to hang/time out
## Summary

After the fix for #1501 (PR #1507) made the generated nginx location blocks
trailing-slash-terminated (`location /<server>/`), a request to a registered MCP
server **without** the trailing slash now triggers nginx's automatic
"add-the-slash" **301 redirect**. Because the registry container's front-door
nginx uses `absolute_redirect on` (the nginx default), that redirect's `Location`
is built from the **internal** listener scheme and port rather than the
external-facing origin:

```
Location: http://<gateway-host>:8080/registry/<server>/
```

i.e. plaintext `http://` (TLS actually terminates upstream, e.g. at an Istio/ALB
gateway) and the **internal** container port `:8080`, which is not reachable from
outside the cluster. MCP clients (e.g. Claude Code / the MCP SDK) **follow** this
301 and then **hang until they time out** on the unreachable `:8080` address.

Net effect: the MCP endpoint only works when the client uses the **trailing-slash**
URL. This is made worse by the registry UI's "connect" dialog, which presents the
endpoint **without** the trailing slash (see "Additional" below).

## Environment

- Version: **1.28.0** (`charts/mcp-gateway-registry-stack`)
- Deployment: Kubernetes / Helm, `registry.app.deploymentMode: with-gateway`
  (the front-door nginx runs inside the `registry` container, listening on `:8080`
  internally; **TLS is terminated upstream** by the ingress gateway).
- Routing: `global.ingress.routingMode: path` (servers exposed under `/registry/...`).

## Steps to reproduce

Register any streamable-HTTP MCP server (path `/mcp-sonar` in this example), then
POST an `initialize` to the endpoint **without** a trailing slash vs **with** one.

**Without trailing slash:**
```
$ curl -sS -i -X POST "https://<gateway-host>/registry/mcp-sonar" \
    -H "Authorization: Bearer <token>" \
    -H "Accept: application/json, text/event-stream" \
    --data '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{...}}'

HTTP/1.1 301 Moved Permanently
Location: http://<gateway-host>:8080/registry/mcp-sonar/     <-- http + :8080 (internal)
Content-Type: text/html
```

**With trailing slash:**
```
$ curl -sS -X POST "https://<gateway-host>/registry/mcp-sonar/" ...
HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"...","capabilities":{...},"serverInfo":{...}}}
```

An MCP client configured with the no-slash URL follows the 301 to
`http://<gateway-host>:8080/...` and **times out** (the port is not exposed
externally and the scheme is wrong). curl only shows the fast 301 because it does
not follow the redirect by default; the MCP SDK does follow it.

## Expected behavior

A request to the no-slash URL should either be served directly, or redirected to a
**reachable** external origin — i.e. the `Location` should preserve the external
scheme/host and **not** leak the internal `http`/`:8080`. Ideally a relative
`Location: /registry/mcp-sonar/`.

## Root cause

- #1501 / PR #1507 normalized the generated location blocks to a trailing slash
  (`location /<server>/`) in `registry/core/nginx_service.py`. This is correct for
  the route-hijacking bug it fixed, but it means a no-slash request no longer
  matches and instead gets nginx's automatic trailing-slash 301.
- That 301's `Location` is absolute (`absolute_redirect on`, nginx default) and is
  constructed from the internal listener: scheme `http` (TLS terminates upstream)
  and `port_in_redirect on` appending the internal `:8080`. Behind a
  TLS-terminating reverse proxy this produces an unreachable URL.

This is effectively an **unaddressed side-effect of the #1501/#1507 fix**. #1501
itself does not mention 301s, `absolute_redirect`, the internal port, or the
resulting timeout.

## Suggested fix

In the generated nginx config (front-door `server {}` and/or the MCP location
blocks in `nginx_service.py`), one of:

- `absolute_redirect off;` — emit a **relative** `Location: /registry/<server>/`,
  which the client resolves against the original `https://<gateway-host>` origin.
  Simplest and robust behind TLS-terminating proxies.
- or `port_in_redirect off;` plus deriving the scheme from `X-Forwarded-Proto`
  (e.g. `map $http_x_forwarded_proto $redirect_scheme { ... }` / `absolute_redirect`
  with correct scheme) so the redirect uses `https://<host>` without the internal
  port.

Either makes the no-slash URL work transparently and removes the client-facing
timeout.

## Additional: UI connect dialog omits the trailing slash

The registry UI's server "connect" dialog presents the MCP endpoint URL **without**
the trailing slash, which is exactly the form that triggers the broken 301. Until
the redirect is fixed, the dialog should present the **slash-terminated** URL so
copy-paste into an MCP client works out of the box.

## References

- #1501 — MCP server nginx location blocks without trailing slash silently hijack
  unrelated browser routes (closed; fixed by PR #1507). This bug is a side-effect
  of that fix.

zsxh1990:
## Problem

After PR #1507 added trailing slashes to nginx location blocks, requests without trailing slash trigger nginx's automatic 301 redirect. With `absolute_redirect on` (the default), the `Location` header leaks the internal scheme and port:

```
Location: http://<gateway-host>:8080/registry/<server>/
```

MCP clients follow this redirect and hang on the unreachable internal address.

## Fix

Add `absolute_redirect off;` to both nginx templates so redirects use relative `Location` headers:

```
Location: /registry/<server>/
```

This is resolved by the client against the original `https://<gateway-host>` origin, which works correctly behind TLS-terminating proxies.

## Changes

- `docker/nginx_rev_proxy_http_and_https.conf`: Add `absolute_redirect off;`
- `docker/nginx_rev_proxy_http_only.conf`: Add `absolute_redirect off;`

Fixes #1606

### DIFF
--- a/docker/nginx_rev_proxy_http_and_https.conf
+++ b/docker/nginx_rev_proxy_http_and_https.conf
@@ -165,6 +165,8 @@ map $host $rl_retry {
 {{VERSION_MAP}}
 # First server block now directly handles HTTP requests instead of redirecting
 server {
+    # Use relative redirects to avoid leaking internal scheme/port behind TLS-terminating proxies
+    absolute_redirect off;
     listen 8080;
     # {{ADDITIONAL_SERVER_NAMES}} is replaced with custom domains/IPs for gateway access
     server_name localhost {{ADDITIONAL_SERVER_NAMES}};
--- a/docker/nginx_rev_proxy_http_only.conf
+++ b/docker/nginx_rev_proxy_http_only.conf
@@ -166,6 +166,8 @@ map $host $rl_retry {
 {{VERSION_MAP}}
 # First server block now directly handles HTTP requests instead of redirecting
 server {
+    # Use relative redirects to avoid leaking internal scheme/port behind TLS-terminating proxies
+    absolute_redirect off;
     listen 8080;
     # {{ADDITIONAL_SERVER_NAMES}} is replaced with custom domains/IPs for gateway access
     server_name localhost {{ADDITIONAL_SERVER_NAMES}};


zsxh1990:
Hi team — friendly ping on this PR 🙂

CI is fully green. This fixes the trailing slash redirect issue that causes MCP clients to hang when using the no-slash URL (issue #1606).

The fix is minimal: adding  to both nginx templates so redirects use relative Location headers.

Happy to address any feedback!

zsxh1990:
Hi! Pre-commit passes. Let me know if you'd like any adjustments 🙂

aarora79:
Thank you for this fix and for the clear diagnosis of the absolute-redirect leak, along with the reproduction steps.

We've adopted your commit (cherry-picked with authorship preserved) in #1635 and extended it to also cover the `:8443` ssl server block. #1610 patched the `:8080` block, but a gateway that terminates TLS on `:8443` (for example the public deployment) still emitted absolute redirects and still hung on slash-less URLs, so the same directive was needed there too.

#1635 has merged to `main` and closes #1606. Verified on a live stack: slash-less redirects are now relative on both the `:8080` and `:8443` paths, following them resolves to the correct external origin with no hang, and MCP list/call flows continue to work through the gateway.

Closing this in favor of #1635 with thanks for the contribution.

PR #1613, opened 2026-08-07
feat(egress): support public OAuth clients (token_endpoint_auth_method=none) with PKCE

doppelc:
Adds `token_endpoint_auth_style: "none"` (RFC 7591 `token_endpoint_auth_method=none`) for the `custom` egress provider, so `oauth_user` egress can be configured against **public PKCE clients** — clients that have no client_secret by design, such as those minted by an MCP resource server's Dynamic Client Registration endpoint, e.g. Datadog MCP.

Previously the configure route required a `client_secret` unconditionally, making these providers impossible to configure.

## Design notes

- Only the `custom` provider can select the public style — every built-in provider is confidential, so nothing changes for existing configs.
- PKCE stays mandatory for custom providers; the S256 verifier is the proof of possession that replaces the secret (RFC 7636).
- No schema migration: `token_endpoint_auth_style` already existed on the provider config; this adds one enum value and the validation around it.

## Testing

- `uv run pytest tests/ -n 8`: 6754 passed, 74 skipped, 0 failures (coverage 65.85%).
- New unit tests: token-request shape for `none` (client_id only, no secret in body or headers), fail-closed confidential styles, secret resolution in the service layer, provider validation, configure route (client_id required for `none`, stored secret dropped on style switch, confidential path unchanged).
- Frontend: new ServerEditModal tests for the public-client fields (18/18 passing); tsc and eslint clean in touched files.

### DIFF
--- a/api/registry_client.py
+++ b/api/registry_client.py
@@ -3910,8 +3910,9 @@ class RegistryClient:
                 ``provider="custom"``.
             custom_scope_separator: Scope delimiter when the provider does not
                 use a space (custom only).
-            custom_token_auth_style: Where the client secret goes on the token
-                request -- "post_body" (default) or "basic_header" (custom only).
+            custom_token_auth_style: Token endpoint authentication style:
+                "post_body" (default), "basic_header", or "none" for a public
+                PKCE client (custom only).
             custom_resource: RFC 8707 resource indicator, sent on both the
                 authorize and token requests to bind the token to one protected
                 resource (custom only).
--- a/api/registry_management.py
+++ b/api/registry_management.py
@@ -6752,8 +6752,8 @@ Examples:
     )
     egress_configure_parser.add_argument(
         "--custom-token-auth-style",
-        choices=["post_body", "basic_header"],
-        help="Where the client secret goes on the token request (custom only, default post_body)",
+        choices=["post_body", "basic_header", "none"],
+        help="Token endpoint auth style; none selects a public PKCE client (custom only)",
     )
     egress_configure_parser.add_argument(
         "--custom-resource",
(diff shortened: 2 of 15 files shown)


aarora79:
Thank you for this contribution, @doppelc — this is a genuinely valuable addition.

Public OAuth clients (`token_endpoint_auth_method=none`) were a hard blocker, not an inconvenience: the configure route required a `client_secret` unconditionally, so any MCP server whose authorization server only issues public clients simply could not be onboarded for per-user egress. Datadog is exactly that case — its AS metadata advertises `token_endpoint_auth_methods_supported: ["none"]` with `pkce_required: true` — and there are more servers like it arriving as DCR-minted public clients become the norm for remote MCP servers. This unblocks that whole class.

We verified it end to end on a live deployment, not just in unit tests: registered the Datadog MCP server, configured egress with `custom_token_auth_style=none` plus the RFC 8707 resource indicator, and completed the 3LO consent flow — the connection comes back `active` and the gateway vends the user token with no client secret anywhere in the exchange.

Two things we particularly appreciated beyond the feature itself:

- Keeping PKCE structurally mandatory and gating the public style to the `custom` provider, so no built-in can ever be downgraded to secretless.
- Making the *confidential* path fail closed on a missing secret instead of sending an empty one, and dropping the stored secret when switching styles. Those are hardening wins that were not strictly required by the feature.

We are following up with a docs pass (the decision procedure for choosing `post_body` / `basic_header` / `none` from a provider’s published metadata), a FAQ page for configuring the Datadog MCP server, and a `cli/examples/` registration file. Thanks again.
Cc: @omrishiv

aarora79:
Added follow-up documentation for this feature in #1638: how to choose `custom_token_auth_style` from a provider's published metadata, plus a Datadog MCP configuration FAQ and a `cli/examples/` registration file.

Issue #1562, opened 2026-07-29
Bug: Health check hangs indefinitely on MCP servers using SSE/chunked transfer encoding for streamable-http transport


hochbit:
Hi, my postgres MCPg was never working through the gateway. this is what my Agent found:

# Quote (Claude 4.6)

## Describe the bug

The registry health checker marks MCP servers as unhealthy when those servers respond to `POST /mcp` with `Content-Type: text/event-stream` and `Transfer-Encoding: chunked` — which is valid and common behavior for streamable-http MCP servers (e.g. MCPg, supergateway-wrapped stdio servers).

The root cause is in `registry/health/service.py` in `_initialize_mcp_session`. It uses a plain `httpx` POST:

```python
response = await client.post(endpoint, headers=init_headers, json=initialize_payload, ...)
```

httpx buffers the full response body before returning. For SSE streams the body never closes, so the call blocks until `HEALTH_CHECK_TIMEOUT_SECONDS` is exceeded — at which point the server is marked unhealthy, even though the MCP protocol exchange completed successfully.

As a consequence, unhealthy servers have their nginx location block commented out, meaning the gateway returns **404** for any request to that path. The server is completely unreachable through the gateway while marked unhealthy.

---

## To Reproduce

1. Register any MCP server that responds to `POST /mcp` with `Content-Type: text/event-stream` (e.g. MCPg — PostgreSQL MCP server, or any supergateway-wrapped stdio server)
2. Configure bearer auth correctly in the server entry
3. Trigger a health check
4. Server is marked unhealthy despite being fully reachable and responding correctly

---

## Verification

The following curl confirms the server responds correctly at the protocol level:

```bash
# Step 1 — initialize
curl -X POST http://<proxy_pass_url>/mcp \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":"0","method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1.0"}}}'
# Returns 200 with Mcp-Session-Id header and SSE body

# Step 2 — ping with session ID
curl -X POST http://<proxy_pass_url>/mcp \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
[...]

PR #1614, opened 2026-08-09
fix(health): stream MCP initialize instead of buffering the full response

AmirF194:
`_initialize_mcp_session` (`registry/health/service.py`) uses a plain `client.post()` to send the MCP `initialize` request during a health check. httpx's `post()` waits for the response body to complete before returning. A streamable-http MCP server that keeps the connection open with `Content-Type: text/event-stream` / chunked framing after sending the initialize result never satisfies that: the call blocks until `HEALTH_CHECK_TIMEOUT_SECONDS` even though the status code and `Mcp-Session-Id` header, everything this function actually reads, arrived in the first chunk. The server is then marked unhealthy and nginx stops routing to it, despite responding correctly.

Fix: use `client.stream("POST", ...)` and read the status and headers without waiting for the body to close. The failure branch still calls `response.aread()` for the log line, since a non-streaming error response is finite and closes normally.

Related: #1562, filed against this exact function with a working curl repro. The reporter later closed it as resolved by an unrelated `SSRF_ALLOWED_CIDRS` fix on their end, so it is not something this PR closes, but the buffering mechanism their diagnosis identified is real and reproduces independently of their specific network config, verified below.

**Verification**
- Reproduced against the real function, not an extracted copy: a local asyncio server that answers with the session ID in the first chunk and then holds the connection open for 20s (matching a real streamable-http server). Unpatched `_initialize_mcp_session` on current `main` (`1b18a5a9`) times out at 5.02s and returns `None`; the same call against the patched file returns the server's session ID in 0.01s.
- `tests/unit/health/test_health_service.py` (84 tests, includes the 4 existing `_initialize_mcp_session` cases updated for the new call shape, plus a new one asserting the success path never calls `response.aread()`): 84 passed.
- `ruff check` / `ruff format --check` on both changed files: clean.
- Not verified: Bandit and the full `Test (Python 3.14)` CI job (MongoDB-backed, most of this repo's test suite pulls in unrelated services); ran the target test file directly instead.

### DIFF
--- a/registry/health/service.py
+++ b/registry/health/service.py
@@ -624,35 +624,38 @@ class HealthMonitoringService:
                 },
             }
 
-            response = await client.post(
+            # Streamed: a streamable-http server can hold the connection open (SSE/
+            # chunked) past the initialize response, and .post() would block reading
+            # the full body even though the headers we need already arrived.
+            async with client.stream(
+                "POST",
                 endpoint,
                 headers=init_headers,
                 json=initialize_payload,
                 timeout=httpx.Timeout(5.0),
                 follow_redirects=True,
-            )
+            ) as response:
+                # Check if initialize succeeded
+                if response.status_code not in [200, 201]:
+                    logger.warning(
+                        f"MCP initialize failed endpoint={redact_url(endpoint)} status={response.status_code}",
+                    )
+                    return None
 
-            # Check if initialize succeeded
-            if response.status_code not in [200, 201]:
-                logger.warning(
-                    f"MCP initialize failed endpoint={redact_url(endpoint)} status={response.status_code}",
+                # Get session ID from response headers (server-generated)
+                server_session_id = response.headers.get("Mcp-Session-Id") or response.headers.get(
+                    "mcp-session-id"
                 )
-                return None
-
-            # Get session ID from response headers (server-generated)
-            server_session_id = response.headers.get("Mcp-Session-Id") or response.headers.get(
-                "mcp-session-id"
-            )
-            if server_session_id:
-                logger.debug(f"Server returned session ID: {server_session_id}")
-                return server_session_id
-            else:
-                # If server doesn't return a session ID, generate one for stateless servers
-                client_session_id = str(uuid.uuid4())
-                logger.debug(
-                    f"Server did not return session ID, using client-generated: {client_session_id}"
-                )
-                return client_session_id
+                if server_session_id:
+                    logger.debug(f"Server returned session ID: {server_session_id}")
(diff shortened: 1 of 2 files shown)


aarora79:
thanks, will test and merge.

AmirF194:
Rebased onto main to resolve a conflict with the recent egress-hardening security PR (#1564), which reformatted the nearby error logging. Kept the streaming fix and adopted the newer redacted log format; full health test suite still green (94/94), ruff clean. Ready for another look whenever you get to it.

aarora79:
Thanks for this — the diagnosis is right and the `client.stream` technique is the correct fix. Against a server that flushes the session ID and then holds the stream open, your patch takes the initialize hop from a 5.02s timeout to 0.02s.

Two notes on what I did rather than sending it back to you.

**I pushed one commit to your branch** (2e88ca2f) migrating the stub in `test_health_logs_strip_query_and_response_body` to `client.stream`. It was still stubbing `client.post`, so after your change the mocked 500 was bypassed and `status_code` came back as an auto-generated `AsyncMock`, failing the `status=500` assertion. Worth knowing why it mattered beyond the red: that test is #1564's log-redaction guard, and with the mock bypassed its two "secret not in log" assertions were passing against an empty mock rather than against the real logging path. Migrating the stub puts the secret-bearing response back in front of the log call. `tests/unit/health/` is 100 green now.

**The ping hop needs the same treatment**, and I've opened #1645 for it rather than asking you for another round. The probe is two hops — `initialize` at `service.py:630` and the `ping` at `service.py:886` (plus `:998` on the default-transport path) — and a server that holds its stream open after `initialize` does the same after `ping`. So hop 2 kept buffering and the check still reported `unhealthy: ReadTimeout` via `service.py:934`. #1645 adds a small shared head-only probe used by both ping sites, which takes the ping from 5.02s to 0.01s, and a regression test that drives `_check_server_endpoint_transport_aware` so both hops are covered together.

That second part is the only thing your repro could not have surfaced — it drove `_initialize_mcp_session` directly, which was otherwise a more careful verification than most PRs arrive with, and the explicit "not verified" list was appreciated.

Merging this as-is now; #1645 lands on top. Thanks again.

aarora79:
see follow-up PR https://github.com/agentic-community/mcp-gateway-registry/pull/1645. Both will be included in 1.30.0 release.

AmirF194:
Good catch on the stub, migrating it to client.stream so the redaction assertions test the real path instead of an empty mock is the right fix. Thanks for catching the ping hop too, makes sense that it needed the same treatment.

PR #1618, opened 2026-08-10
feat(entra): Microsoft Entra v1 scope pass-through

### LINKED ISSUE #990: [Phase 2] Entra v1 api:// scope verbatim pass-through in PRM + audience normalization
# Sub-issue F: Entra v1 `api://<app-id>/<scope>` scope verbatim pass-through in PRM + audience normalization

**Parent:** #988
**Labels:** `enhancement`, `oauth`, `entra-id`, `mcp-spec`, `authentication`, `subtask`
**Phase:** 2
**Depends on:** A
**Blocks:** none (but coordinates with E)

---

## Summary

When the configured IdP is Microsoft Entra v1, the PRM `scopes_supported` array MUST contain scope strings in the exact `api://<app-id>/<scope-name>` form that Entra expects on the `/authorize` request. Anything else and Entra returns `AADSTS650053: The application ... asked for scope X that doesn't exist on the resource Y.` This issue tracks the verbatim pass-through logic in the PRM endpoint plus the dual-audience acceptance on token validation.

## Background

Entra v1 (the `api://...` scope form) requires the client to request scopes as `api://<app-id>/<scope>` (or `api://<app-uri>/<scope>`), not just the bare scope name. Unlike Entra v2 which accepts fragment-only scopes, v1 rejects the request if the resource prefix is missing.

Matching behavior on the token side: the `aud` claim for an Entra v1-issued token can be either the bare GUID (`<app-id>`) or the URI form (`api://<app-id>`). The Registry's audience validator must accept both as equivalent.

This is a known failure mode both for our own Registry deployment (observed in field deployments using Entra v1) and for the AgentCore Identity integration. Fixing it here unblocks both.

Existing code at [auth_server/providers/entra.py:149](auth_server/providers/entra.py#L149) already does dual-audience acceptance on the inbound path — this issue ensures the PRM emitted by sub-issue A mirrors that behavior on the outbound advertisement path so Claude Code / Cursor / VS Code send the right scope strings in the first place.

### In scope

- **PRM `scopes_supported` emission for Entra v1**:
  - Read the registered MCP server scopes from the gateway's registration source.
  - If provider is Entra and the configured scope format is v1, emit each scope as `api://<app-id-or-uri>/<scope-name>` verbatim. No URL encoding, no transformation.
  - The `<app-id-or-uri>` prefix must match the value that Entra expects on `/authorize`; read from provider config, not reconstructed.
- **PRM `scopes_supported` emission for Entra v2**:
  - Emit bare scope names (Entra v2 accepts fragment form).
- **PRM `scopes_supported` emission for non-Entra providers**:
  - Emit whatever the provider declares. Keycloak, Okta, Cognito, Auth0 have their own canonical scope forms; do not transform.
- **Audience normalization on token validation** (coordinates with sub-issue E):
  - Accept both `api://<app-id>` and bare `<app-id>` GUID as equivalent audience values for Entra v1 tokens.
  - Formalize the logic currently at [auth_server/providers/entra.py:149](auth_server/providers/entra.py#L149) as a provider-level method, not ad-hoc code.
- **Regression tests**:
  - Entra v1 provider config → PRM emits `api://<app-id>/mcp.read` form.
  - Entra v2 provider config → PRM emits `mcp.read` form.
  - Keycloak / Okta / Cognito / Auth0 → PRM emits whatever the provider declares.
  - Audience validator accepts both Entra v1 forms.
  - Regression test for the AADSTS650053 failure mode: mock an Entra v1 `/authorize` interaction and assert the scope strings match.

### Out of scope

- General multi-IdP configuration (→ #895, #896).
- CIMD publisher / consumer (→ B, C).
- Fixing IdP-side config; operators own their Entra app registration.

## Design notes

- Put the scope-format logic in the provider module ([auth_server/providers/entra.py](auth_server/providers/entra.py)). The discovery endpoint from sub-issue A calls `provider.protected_resource_metadata()`; that provider method owns the formatting.
- Config surface: add an explicit `entra_scope_format: v1 | v2` setting (default `v2`), with clear docs on when to use which. Don't auto-detect; operators need a predictable behavior.
- Dual-audience acceptance is asymmetric: we accept either form inbound, we emit one form outbound (the one Entra expects). Don't conflate.

## Acceptance criteria

- [ ] PRM endpoint emits scope strings in the format the configured Entra version expects, verbatim, no encoding.
- [ ] Non-Entra providers unchanged.
- [ ] Audience validator on `/oauth/token` and MCP-data-plane paths accepts both `api://<app-id>` and bare `<app-id>` forms for Entra v1.
- [ ] Config surface documented: `entra_scope_format: v1 | v2`.
- [ ] Regression tests for each provider.
- [ ] Regression test covering the AADSTS650053 failure mode.
- [ ] Docs page in [docs/](docs/) addressing: "if you're on Entra v1, set this flag; if you're on Entra v2, don't."
- [ ] Stakeholder update: notify owners of the AgentCore Identity integration and any field deployments using Entra v1 that this is scoped and tracked.

## Risks and open questions

- **Version detection.** Can we auto-detect Entra v1 vs v2 from the issuer URL? Yes (Entra v1 issuer contains `/v1.0/`; v2 contains `/v2.0/`), but auto-detection hides operator intent. Stick with explicit config.
- **App-uri vs app-id form.** Entra allows either `api://<app-id>` or a custom `api://<tenant-specific-uri>`. Provider config must capture whichever the operator set up in the Entra app registration.

## References

- [Microsoft Entra v1 vs v2 endpoints](https://learn.microsoft.com/en-us/entra/identity-platform/azure-ad-endpoint-comparison)
- [AADSTS650053 error reference](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes)
- [auth_server/providers/entra.py:149](auth_server/providers/entra.py#L149) (dual-audience pattern)
- Recommendation doc §6 Phase 2: [.scratchpad/coding-assistant-oauth/recommendation-2026-05-04.md](.scratchpad/coding-assistant-oauth/recommendation-2026-05-04.md)
- Related (but distinct) threads: Entra v1 scope format interop, Entra PKCE interop, AgentCore Identity verbatim scope requirement

omrishiv:
Addresses #990.

### 1. v1 `api://` scope verbatim pass-through in PRM (e5da0927)
Under Entra v1, custom resource scopes must be requested on `/authorize` as
`api://<app-id>/<scope>` or Entra rejects the login with **AADSTS650053**. The gateway
now advertises the correct scope form in its OAuth Protected Resource Metadata.

- `EntraIdProvider` gains an explicit `scope_format` (`v1`|`v2`, default `v2`) and
  `application_id_uri`, wired from `ENTRA_SCOPE_FORMAT` / `ENTRA_APPLICATION_ID_URI`.
- `format_advertised_scopes()` rewrites custom scopes to `api://<app-id-or-uri>/<scope>`
  under v1, leaves them bare under v2, and always leaves standard OIDC scopes and
  already-URI-qualified scopes untouched.
- Formalized ad-hoc dual-audience acceptance into a closed `accepted_audiences()`
  allowlist (bare GUID, `api://<app-id>`, configured Application ID URI, per-server OBO
  resources). `verify_aud` stays enforced.

### 2. Per-server PRM for all servers on Entra (00c98662)
On Entra, **every** server — including plain servers (`egress_auth_mode` none/unset),
not just `obo_exchange` / `oauth_user` — now advertises a per-server RFC 9728 PRM whose
`resource` is its exact connection URL.

Why: the gateway-wide PRM advertises the bare origin, which a spec-compliant client
canonicalizes with a trailing slash (RFC 9728 §3.3); Entra App ID URIs cannot end in a
slash, so the exchange fails with **AADSTS9010010**. The per-server connection URL is
the only value that satisfies both the client and Entra. Verified end-to-end against a
live Entra tenant.

Trade-off: operators on Entra must register each connection URL as an App ID URI with a
`user_impersonation` scope (listed by `GET /api/egress/obo-identifier-uris`).

### 3. Access-token hardening & egress-off fixes (97fb1227)
- **Egress-off Entra plain-server login:** the per-server ingress audience
  was gated on `egress_auth_enabled` (default off), so tokens audienced to
  `https://gw/<server>/mcp` matched nothing → 401. Now gated on
  `_server_advertises_per_server_prm()` (`auth_provider==entra` OR egress enabled),
  mirroring the registry. Path-bound, so widening cannot reach a different server.
- **Audience-URL agreement trap:** build a per-server resource for each distinct
  configured base URL (external, `registry_url`, `REGISTRY_URL`) and accept any.
- **id_token/access_token confusion (security):** `validate_token` accepted an Entra
  `id_token` as a bearer. Added `_reject_non_access_token()` (post-signature), rejecting
  tokens with id_token-only claims (`nonce`/`at_hash`/`c_hash`). Does not require
  `scp`/`roles` to avoid false-rejecting roleless client-credentials M2M tokens.
- Per-server PRM route returns 502 (not an unhandled 500 + traceback) on lookup failure;
  `GET /api/egress/obo-identifier-uris` returns `[]` instead of 404 when egress is off.

## Config surface
New `ENTRA_SCOPE_FORMAT` / `ENTRA_APPLICATION_ID_URI` wired across `.env.example`, Helm
(auth-server / registry / stack + reserved-env lists), Terraform, and CDK.

## Testing
- `test_entra.py`: scope formatting, audience normalization, AADSTS650053 regression,
  `_reject_non_access_token` cases (id_token rejected; delegated / app-role / roleless-M2M
  accepted).
- `test_obo_extra_audiences.py`: egress-off Entra acceptance, multi-base-URL, path-binding.
- `test_wellknown_routes.py`: per-server PRM for Entra (all modes), lenient-IdP regression
  guard, v1/v2 global PRM, 502-on-lookup-failure.
- Helm reserved-name coverage (`entra_scope_format_test.yaml`, `extra_env_test.yaml`).


aarora79:
Merged, thanks! Added a small follow-up in #1622 addressing the two review Should-Fix items: a shared source of truth for the Entra per-server-PRM predicate (so the registry advertise-side and auth-server accept-side can't drift) and surfacing `entra_scope_format` / `entra_application_id_uri` in the System Config groups.

Issue #1501, opened 2026-07-16
bug: MCP server nginx location blocks without trailing slash silently hijack unrelated browser routes


zoltan-fedor:
## Summary

MCP server nginx location blocks are generated without a trailing slash (`location /path`), which causes nginx to prefix-match them against **any request whose URL starts with that string**. A server registered at a short
path such as `/a` silently intercepts requests to `/api/anything`, `/auth`, or any other route beginning with `/a` — including browser-facing SPA routes — and subjects them to `auth_request /validate`. This produces `{"error": "Authentication required"}` for unauthenticated users navigating to those routes.

## Steps to reproduce

1. Register an MCP server with a short path, e.g. `/a`.
2. Sign out of the application so the session cookie is cleared.
3. Observe that any browser navigation to a URL starting with `/a` (e.g.   `/api/auth/login`, or any SPA route whose path begins with `a`) returns `{"error": "Authentication required"}` instead of the expected page.

In practice this is difficult to spot because:

- The affected routes may work fine while a valid session cookie is present   (auth_request returns 200 and the request passes through normally).
- Only unauthenticated requests — such as post-logout redirects — expose the   mis-routing.
- Short server path names (`/a`, `/b`, `/l`, `/lo`, etc.) are the most dangerous but any path that is a prefix of another legitimate route is  affected.

## Root cause

`nginx_service.py::_create_location_block` generates location blocks using the server path with no trailing slash:

```python
# registry/core/nginx_service.py
location_path = path   # e.g. "/a"
return f"""location {{ROOT_PATH}}{location_path} {{ ... }}"""
```

In nginx, `location /a` is a prefix match that matches **every URL that starts with `/a`** — including `/api/`, `/auth/`, `/about`, etc. — not just paths under an `/a/` subtree.

Because these generated blocks contain `auth_request /validate`, any request that matches and carries no valid session token receives a 401 that nginx converts to `{"error": "Authentication required"}`.

The `^~` (preferential prefix) guards on `/api/auth`, `/api/health`, `/oauth2/`, etc. protect those specific paths because they are longer matches, but any other route that shares a prefix with a registered server name is
unprotected.

## Expected behaviour

A server registered at `/a` should only own routes under `/a/` — it must not interfere with `/api/`, `/auth`, or any other path that merely starts with the same letter.

## Proposed fix

Always normalise the nginx location path to end with a trailing slash in `_create_location_block`:

```python
# registry/core/nginx_service.py
[...]

PR #1620, opened 2026-08-11
fix(nginx): send relative redirects from the front-door server blocks (#1606)

### LINKED ISSUE #1606: bug: MCP server endpoint without trailing slash returns a 301 to an internal `http://<host>:8080/...` origin, causing MCP clients to hang/time out

rahul188:
Fixes #1606.

## Root cause

#1501 / PR #1507 made the generated location blocks trailing-slash terminated (`location /<server>/`) so a bare prefix could not hijack unrelated routes. A correct fix — but it means a request to a registered MCP server *without* the trailing slash now falls through to nginx's automatic "add the slash" **301**.

`absolute_redirect` is on by default, so nginx rebuilds that `Location` from the listener that answered:

```
Location: http://<gateway-host>:8080/registry/<server>/
```

Two things are wrong with it, and both come from the same place — the internal listener is not the origin the client dialed:

- **`http://`** — TLS terminates upstream (Istio/ALB), so the scheme is wrong
- **`:8080`** — the internal container port, not routable from outside the cluster

MCP clients follow the 301 and hang until they time out. Net effect: the endpoint works only when the caller already includes the trailing slash — which the registry UI's connect dialog does not do.

Confirmed on current `main` (`6d2e6ed4`): `grep -rn absolute_redirect registry/ docker/ charts/` returns nothing, so the nginx default is in force everywhere.

## The fix

`absolute_redirect off;` in each front-door `server` block — both blocks in `nginx_rev_proxy_http_and_https.conf` (`:8080` and `:8443 ssl`) and the one in `nginx_rev_proxy_http_only.conf`.

nginx then emits a **relative** `Location: /registry/<server>/`, which the client resolves against the origin it actually dialed. That is correct no matter how many proxies sit in front and needs no knowledge of the external scheme, host, or port — unlike `port_in_redirect off` or a `server_name_in_redirect` scheme, which would still have to guess the external scheme.

Set per server block rather than once at `http` scope, because what leaks is the answering listener's own scheme and port; scoping it to the listeners makes that relationship explicit.

## Testing

New `tests/unit/core/test_nginx_absolute_redirect.py`, 4 tests:

- `test_every_server_block_disables_absolute_redirect` (parametrized over both templates) — brace-counts the top-level `server { ... }` blocks and asserts each sets `absolute_redirect off`. It reports the offending block's `listen` line, so a future server block added without it fails with a message naming which one.
- `test_absolute_redirect_is_never_re_enabled` (both templates) — guards against a later `absolute_redirect on` silently undoing this in a narrower scope.

**Verified failing on unfixed `main`**: both parametrized cases of the first test fail; the two guard tests pass either way, as intended.

`tests/unit/core/`: **673 passed, 10 failed** — the same 10 (`test_telemetry.py`, `test_config.py`) fail on a clean `main` checkout, unchanged by this PR.

`black --check` clean on the new test file.

One limitation worth flagging: I could not run `nginx -t` against the result, since these are templates containing `{{ROOT_PATH}}` / `{{ADDITIONAL_SERVER_NAMES}}` placeholders and there is no nginx binary in the test environment. `absolute_redirect` is a plain `ngx_http_core_module` directive valid at `http`, `server`, and `location` scope, so it is placement-safe, but a container smoke test on your side would be worth having before merge.

### DIFF
--- a/docker/nginx_rev_proxy_http_and_https.conf
+++ b/docker/nginx_rev_proxy_http_and_https.conf
@@ -169,6 +169,14 @@ server {
     # {{ADDITIONAL_SERVER_NAMES}} is replaced with custom domains/IPs for gateway access
     server_name localhost {{ADDITIONAL_SERVER_NAMES}};
 
+    # nginx answers a directory-style request that is missing its trailing slash
+    # with an automatic 301. With absolute_redirect on (the nginx default) that
+    # Location is rebuilt from this listener -- http:// and the internal 8080/8443
+    # container port -- which is not reachable from outside, and TLS terminates
+    # upstream anyway. MCP clients follow the redirect and hang until they time
+    # out. Emitting a relative Location keeps the client on the origin it dialed.
+    absolute_redirect off;
+
     # NOTE: do NOT add a server-scope `set $backend_url "";` default here.
     # nginx auth_request creates the /validate subrequest via
     # ngx_http_subrequest(), which sets sr->variables = r->variables -- the
@@ -1086,6 +1094,14 @@ server {
     listen 8443 ssl;
     # {{ADDITIONAL_SERVER_NAMES}} is replaced with custom domains/IPs for gateway access
     server_name localhost {{ADDITIONAL_SERVER_NAMES}};
+
+    # nginx answers a directory-style request that is missing its trailing slash
+    # with an automatic 301. With absolute_redirect on (the nginx default) that
+    # Location is rebuilt from this listener -- http:// and the internal 8080/8443
+    # container port -- which is not reachable from outside, and TLS terminates
+    # upstream anyway. MCP clients follow the redirect and hang until they time
+    # out. Emitting a relative Location keeps the client on the origin it dialed.
+    absolute_redirect off;
 
     # SSL Configuration - requires user-provided certificates
     # Mount certificates to /etc/ssl/certs/fullchain.pem and /etc/ssl/private/privkey.pem
--- a/docker/nginx_rev_proxy_http_only.conf
+++ b/docker/nginx_rev_proxy_http_only.conf
@@ -170,6 +170,14 @@ server {
     # {{ADDITIONAL_SERVER_NAMES}} is replaced with custom domains/IPs for gateway access
     server_name localhost {{ADDITIONAL_SERVER_NAMES}};
 
+    # nginx answers a directory-style request that is missing its trailing slash
+    # with an automatic 301. With absolute_redirect on (the nginx default) that
+    # Location is rebuilt from this listener -- http:// and the internal 8080/8443
+    # container port -- which is not reachable from outside, and TLS terminates
+    # upstream anyway. MCP clients follow the redirect and hang until they time
+    # out. Emitting a relative Location keeps the client on the origin it dialed.
+    absolute_redirect off;
+
     # NOTE: do NOT add a server-scope `set $backend_url "";` default here.
     # nginx auth_request creates the /validate subrequest via
     # ngx_http_subrequest(), which sets sr->variables = r->variables -- the
--- a/tests/unit/core/test_nginx_absolute_redirect.py
+++ b/tests/unit/core/test_nginx_absolute_redirect.py
@@ -1,0 +1,71 @@
+"""The front-door nginx templates must not emit absolute redirects.
+
+Issue #1606: the #1501 fix made generated location blocks trailing-slash
+terminated, so a request to a registered MCP server *without* the trailing
+slash now hits nginx's automatic "add the slash" 301. With ``absolute_redirect``
+at its nginx default (on), that Location is rebuilt from the listener that
+served the request -- ``http://`` and the internal container port 8080/8443 --
+which is not routable from outside the cluster, and TLS terminates upstream
+anyway. MCP clients follow the redirect and hang until they time out.
+
+``absolute_redirect off`` makes nginx send a relative Location, so the client
+stays on the origin it dialed. It has to be set in every server block: the
+internal port is baked into the redirect by whichever listener answered.
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+
+_DOCKER_DIR = Path(__file__).resolve().parents[3] / "docker"
+_TEMPLATES = (
+    "nginx_rev_proxy_http_and_https.conf",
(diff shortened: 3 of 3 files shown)


aarora79:
thanks, will test and merge.

aarora79:
Thanks for this, @rahul188 — the diagnosis is spot-on and the write-up is excellent.

This one has been overtaken, though. The same fix already landed in #1635 (merged Aug 15), which added `absolute_redirect off;` to all three front-door server blocks — both `:8080` blocks and the `:8443` ssl block — and closed #1606. #1635 adopted @zsxh1990's #1610 and completed it with the `:8443` coverage. Your PR predates that merge (it references `main` at `6d2e6ed4`), so this is no fault of yours — the config change is already on `main`.

We're going to close this rather than merge it, for two reasons:

1. **The config hunks are now redundant** — `absolute_redirect off;` is already present in all three blocks on `main`.
2. **Merging as-is would break nginx.** GitHub still shows the PR as mergeable, but that check is purely textual. A clean merge would insert a *second* `absolute_redirect off;` into each block, and nginx rejects a duplicate flag directive at `nginx -t` (`"absolute_redirect" directive is duplicate`), so the gateway wouldn't start.

That said, your tests were the one thing more complete than what merged — **#1635 shipped without any tests.** So we've kept them: I opened #1669 with your `test_nginx_absolute_redirect.py`, git authorship preserved to you, dropping only the redundant config edits. All 4 tests pass against current `main`, so your coverage lands and now guards the fix against regression.

Closing in favor of #1635 (config, already merged) + #1669 (your tests). Thanks again for the careful work.

PR #1627, opened 2026-08-13
fix(nginx,lua): forward caller identity headers in virtual server backend locations

doncouture:
…kend locations

Virtual server backends received no caller identity because:
1. auth_request_set variables ($auth_user, $auth_username) are scoped to the parent request and do NOT propagate into ngx.location.capture subrequests via proxy_set_header.
2. The /_vs_backend_* internal locations had no mechanism to receive identity from the Lua router.

Fix (two-part):
- virtual_router.lua: set X-User and X-Username as request headers (ngx.req.set_header) before any ngx.location.capture call. Subrequests inherit these headers from the parent.
- nginx_service.py: add proxy_set_header X-User $http_x_user and proxy_set_header X-Username $http_x_username to the _vs_backend location template so they are forwarded to the upstream MCP server.

This matches the behaviour of direct (non-virtual) server paths, enabling backends behind virtual servers to enforce write-gates and record audit attribution.

Tested: Confluence page update through /virtual/development/mcp with identity correctly resolved by the backend (mcp-atlassian).

*Issue #, if available:*
1626

*Description of changes:*
Forwards X-User/X-Username identity headers through virtual server backend locations to match direct path behaviour.

### DIFF
--- a/docker/lua/virtual_router.lua
+++ b/docker/lua/virtual_router.lua
@@ -947,6 +947,21 @@ function _M.route()
     -- all ngx.location.capture subrequests to backends inherit it.
     ngx.req.set_header("Accept", "application/json, text/event-stream")
 
+    -- Forward the validated caller identity to backend subrequests.
+    -- auth_request_set variables ($auth_user, $auth_username) are scoped to the
+    -- parent request and do NOT propagate into ngx.location.capture subrequests
+    -- via proxy_set_header. Setting them as request headers here ensures the
+    -- _vs_backend_* locations pass them through to the upstream MCP server,
+    -- enabling backends to enforce write-gates and record audit attribution.
+    local auth_user = ngx.var.auth_user
+    local auth_username = ngx.var.auth_username
+    if auth_user and auth_user ~= "" then
+        ngx.req.set_header("X-User", auth_user)
+    end
+    if auth_username and auth_username ~= "" then
+        ngx.req.set_header("X-Username", auth_username)
+    end
+
     local request_method = ngx.var.request_method
 
     -- Handle HTTP GET: per MCP Streamable HTTP 2025-11-25 spec section 3.3,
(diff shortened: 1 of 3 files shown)


aarora79:
Thanks for this, @doncouture — virtual MCP servers are an important piece of the platform, and wiring caller identity through to their backends is exactly the parity gap that needed closing. The diagnosis of why `auth_request_set` vars don't survive the `ngx.location.capture` subrequests is spot-on, and the happy-path fix is clean.

We're going to merge this to get the functionality in, and there are a couple of hardening items we'll pick up ourselves in a quick follow-up so you don't have to round-trip:

1. **Fail-closed identity headers.** Because the backend value is forwarded from `$http_x_user` (the client-supplied request header) and the Lua only overwrites it when `auth_user` is non-empty, a caller with a userless token (M2M / client-credentials, where `$auth_user` is empty) could send a spoofed `X-User` that survives to the backend. We'll add an `else ngx.req.clear_header("X-User")` / `X-Username` so the header is always the validated value or nothing — matching the unconditional behavior of the direct-server path.
2. **Test coverage for that path** — a Lua-level case (empty `auth_user` -> header cleared) in `tests/lua/test_virtual_router.lua`, plus an end-to-end spoofing check in `tests/e2e/`.
3. **Follow-up parity pass** on the other identity headers (`X-Client-Id`, `X-Scopes`, `X-Groups`) so the virtual path stays aligned with the direct path.

None of this takes away from the fix — thank you for driving it. We'll link the follow-up PR here once it's up.

aarora79:
Follow-up PR with the fail-closed fix + Lua regression tests is up: #1672. Blocker (identity spoofing) and its test coverage are addressed there; the end-to-end spoof test and the `X-Client-Id`/`X-Scopes`/`X-Groups` parity pass are tracked as the remaining items in that PR description.

Issue #1565, opened 2026-07-30
Securely proxy arbitrary registered resources through the gateway


omrishiv:
## Summary

Add an opt-in, secure gateway proxy for arbitrary registered resources—including skills, agents, and custom entities—using the registry as the control plane and nginx/auth-server as the data plane.

This allows clients to invoke eligible resources through one governed gateway endpoint instead of connecting directly to each upstream. The gateway applies authentication, authorization, auditability, SSRF defenses, transport limits, and optional upstream credentials consistently across resource types.

## Goals

- Let supported registered entities opt into gateway proxying.
- Generate stable gateway client URLs while keeping real upstream URLs private.
- Authorize each request against its entity type and registered path.
- Support ordinary request/response traffic and opt-in streaming.
- Provide secure static upstream-header vending and controlled caller overrides.
- Apply layered SSRF defenses before and during outbound connections.
- Support Docker Compose, Helm/Kubernetes, and Terraform/ECS deployments.
- Keep the feature disabled by default until deployment/network controls are configured.

### Entity and control-plane support

- Add common proxy settings to proxy-capable entity models.
- Validate proxy configuration during registration and updates.
- Persist and list proxied entities efficiently.
- Ensure private proxy configuration does not cross federation boundaries.
- Derive gateway-facing client URLs without exposing upstream endpoints.

### Gateway runtime and authorization

- Generate nginx routes for each eligible proxied entity.
- Use the auth-server validation hop for caller authorization.
- Mint a short-lived internal token bound to the authorized proxy route.
- Keep generic-proxy internal tokens separate from existing MCP proxy tokens.
- Enforce request size, timeout, concurrency, and method policies.
- Fail closed when route, token, or feature-latch validation fails.

### SSRF and network security

- Validate and resolve upstream hosts during registration.
- Use the canonical `url_guard` IP classifier and egress policy.
- Deny cloud metadata, loopback, link-local, multicast, reserved, and other unsafe targets.
- Use pinned outbound transport so the validated address is the address contacted.
- Preserve TLS hostname verification while connecting to pinned addresses.
- Perform startup egress self-checks.
- Provide opt-in Kubernetes egress NetworkPolicy controls.
- Keep private-target access explicit and default-denied.

### Scopes and UI

- Add proxy-resource scope editing to IAM administration.
- Add proxy toggles and fields to entity forms.
- Show proxied status on resource cards.
- Display the generated gateway client URL.
[...]

PR #1628, opened 2026-08-13
Gateway proxy for any registry resource - foundation (PR 1 of 1565)

omrishiv:
*Issue #, if available:*

#1565 PR 1

## Summary

First slice of "proxy any registry entity through the gateway" (#1565). Today only MCP servers and A2A agents can be fronted by the gateway; this stack extends gateway proxying to skills, agents, and custom entities through a uniform generic-proxy hop, with SSRF defense in depth at every layer.

This PR lands the **foundation and the full generic-proxy path end to end**, but ships **dark**: `gateway_generic_proxy_enabled` and `gateway_canonical_namespace_enabled` both default to `False`, and `gateway_proxy_allow_private_targets` defaults to `False`. With the feature off, no nginx location block is rendered, `/validate` never mints a generic token, and the render path issues **zero** additional per-tick DB queries. It is a no-op in production until a later slice flips the flags.

## What's included

**Model layer — opt-in proxying**
- New `registry/schemas/proxy_mixin.py`: `ProxyableMixin` (`is_proxied`, `proxy_target_url`, plus refresh bookkeeping fields), an edge-only egress-guard validator, and `resolve_proxy_target()`.
- `ProxyableMixin` wired into the five storage models and their request/patch models. Virtual servers are alias-only (`proxy_target_url` rejected).
- Load-bearing split: **storage models never raise** on a stored `proxy_target_url` (they are reconstructed on every read; raising would vanish a bypass-written record from every listing). The egress raise and the target-required raise live on the request/patch models (the API edge, no vanish risk).

**Repository layer**
- `list_proxied()` added to the five entity repositories (DocumentDB overrides + non-abstract default `[]`), issuing an indexed `{"is_proxied": True}` query that projects only what render + `resolve_proxy_target` need, and fails closed to `[]` on any error.
- Shared `ensure_is_proxied_index` helper: prefers a partial index, falls back to a plain index when DocumentDB rejects `partialFilterExpression`, never raises.

**nginx render**
- Render-time generation of generic-proxy location blocks routing proxied non-MCP entities through the auth-server hop at `/proxy/{entity_type}/{path}/`.
- `_safe_generic_block`: a skip-not-fail render guard so one malformed or SSRF-denied row cannot fail `nginx -t` for the whole replica. Guards scheme, egress policy, entity-type token grammar, `..` path segments, and directive-breakout characters.
- Cross-placeholder location-collision dedup with generic as the lowest-precedence tier.

**auth-server generic hop**
- `mint_generic_proxy_token` / `verify_generic_proxy_token`: a short-lived token with an audience distinct from every other internal audience, binding both `entity_type` and the full registered path on a segment boundary, read under a distinct header (`X-Internal-Token-Generic`) so the two hops' tokens can never collide.
- `/proxy/{entity_type}/{entity_path:path}` handler: a uniform buffering proxy (no JSON-RPC parsing). Destination is cryptographically pinned from the token; inbound `X-Upstream-Url` is ignored. Sub-path confinement + post-join host/port pinning; `follow_redirects=False`; a response-header allowlist that still drops `Set-Cookie`/HSTS/CSP/framing while the gateway sets its own security headers; concurrency semaphore and body-size bound.
- `/validate` wiring: discriminates the generic hop, maps HTTP verbs to scope methods, and refuses a state-changing verb under ambient cookie auth before any scope check (Bearer callers exempt; gated by `gateway_generic_require_bearer_for_writes`, default true). The MCP `methods:["*"]` wildcard does **not** authorize an HTTP verb.

**SSRF — defense in depth**
- Registration/update-time DNS resolve-and-validate: at the service layer (single choke point per entity type, so batch/federation paths inherit it), the effective target's hostname is resolved and **every** resolved IP is validated against the egress policy, rejecting a metadata/private/link-local target with a clean 4xx before persist.
- Convergence onto the canonical `url_guard`: the generic hop fetches with `guarded_async_client` (connect-to-pinned-IP + per-target TLS, re-validated per redirect) instead of a bare `httpx` client. `url_guard` gains `allow_private` plumbing on the proxy allowlist (honoring `gateway_proxy_allow_private_targets`); the cloud-metadata denylist (including Alibaba `100.100.100.200`) remains never-overridable. There is no separate `ip_guard` module — all IP classification is the single canonical `url_guard` path.
- Startup egress self-check: probes the cloud metadata IPs; if either is reachable the network egress policy is not enforced, so it logs CRITICAL, emits `gateway_egress_policy_unverified=1`, and disables the feature for the process (fail-closed latch) without failing pod readiness.

**Federation isolation**
- Proxy config never crosses the federation boundary in either direction: `strip_proxy_fields` removes the proxy fields on every peer-content ingest and on the federation export chokepoint (key-drop, not force-false, so a re-sync can't clobber a local admin's opt-in). `resolve_proxy_target` returns `None` for a federated entity as an absolute guard behind the ingest strip.

**Observability**
- `mcpgw_registry_gateway_generic_blocks_dropped_total{reason=invalid|collision}` counter and a `gateway_egress_policy_unverified` gauge.

**Config**
- New settings (all fail-closed): `gateway_generic_proxy_enabled` (false), `gateway_canonical_namespace_enabled` (false), `gateway_proxy_allow_private_targets` (false), `gateway_generic_require_bearer_for_writes` (true), plus body-size / concurrency / TLS-verify knobs and a strict nginx size-token validator on the client-body-size string (a config-injection surface).

## Safety / rollout

- Ships dark: both feature flags default off; private-target relaxation off. No-op in production until a later slice enables it.
- Zero new per-tick DB queries while disabled (verified by test).
- Fail-closed throughout: render guard skips bad rows, `list_proxied` fails to `[]`, startup self-check disables the feature if egress isn't enforced, `/proxy` handler 503s until the latch is confirmed.

## Testing

- 16 new test files (`tests/auth_server/unit/`, `tests/unit/{schemas,core,services,repositories,auth,utils}/`) covering: model edge validation + storage read-safety, `list_proxied` projection + fail-closed, token mint/verify (audience isolation, cross-type/sibling-path/dot-dot rejection, sub-path acceptance), `/validate` verb authorization + privilege-escalation guard + CSRF, nginx block generation + collision dedup + marker-forward template invariants, the hard-gate composite (no denied target renders a live route), DNS resolve-and-validate on each service path, and federation isolation.
- Full unit suite green.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -733,7 +733,7 @@ working in those areas.
 - **Use the canonical helper, never reinvent/copy-paste** (see the "Canonical
   helpers" section in the guidelines doc): outbound HTTP from user/registry URLs →
   `registry/utils/url_guard.py` `guarded_client`/`guarded_async_client` (never a
-  bare `httpx` client), IP categories/unwrapping → `registry/utils/ip_guard.py`,
+  bare `httpx` client), IP categories/unwrapping → `registry/utils/url_guard.py`,
   URL identity/logging → `normalize_url_identity`/`sanitized_url_for_log`; CSRF →
   `registry/auth/csrf.py`; internal service auth →
   `registry/auth/internal.py`; signing-secret validation →
--- a/auth_server/internal_request_token.py
+++ b/auth_server/internal_request_token.py
@@ -34,6 +34,13 @@ MCP_PROXY_TOKEN_USE: str = "mcp-proxy"
 MCP_REGISTRY_UI_AUDIENCE: str = "mcp-registry-ui"
 MCP_REGISTRY_UI_TOKEN_USE: str = "mcp-registry-ui"
 
+# Audience for the generic reverse-proxy hop (/proxy/{entity_type}/{path}), used
+# for non-MCP proxied entities (skills, agents, custom). Distinct from every
+# audience above so a generic-proxy token can't be replayed at /mcp-proxy (which
+# has no tool-list filtering) or the registry API, or vice-versa.
+GENERIC_PROXY_AUDIENCE: str = "generic-proxy"
+GENERIC_PROXY_TOKEN_USE: str = "generic-proxy"
+
 # TTL is clamped to this floor so a misconfigured TTL of 0/negative cannot
 # combine with the leeway into a confusing always-valid window.
 _MIN_TTL_SECONDS: int = 5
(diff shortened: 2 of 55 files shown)


aarora79:
Follow-up #1698 (merged) addresses the before-enabling items from review:

- **Internal-header leak (security):** the generic hop now strips the gateway-internal identity/token/marker headers (`X-Internal-Token-Generic`, `X-User`, `X-Scopes`, `X-Original-URL`, `X-Generic-Proxy-Kind`, `X-Entity-Path`) before forwarding to upstream backends.
- **Location prefix-match (#1501 class):** the generic `location` is now trailing-slash-normalized, so `/skill/foo` no longer prefix-matches `/skill/foobar`.
- **Config docs:** the 9 new `gateway_*` params are documented in `.env.example`, `unified-parameter-reference.md`, and the System Config UI (Terraform/Helm wiring deferred to the enabling slice).

Still tracked for later: full Terraform/Helm parameter wiring and the nice-to-haves (self-check metadata coverage, generic-token `jti`, `Content-Encoding` handling, `VirtualServerConfig` read-safety).

PR #1637, opened 2026-08-15
feat: optional metadata-field projection for search and list APIs

### LINKED ISSUE #1277: feat: optional metadata-field projection (source/select) for search and list APIs
## Problem Statement

Customers can store large, deeply-nested JSON blobs in the `metadata` field of a server, agent, or skill card. Today every read path returns the **entire** `metadata` object, and not all of it is needed all the time. On list and search responses (which return many assets at once) this inflates payloads, increases latency, and ships data the caller never uses.

There is currently no way for a caller to ask for only a subset of `metadata`. The only field-selection control that exists anywhere in the API is `include_tools` on `GET /servers` (an implicit `{"tool_list": 0}` projection) - there is no general `fields` / `projection` / `source` parameter.

## Proposed Solution

Add an **optional metadata projection** to the read APIs: a query parameter that lets the caller specify an allowlist of `metadata` subfields (including nested fields via dot-notation). When supplied, only those metadata fields are returned; the rest of the `metadata` object is omitted. Everything **outside** `metadata` (the rest of the server/agent/skill card) is returned exactly as today - the projection is scoped to the `metadata` field only.

This mirrors MongoDB's projection ("source" in some systems), but deliberately constrained to the `metadata` subdocument so it is simple, safe, and doesn't let callers reshape the core card contract.

### Parameter

- Name: `metadata_fields` (query parameter).
- Semantics: **include-only allowlist.** Caller lists the metadata paths to KEEP; all other metadata keys are dropped from the response.
- Path syntax: **dot-notation**, MongoDB-native. Examples:
  - `metadata_fields=owner` -> keep only `metadata.owner`
  - `metadata_fields=owner,config.region,limits.rps` -> keep only those nested paths
- Omitted / empty -> **no change** (full `metadata` returned, exactly as today).
- Unknown / non-existent paths -> silently produce no value for that path (no error), same as a MongoDB projection over a missing field.

### Design decisions (agreed)

1. **Projection mode: include-only allowlist.** No exclude/denylist in v1 (avoids the "cannot mix include and exclude" edge cases).
2. **Path syntax: dot-notation nested paths**, applied to the `metadata` subdocument only.
3. **Scope: semantic search, list endpoints, and single-asset GET** (consistent everywhere).

## Scope

Apply the `metadata_fields` projection to:

### List endpoints (already return full metadata - projection prunes it)
- `GET /servers` ([server_routes.py:541](registry/api/server_routes.py#L541)) - metadata returned at L704
- `GET /api/agents` ([agent_routes.py:700](registry/api/agent_routes.py#L700)) - metadata returned at L858
- `GET /skills` ([skill_routes.py:162](registry/api/skill_routes.py#L162)) - metadata returned at L207

### Semantic search
- `POST /api/search/semantic` ([search_routes.py:406](registry/api/search_routes.py#L406))
- Note: server and skill search results do **not** currently return `metadata` (`ServerSearchResult` [search_routes.py:97](registry/api/search_routes.py#L97), `SkillSearchResult` [search_routes.py:174](registry/api/search_routes.py#L174)). Agent search returns the full `agent_card` ([search_routes.py:167](registry/api/search_routes.py#L167)), which includes metadata. To honor `metadata_fields` consistently, this issue **adds** an (optionally projected) `metadata` field to `ServerSearchResult` and `SkillSearchResult`. When `metadata_fields` is omitted, search results stay as they are today (no metadata for servers/skills) to preserve backward compatibility - metadata is only added to those results when the caller explicitly projects it.

### Single-asset GET
- The individual server / agent / skill GET handlers - same `metadata_fields` parameter, same metadata-only projection.

## Implementation Notes

- **Backend scope: DocumentDB (MongoDB) only.** The FAISS / file-based search backend is deprecated and slated for removal, so this feature adds no FAISS-specific behavior (shared function signatures are kept intact only so the deprecated path still compiles until it is deleted).
- **Where to project (two layers).** For list + single-GET (reads that query the DB directly), project **in the database** so huge `metadata` is dropped inside DocumentDB and never read into the app. A plain inclusion projection cannot keep the whole document while reshaping one subdocument, so use an aggregation `$set` stage that rebuilds `metadata` from only the requested dot-paths (e.g. `{"$set": {"metadata": {"config": {"region": "$metadata.config.region"}}}}`), composing with the existing `{"tool_list": 0}` exclusion on servers. For semantic search (asset already in memory during assembly), prune in Python via a shared `project_metadata()` helper, which is also the single canonical definition of the projection semantics.
- **A shared helper** `project_metadata(metadata, paths)` implements the include-only dot-path prune once; `build_metadata_set_stage(paths)` produces the equivalent DB `$set`. Both call `normalize_metadata_paths(paths)` first so the two layers agree and an integration test pins their equivalence.
- **Robustness (no read may 500 on a bad projection).** Three layers: (1) the parser rejects empty / `$`-prefixed segments, over-deep/over-long paths, and too many paths with `422`; (2) `normalize_metadata_paths` drops paths that are descendants of another requested path (`config,config.region` -> `config`) so the `$set` can't emit a malformed duplicate key; (3) the repository runs the `$set` pipeline in `try/except` and, on any `PyMongoError`, falls back to an unprojected read + the in-Python prune, logs a WARNING, and increments `registry_metadata_projection_db_fallback_total`. The Python prune is the canonical oracle and safety net, so the DB stage is a pure optimization that can never break the API. A path into a scalar/missing branch (`owner.x`) simply resolves to "missing" in both layers - no error.
- **Skill metadata is structured** (`SkillMetadata` with `author`, `version`, `extra`; [skill_models.py:53](registry/schemas/skill_models.py#L53)). Dot-paths apply over its serialized form (e.g. `metadata_fields=author,extra.team`).
- **Parsing**: accept a comma-separated list (and/or repeated query params). Cap the number of paths and the path depth/length, and reject empty/`$`-prefixed segments, to avoid abuse (return `422`).
- **Search-result metadata sourcing (verified, no N+1).** For semantic search: server and agent results already re-fetch the full asset during assembly, so the full `metadata` is in hand and projection costs no extra query. Skill search results are assembled from a metadata *subset* stored in the search index ([search_repository.py ~L1070](registry/repositories/documentdb/search_repository.py#L1070)), so projecting arbitrary skill metadata paths requires the full skill doc; fetch the page's skills with a single batched query (add `SkillRepository.list_by_paths(paths)` mirroring `ServerRepository.list_by_ids` at [server_repository.py:197](registry/repositories/documentdb/server_repository.py#L197)) only when `metadata_fields` is supplied. Per-page, never per-result.

## User Stories

- As an API consumer listing servers, I want to request only `metadata.owner` and `metadata.config.region` so I don't download multi-MB metadata blobs I don't need.
- As a search client, I want my results to carry just the handful of metadata fields I render, keeping responses small and fast.
- As a registry operator, I want large customer-supplied nested metadata to not be shipped in full on every list/search call by default-able client behavior.

## Acceptance Criteria

- [ ] `metadata_fields` query param added to `GET /servers`, `GET /api/agents`, `GET /skills`.
- [ ] `metadata_fields` accepted by `POST /api/search/semantic`.
- [ ] `metadata_fields` accepted by the single-asset GET for server, agent, and skill.
- [ ] Dot-notation nested paths are honored (e.g. `config.region`), include-only semantics.
- [ ] When `metadata_fields` is omitted/empty, responses are byte-for-byte unchanged from today.
- [ ] Only the `metadata` field is projected; all other card fields are returned in full regardless.
- [ ] `ServerSearchResult` and `SkillSearchResult` gain an optional projected `metadata` field, populated only when `metadata_fields` is supplied.
- [ ] A shared `project_metadata()` helper implements the include-only dot-path prune once and is reused by all endpoints.
- [ ] Unknown paths produce no value (no error). Malformed input (e.g. too many paths, excessive depth) is rejected with `422`.
- [ ] Unit tests for `project_metadata()` (top-level, nested, missing, mixed) and integration tests per endpoint.

## Out of Scope

- Projection of fields **outside** `metadata` (the rest of the card is never reshaped by this param).
- Exclude/denylist semantics (`-field`) - include-only in v1.
- A generic per-asset "view"/projection DSL across the whole card.
- Changing default response shapes when `metadata_fields` is absent.

## Related Code

- Semantic search: [search_routes.py:406](registry/api/search_routes.py#L406), result models at [:97](registry/api/search_routes.py#L97)/[:157](registry/api/search_routes.py#L157)/[:174](registry/api/search_routes.py#L174)
- List endpoints: [server_routes.py:541](registry/api/server_routes.py#L541), [agent_routes.py:700](registry/api/agent_routes.py#L700), [skill_routes.py:162](registry/api/skill_routes.py#L162)
- Existing projection precedent: [server_repository.py:139](registry/repositories/documentdb/server_repository.py#L139)
- Metadata definitions: server dict (`_coerce_metadata_to_dict` [server_routes.py:175](registry/api/server_routes.py#L175)), [agent_models.py:446](registry/schemas/agent_models.py#L446), [skill_models.py:53](registry/schemas/skill_models.py#L53)

cj-taylor:
Closes #1277

## Summary

Adds an optional `metadata_fields` parameter to all read endpoints (list, single-GET, semantic search) and the CLI. When supplied, only the requested dot-notation metadata paths are returned. When omitted, responses are unchanged. Supports both comma-separated (`?metadata_fields=owner,config.region`) and repeated query params (`?metadata_fields=owner&metadata_fields=config.region`).

## Changes

- **Shared helpers** (`registry/utils/metadata.py`): parse, normalize, project, build_set_stage
- **Endpoints**: GET /api/servers, GET /api/agents, GET /skills, GET single-asset for all three, POST /api/search/semantic
- **DB optimization**: aggregation pipeline in server_repository with transparent Python fallback (DocumentDB 5.0 already pinned in `documentdb.tf:322`)
- **Skill search**: batch-fetches full skill docs for deep metadata projection (single $in query via `list_by_paths`)
- **CLI**: `--metadata-fields` on list, agent-list, skill-list, server-search
- **Client lib**: `metadata_fields` param on `semantic_search()` and `list_skills()`
- **Validation**: max 20 paths, depth 5, no $-prefix segments (422 on violation)
- **Tests**: 89 feature-specific tests (76 unit + 10 server API + 3 skill integration), including DB/Python equivalence assertions and edge cases

## Open Item

Search results include `"metadata": null` when `metadata_fields` is not supplied. This matches how `sync_metadata`, `record_kind`, etc. already serialize on the same model. See [comment on the issue](https://github.com/agentic-community/mcp-gateway-registry/issues/1277#issuecomment-5303655803) for context.

## Test Results

- 7,161 passed, 0 failed, 74 skipped
- Ruff lint + format: clean
- Live-tested against running Docker stack (API + CLI)

## Notes

- `uv.lock` churn is from `uv sync` + adding ruff as dev dep, unrelated to feature
- No new config params (no unified-parameter-reference update needed)


cj-taylor:
Ready for review. One open question on the issue about `metadata: null` serialization on search results, happy to adjust either way.

aarora79:
Follow-up #1673 (merged) adds the review's optional hardening — a Unicode-safe path-segment allowlist and `HTTP_422_UNPROCESSABLE_CONTENT` — plus a What's New entry (`1.30.0`), an FAQ, and search-doc cross-links for this feature.

PR #1646, opened 2026-08-18
fix(lua): keep empty inputSchema arrays as [] in virtual tools/list

### LINKED ISSUE #1532: Virtual server tools/list corrupts empty JSON-Schema `required: []` into `{}`, breaking strict MCP clients
## Summary

When a tool's `inputSchema.required` is an **empty array** (`[]`), the **virtual server** `tools/list` response returns it as an **empty object** (`{}`). JSON Schema requires `required` to be an array of strings, so strict MCP clients (any using Zod/JSON-Schema validation, e.g. Amazon Kiro) reject the **entire** `tools/list` response and show **zero** tools.

The same backend tool served through its **per-server proxy path** returns the correct `required: []`. Only the **virtual server aggregation path** exhibits the corruption.

## Environment

- Registry/gateway deployed from upstream images; bug also present in current `main` (`6bccc19`).
- Transport: Streamable HTTP.
- Backend server: AWS Knowledge MCP server (`https://knowledge-mcp.global.api.aws`), which advertises `required: []` for tools that have no required parameters.

## Reproduction

1. Register a backend MCP server that exposes at least one tool whose `inputSchema` has `"required": []` (e.g. the AWS Knowledge server's `read_documentation` and `list_regions`).
2. Create a virtual server that includes those tools.
3. Call `tools/list` on both endpoints and compare `inputSchema.required` for a no-required-params tool.

### Direct per-server path — CORRECT

`POST https://<gateway>/aws-kb/mcp` → `tools/list`:

```
aws___read_documentation   required = []      (array) ✅
aws___list_regions         required = []      (array) ✅
```

### Virtual server path — BROKEN

`POST https://<gateway>/virtual/<name>/mcp` → `tools/list`:

```
aws___read_documentation   required = {}      (object) ❌
aws___list_regions         required = {}      (object) ❌
```

(Tools that *do* have required params, e.g. `search_documentation` with `required: ["search_phrase"]`, are unaffected — only the empty case is corrupted.)

### Client-side symptom (Amazon Kiro)

```
Error listing tools: [
  { "expected": "array", "code": "invalid_type",
    "path": ["tools", 0, "inputSchema", "required"],
    "message": "Invalid input: expected array, received object" },
  { "expected": "array", "code": "invalid_type",
    "path": ["tools", 2, "inputSchema", "required"],
    "message": "Invalid input: expected array, received object" }
]
```

Because validation fails on the whole array, **all** tools from the virtual server are dropped, even the valid ones.

## Expected vs actual

- **Expected:** `inputSchema.required` is emitted as a JSON array in the virtual server response (`[]` when there are no required properties), matching the backend and the per-server path.
- **Actual:** empty `required` is emitted as `{}` in the virtual server response.

## Likely location

The virtual server builds each tool's schema from the backend server's **stored** `tool_list` and passes it through `_ensure_mcp_compliant_schema`, which only normalises the top-level `type` and never touches `required`:

- `registry/core/nginx_service.py:1727` — `input_schema = st.get("inputSchema", st.get("input_schema", {}))`
- `registry/core/nginx_service.py:1730` — `input_schema = _ensure_mcp_compliant_schema(input_schema)`
- `registry/core/nginx_service.py:453` — `_ensure_mcp_compliant_schema(...)`

Since the per-server proxy (live pass-through) returns `[]` but the virtual server (built from the persisted `tool_list`) returns `{}`, the empty array appears to be turned into an empty object at tool discovery/persistence time (e.g. an empty-list → empty-dict coercion when the schema is stored/retrieved), and it is never re-normalised before being served.

## Suggested fix

Ensure `required` is always a JSON array before it is served. A minimal, defensive normalisation inside `_ensure_mcp_compliant_schema` (so every served schema is corrected regardless of how it was persisted):

```python
# normalise `required` to an array (JSON Schema requires an array)
req = input_schema.get("required")
if isinstance(req, dict):
    input_schema = {**input_schema, "required": list(req.keys())}  # {} -> []
```

Ideally also fix the persistence path so `required: []` is stored as an array rather than `{}`. The same guard could sensibly be applied to any array-typed JSON Schema keyword that can be empty (`enum`, `allOf`, etc.).

## Impact

Any strict MCP client that validates tool schemas cannot use a virtual server that aggregates a tool with no required parameters — the whole tool list is rejected. This makes virtual servers unusable with such clients whenever any member tool has empty `required`.

ebarkhordar:
`lua-cjson` cannot distinguish an empty array from an empty object: both decode to a bare Lua table, and that table re-encodes as `{}`. The virtual router decodes each backend's `tools/list` response and re-encodes it, once into the enriched shared-dict cache and once into the JSON-RPC result, so a backend advertising `"required": []` reaches the client as `"required": {}`.

That also accounts for the two asymmetries in the report: a non-empty `required` carries integer keys and re-encodes as an array, and the per-server proxy path is a byte pass-through with no cjson round trip, so only the virtual aggregation path corrupts the schema.

### Why the change is not in `nginx_service.py`

The report points at `_ensure_mcp_compliant_schema` and the persisted `tool_list`. Python's `json` module round-trips `[]` as `[]`, and that helper feeds the mapping file, which the Lua router reads only when every backend `tools/list` fetch has failed. In a deployment with a reachable backend that branch never runs, so a change there would not reach the reported symptom. Normalising in Lua covers the mapping-file branch as well.

### Fix

`_ensure_mcp_schema` is the single point every `tools/list` schema crosses on all three paths: live enrichment, the per-backend mapping fallback, and the scope filter that rebuilds the response on every request, cache hit or miss. It now walks the schema and marks empty array-valued keywords with `cjson.empty_array_mt`.

Object-valued keywords are left alone, since `{}` is correct for `properties` and `additionalProperties`. Members of `properties` are not matched against the keyword list, so a tool parameter that happens to be named `required` keeps its object form.

### Verification

- New checks in `tests/lua/test_virtual_router.lua`, driven through `_handle_tools_list`: 3 fail on `main` and pass here (empty `required` on the live path, the same value served from cache, and a nested empty `enum`).
- The full Lua suite passes, including guard checks that a non-empty `required`, an empty `properties` and a property named `required` are all unchanged.
- Run in `openresty/openresty:alpine` with the same luajit invocation as `lua-test.yml`, plus that workflow's syntax check and `pre-commit` over both changed files.
- Not verified end to end: I did not stand up nginx, MongoDB and a live backend. The checks drive the aggregation path with `ngx.location.capture` mocked, so they pin the encoding behaviour rather than the deployment.

Fixes #1532

### DIFF
--- a/docker/lua/virtual_router.lua
+++ b/docker/lua/virtual_router.lua
@@ -120,11 +120,48 @@ local function _forward_identity_headers()
 end
 
 
+-- JSON Schema keywords whose value must serialize as an array.
+local SCHEMA_ARRAY_KEYS = {
+    required = true, enum = true, allOf = true, anyOf = true,
+    oneOf = true, examples = true, prefixItems = true,
+}
+
+-- Keywords holding a map of caller-chosen names to subschemas. A member named
+-- "required" there is a property name, not the keyword, so descend into the
+-- members without matching their names against SCHEMA_ARRAY_KEYS.
+local SCHEMA_MAP_KEYS = {
+    properties = true, patternProperties = true,
+    definitions = true, ["$defs"] = true,
+}
+
+-- cjson decodes an empty JSON array to a bare table, which re-encodes as {}.
+-- Tag those tables so the keywords above survive the round trip as [].
+local function _tag_empty_schema_arrays(node, depth)
+    if type(node) ~= "table" or depth > 16 then
+        return
+    end
+    for key, value in pairs(node) do
+        if type(value) == "table" then
+            if SCHEMA_MAP_KEYS[key] then
+                for _, subschema in pairs(value) do
+                    _tag_empty_schema_arrays(subschema, depth + 1)
+                end
+            elseif SCHEMA_ARRAY_KEYS[key] and next(value) == nil then
+                setmetatable(value, empty_array_mt)
+            else
+                _tag_empty_schema_arrays(value, depth + 1)
+            end
+        end
+    end
+end
+
+
 -- Ensure inputSchema has "type": "object" as required by MCP spec
 local function _ensure_mcp_schema(schema)
     if not schema or type(schema) ~= "table" then
         return { type = "object", properties = {} }
     end
+    _tag_empty_schema_arrays(schema, 0)
     if schema.type == "object" then
         return schema
     end
--- a/tests/lua/test_virtual_router.lua
+++ b/tests/lua/test_virtual_router.lua
@@ -202,6 +202,61 @@ do
 end
 
 -- ---------------------------------------------------------------------------
+print("test: _handle_tools_list keeps empty schema arrays as [] (issue #1532)")
+do
+    dict._store["tools_enriched:srv3"] = nil
+    local mapping = { required_scopes = nil, tools = {
+        { name = "no_arg", original_name = "no_arg", backend_location = "/ok" },
+        { name = "one_arg", original_name = "one_arg", backend_location = "/ok" },
+    } }
+    capture_responses = {}
+    capture_responses["/ok"] = { status = 200, body = cjson.encode({ result = { tools = {
+        { name = "no_arg", inputSchema = {
+            type = "object", properties = {},
+            required = setmetatable({}, cjson.empty_array_mt) } },
+        { name = "one_arg", inputSchema = {
+            type = "object",
+            properties = { q = { type = "string",
+                                 enum = setmetatable({}, cjson.empty_array_mt) } },
+            required = { "q" } } },
+    } } }) }
+
+    local body = M._handle_tools_list("1", mapping, "", nil, "srv3")
+    check(body:find('"required":[]', 1, true) ~= nil,
+        "empty required serializes as [] not {}")
+    check(body:find('"required":["q"]', 1, true) ~= nil,
+        "non-empty required is still an array")
+    check(body:find('"properties":{}', 1, true) ~= nil,
+        "empty properties stays an object")
+    check(body:find('"enum":[]', 1, true) ~= nil,
+        "empty enum nested under properties serializes as []")
+
+    -- The cached path decodes and re-encodes, so it must hold the same shape.
+    local cached_body = M._handle_tools_list("1", mapping, "", nil, "srv3")
+    check(dict._store["tools_enriched:srv3"] ~= nil, "result was cached")
+    check(cached_body:find('"required":[]', 1, true) ~= nil,
+        "empty required is still [] when served from cache")
+end
+
+-- ---------------------------------------------------------------------------
+print("test: a property named like a schema keyword stays an object")
+do
+    dict._store["tools_enriched:srv4"] = nil
+    local mapping = { required_scopes = nil, tools = {
+        { name = "odd", original_name = "odd", backend_location = "/ok" },
+    } }
+    capture_responses = {}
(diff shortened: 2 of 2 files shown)


aarora79:
Thanks for finding and fixing this, @ebarkhordar. The diagnosis is right: the cjson round trip on the virtual aggregation path is what turns `"required": []` into `"required": {}`, the non-empty case survives because it has integer keys, and the direct proxy path is unaffected because it's a byte passthrough. The `_tag_empty_schema_arrays` approach and the properties-vs-array keyword handling are solid.

We tested it end to end against a running stack and hit something worth flagging: the Lua fix is correct but it does nothing on the shipped runtime.

It depends on `cjson.empty_array_mt`, which only exists in OpenResty's lua-cjson. The registry image (`docker/Dockerfile.registry`) installs Debian `lua-cjson` 2.1.0, where `empty_array_mt` is `nil`, so `setmetatable(value, empty_array_mt)` is a no-op and the empty array still serializes as `{}`. The CI Lua suite passes because `lua-test.yml` runs in `openresty/openresty:alpine`, which has the extension. So CI is green while the deployed gateway is not.

Confirmed three ways:
1. E2E: created a virtual server over a backend that advertises `required: []`, ran `initialize` + `tools/list` through the virtual endpoint, got `"required": {}` even with your fix deployed (verified byte-identical in the container).
2. The container runs Debian `nginx/1.26.3` + `lua-cjson 2.1.0+dfsg`, not OpenResty.
3. Clean-room repro of that exact package: `empty_array_mt = nil`, and applying the fix produces `"required":{}`.

The complete fix needs your Lua change plus a runtime change that ships OpenResty's lua-cjson in the registry image (build it against the nginx-lua LuaJIT, drop the Debian package, pin `lua_package_cpath`). Both halves have to land together to close #1532, so we're going to combine them into a single PR rather than keep two that are each inert on their own. That PR will cherry-pick your commit (`ced1500e`) with your authorship preserved, add the runtime change, a build-time check that fails the image build if the shipped cjson lacks `empty_array_mt`, and an integration test that runs this E2E against the real image. We rebuilt the image with the runtime change and re-ran the E2E: `tools/list` now returns `"required": []`, and empty objects like `"properties": {}` stay `{}`.

We'll close this PR in favor of the combined one once it's up and link it here. Your commit and authorship carry over.

One more note: switching to OpenResty's cjson helps beyond this issue. `_as_json_array` already uses `empty_array_mt`, so on the Debian runtime any empty array the virtual router emits (empty `tools`/`resources`/`prompts` lists, not just schema fields) had the same problem. This makes that whole class of serialization correct and lines the shipped runtime up with what CI already tests. Thanks again for raising it.

aarora79:
@dheerajoruganty see comment above. @omrishiv FYI.

aarora79:
Combined PR is up: #1676. It cherry-picks your commit (`ced1500e`) with your authorship, adds the OpenResty lua-cjson runtime change, a build-time guardrail, and an E2E test. That one closes #1532.

ebarkhordar:
Thanks for testing it against the real image, and for carrying the commit over.

That runtime gap is one my verification could not have caught: I ran the repo's own `lua-test.yml` in `openresty/openresty:alpine`, which is the one image that has `empty_array_mt`, so a green suite there says nothing about the shipped registry. `docker/Dockerfile.registry` on main does install the Debian `lua-cjson` package next to `nginx-extras`, so the metatable is absent where it matters.

The build-time check that fails the image build is the part worth keeping. Glad to have it land in #1676.

PR #1644, opened 2026-08-17
fix(nginx): add proxy_http_version 1.1 to /validate block + forward i…

doncouture:
…dentity headers in virtual server backends

Two fixes for ECS Service Connect deployments:

1. /validate auth subrequest missing proxy_http_version 1.1 ECS Service Connect uses HTTP/1.1 keepalive connections between the envoy sidecar and the upstream container. Without proxy_http_version 1.1 on the /validate location, nginx defaults to HTTP/1.0 which closes the connection after each subrequest. Under load this exhausts the connection pool and causes intermittent 502s on auth validation.

2. Virtual server _vs_backend_* locations do not forward caller identity (same fix from PR #1650, rebased onto 1.29.0)
   - virtual_router.lua: set X-User/X-Username via ngx.req.set_header
   - nginx_service.py: proxy_set_header X-User/X-Username using $http_x_*

Both fixes tested on ECS Fargate with Cognito auth and Service Connect.

*Issue #, if available:*
Closes #1643  (proxy_http_version), relates to #1650 (identity headers)

*Description of changes:*
Adds proxy_http_version 1.1 to the /validate auth subrequest location (fixes intermittent 502s on ECS Service Connect) and rebases the virtual server identity header fix from #1650 onto 1.29.0.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #1643, opened 2026-08-17
/validate auth subrequest missing proxy_http_version 1.1 - causes 502s on ECS Service Connect


doncouture:
## Summary

The `/validate` auth subrequest location in `nginx_rev_proxy_http_and_https.conf` and `nginx_rev_proxy_http_only.conf` is missing `proxy_http_version 1.1;`. This causes intermittent 502 errors on ECS deployments using Service Connect.

## Environment

- MCP Gateway Registry: v1.29.0
- Deployment: AWS ECS Fargate with ECS Service Connect (envoy sidecar)
- Auth provider: Cognito (same issue would affect any provider)

## Root cause

ECS Service Connect injects an envoy sidecar that proxies inter-service traffic using HTTP/1.1 keepalive connections. The `/validate` location block uses the default `proxy_http_version 1.0`, which sends `Connection: close` on every subrequest. This works initially, but under sustained load:

1. nginx sends HTTP/1.0 requests to the auth-server via the envoy sidecar
2. envoy expects HTTP/1.1 and sometimes reuses a connection nginx already signaled to close
3. The result is intermittent `502 Bad Gateway` on the auth subrequest
4. Every authenticated request (MCP proxy, API, virtual server) fails

Every other `proxy_pass` location in the same config files already has `proxy_http_version 1.1;` - only the `/validate` block was missed.

## Steps to reproduce

1. Deploy the registry to ECS with Service Connect enabled
2. Generate sustained traffic (multiple concurrent MCP sessions)
3. Observe intermittent 502s on requests that require auth validation

The issue is timing-dependent and may not appear under light load since HTTP/1.0 closes the connection cleanly on each request. It surfaces when envoy's connection pool tries to reuse a connection that nginx flagged for close.

## Fix

Add `proxy_http_version 1.1;` after `proxy_pass` in the `/validate` location blocks:

```nginx
location = /validate {
    internal;
    proxy_pass http://{{AUTH_SERVER_HOST}}:{{AUTH_SERVER_PORT}}/validate;
    proxy_http_version 1.1;

    # Pass original request info
    ...
}
```

Affected files:
- `docker/nginx_rev_proxy_http_and_https.conf` (2 occurrences - HTTP and HTTPS server blocks)
- `docker/nginx_rev_proxy_http_only.conf` (1 occurrence)

This matches every other `proxy_pass` location in the same files which already specify `proxy_http_version 1.1;`.

## Impact

[...]

PR #1653, opened 2026-08-18
feat(go-validate): Go fast-path sidecar for the /validate hot path

### LINKED ISSUE #1652: Go fast-path sidecar for the auth /validate hot path (with Python fallback)
## Problem

nginx fires an `auth_request` subrequest to the auth-server's `GET /validate` on **every** proxied request through the gateway (`auth_server/server.py:2890`). That handler runs on a **single uvicorn worker** (`server.py:4830`, no `workers=`), so it is GIL-bound to one core. The repo's own stress baseline recorded it topping out at **~172.7 rps at concurrency 50** (`tests/stress/results/validate/baseline.json`), and at high concurrency p99 latency climbs into the second range.

Because gateway mode is the dominant deployment shape, this single endpoint is effectively the throughput ceiling of the entire authenticated data path. As traffic grows, it becomes the gateway's dominant scaling limit.

## Proposed solution

Introduce a small **Go fast-path sidecar** that fronts `/validate`:

- Verifies the token type that carries the traffic (RS256 JWT from the configured IdP), builds the exact identity headers nginx consumes, and mints the HS256 internal token — all **byte-identical** to the Python output.
- **Reverse-proxies everything it does not implement** (session cookies, other IdPs, opaque/introspection tokens, OBO, admin/federation tokens) to the unchanged Python auth-server. Blast radius is zero: you cannot break what you did not implement.
- The cutover is **one line of nginx config** (`proxy_pass`, `docker/nginx_rev_proxy_http_only.conf:~522`) plus a reload; instant rollback.

The RS256 verification is IdP-agnostic; each additional IdP is a config entry + a small claim-map, not a rewrite. Anything not yet fast-pathed simply falls through to Python and keeps working.

## User stories

- As a **gateway operator**, I want the per-request auth check to sustain far higher throughput and lower tail latency, so the gateway stops being the bottleneck as traffic grows.
- As a **platform owner**, I want to adopt the speedup with a one-line, instantly-reversible change and no behavior difference, so I take on zero migration risk.
- As a **maintainer**, I want unsupported auth flows to keep working unchanged, so the fast path never regresses correctness for any deployment.

## Acceptance criteria

- [ ] A Go sidecar serves `GET /validate` and returns **byte-identical** status + identity headers + body vs Python for the RS256 fast path (machine-checked diff, ≥3 token/group/scope shapes).
- [ ] The 4 negative cases (tampered signature, malformed JWT, expired token, missing `Authorization`) return the same status on both.
- [ ] All non-fast-path requests (cookie / other IdP / opaque / unknown `kid`) are transparently reverse-proxied to Python and succeed unchanged.
- [ ] After the nginx `proxy_pass` flip, a real MCP request through the gateway front door succeeds (the `airegistry-tools` canary returns `serverInfo`), and a fallback request also succeeds.
- [ ] Benchmark shows a large, honestly-reported throughput/latency/CPU/memory win, including the 401-path fairness run and a `uvicorn --workers N` comparison.
- [ ] Fail-closed preserved: recognized-but-invalid → 401; unrecognized → defer to Python, never a silent allow.
- [ ] Security: strip client-supplied identity/trust headers before minting; validate `SECRET_KEY` (missing/weak) at startup; never log secrets/tokens/claims.

## Out of scope

- Replacing the full Python handler (federation/admin static tokens, session cookies, OBO exchange, per-tool ACL, rate limiting, audit) — these stay in Python via fallback.
- Terraform/ECS and Helm/EKS wiring, unified parameter reference, System Config page (deferred to productionization).
- Accelerating IdPs beyond the one that carries traffic (added incrementally as config + a claim map).

## Notes

- Baseline reference: `tests/stress/results/validate/baseline.json` (~172.7 rps @ c=50).
- Design docs (LLD, testing plan, expert review) prepared under `.scratchpad/` (not committed).
- Suggest adding a `go` label for Go-related work.

aarora79:
## Summary

Adds a stdlib-only Go **fast-path sidecar** (`go-validate/`) that fronts the auth-server's `GET /validate` `auth_request` endpoint. nginx calls `/validate` on every proxied request; the Python handler runs on a single uvicorn worker and is the throughput ceiling of the authenticated data path (repo baseline: ~172.7 rps @ c=50, `tests/stress/results/validate/baseline.json`).

`go-validate` serves the steady-state configured-IdP RS256 bearer path in Go and **reverse-proxies everything else to the unchanged Python auth-server**. It is byte-identical where it answers and zero-risk where it does not — the cutover is a single nginx `proxy_pass` line.

Closes #1652.

## What it does

- **Fast path:** RS256 verify against a cached JWKS (atomic keyset, background refresh, last-good retention) -> claim map -> mint the HS256 `X-Internal-Token-Registry` -> write the identity headers nginx consumes.
- **Fallback:** cookies, other IdPs, opaque tokens, unknown `kid`, or an unconfigured fast path are reverse-proxied to `auth-server:8888`.
- **Endpoints:** `GET /validate`, `GET /health` (readiness; degraded when JWKS unhealthy), `GET /metrics` (plaintext counters).

## Security invariants baked in

- Strips client-supplied identity/trust headers before minting (no identity injection).
- Validates `SECRET_KEY` for missing AND weak at startup — fails closed.
- Enforces `iss`/`aud` from config, never from the token.
- Recognized-but-invalid -> `401`; unrecognized -> fallback (never a silent allow).

## Scope (hackathon)

**In:** RS256 fast path + Python fallback, Docker/compose surface.
**Deferred (stays on Python via fallback):** federation/admin static tokens, session cookies, OBO exchange, per-tool ACL, rate limiting, audit. Terraform/ECS + Helm/EKS wiring, unified parameter reference, and System Config page are deferred to productionization.

## Testing

- `go build ./...` and `go vet ./...` pass; static binary ~9.4 MB.
- Design, testing plan, and expert review (with blockers folded in) prepared under `.scratchpad/` (not committed).
- Before merge / demo: byte-identical parity harness (>=3 token shapes + 4 negative cases), MCP-canary connectivity test through the gateway after the nginx flip, and the benchmark sweep (incl. 401 fairness + `uvicorn --workers N`).

## Build with Claude Code

Engineered with Claude Code + Claude Opus via the `new-feature-design` skill: iterative issue -> design (LLD + diagrams) -> testing plan -> adversarial multi-persona review -> implementation, refined with human feedback.


aarora79:
Moving this to draft. Not because anything is broken — main is merged in, conflicts are resolved, and CI is green — but because the decision this PR asks for should be made on data we do not have yet, and the cheapest way to get that data does not involve this PR.

**Where it stands.** Rebased onto main (the repo only allows rebase-merge, so the earlier merge-commit resolution could not be replayed). Two conflict rounds: the three nginx `/validate` blocks, where this branch templates the upstream and main added `proxy_http_version 1.1` in #1644 — both kept, since dropping the latter reintroduces the 502s from #1643; and the terraform variable blocks in the root and module files, both additive. Verified after the rebase: 6868 Python unit tests, `go build`/`vet`/`test`, ruff, helm unittest 69/69 and 90/90, the stack chart rendering 60 resources, `terraform validate`, and 22/22 CI checks.

Also fixed two things found while reviewing it:
- `go-validate/Dockerfile` declared no `ARG BUILD_VERSION`, so the arg every build path already passes (compose, `scripts/build-images.sh`, and the release workflow's `github.ref_name`) was silently discarded and the binary carried no version. It is now stamped via `-ldflags -X` and reported on the startup log, `GET /health` (header + body), and `GET /metrics` as `govalidate_build_info{version="..."}`.
- The fast-path prose in `docs/unified-parameter-reference.md` had been inserted mid-table, orphaning the `App secret key` and `Advertised OAuth scopes` rows below it with no header.

**Why it is not merging yet.**

1. **The fast path bypasses three controls, and the fallback cannot save it.** `auth_server/server.py:4249` calls `_enforce_rate_limit` inside `/validate`; quarantine is enforced through that same limiter (`decision.quarantined`, `server.py:1799`); and MCP access audit records are written at `server.py:4315` and `:4414`. `grep -riE "rate.?limit|quarantin|audit" go-validate/*.go` returns nothing. The fallback triggers on **token shape**, not on feature state — so an RS256 bearer is served in Go whether or not rate limits are configured and whether or not the caller is quarantined. An admin who quarantines a caller would watch them keep working, and the request would be missing from the audit trail. The PR describes these as "deferred, stays on Python via fallback", which does not hold for a deployment that has them enabled.

2. **Nobody knows the addressable share of traffic.** The UI's "Get JWT Token" and per-server tokens are self-signed HS256, which always falls back — `TESTING.md` calls this out as the top source of "it's not accelerating" confusion. If those dominate production `/validate` traffic, this accelerates very little. That number decides whether a second auth codebase and an extra container per task are worth carrying, and it can be measured **in the Python handler with label additions alone** — no container, no task-definition change, no second implementation in main. Tracked in the follow-up issue.

The proof of concept did its job: it established that `/validate` is the throughput ceiling of the authenticated data path (172.7 rps @ c=50) and that the check itself is cheap in Go. What it has not established is how much real traffic is eligible, or that acceleration is safe for deployments using rate limiting, quarantine, or audit.

**Gate to come out of draft:** eligibility mix from the metrics work, plus a feature-aware fallback (defer to Python when rate limiting or durable audit is enabled) so the flag stops being a trap. If the eligible share turns out small, the honest comparison is `uvicorn --workers N`, which this PR's own testing plan flags as the control nobody has run.

Branch state is tagged so nothing here is lost, and one more rebase later is cheaper than carrying an unjustified container.

PR #1675, opened 2026-08-21
feat(egress): application-layer encryption for per-user egress credentials

### LINKED ISSUE #1665: Security: add application-layer encryption for per-user egress credentials before vault persistence
## Summary

Add application-layer authenticated encryption for per-user egress credentials before they are persisted to AWS Secrets Manager or OpenBao.

Today the egress credential vault stores the serialized `StoredToken` payload directly in the configured secret-store backend. This means that anyone with sufficient privileges to read the underlying AWS Secrets Manager secret or OpenBao KV entry can obtain the user's upstream bearer credentials.

The application should encrypt egress credential material before handing it to the secret-store backend, so the backing vault contains ciphertext rather than usable PAT, OAuth access-token, or OAuth refresh-token values.

## Current behavior

Both egress credential backends currently persist the `StoredToken` representation directly:

AWS Secrets Manager:

```python
await self._mutate(
    root_name,
    keys.map_key(provider, server_path),
    token.model_dump(),
)
```

The resulting document is subsequently serialized and supplied to AWS Secrets Manager as `SecretString`.

OpenBao similarly writes:

```python
self._client.secrets.kv.v2.create_or_update_secret(
    path=path,
    secret=token.model_dump(),
    mount_point=self._mount,
)
```

This means the secret-store provider performs its normal server-side encryption at rest, but the application itself does not encrypt the credential before persistence.

## Motivation / threat model

Secrets Manager and OpenBao are appropriate storage systems for credentials, but they are also an infrastructure-level security boundary.

For high-security deployments it is desirable to add another independent layer:

```text
PAT / OAuth token
        |
        v
Application-layer authenticated encryption
        |
        v
Ciphertext
        |
        +--> AWS Secrets Manager
        |
        +--> OpenBao
```

This provides defense in depth against scenarios such as:

* a privileged Secrets Manager/OpenBao operator gaining read access to stored egress credentials;
* accidental over-permission of `GetSecretValue` or equivalent vault permissions;
* secret-store backups/exports being exposed;
* compromise of the vault storage layer without simultaneous compromise of the application encryption key.

Reading the vault alone should therefore not be sufficient to recover an MCP user's upstream PAT, OAuth access token, or refresh token.

## Proposed design

Introduce application-layer authenticated encryption around `StoredToken` before it reaches either persistent secret-store implementation.

A modern AEAD construction such as **AES-256-GCM** should be used, with a fresh random nonce for every encryption operation.

The application should have a dedicated high-entropy root secret, for example:

```text
EGRESS_CREDENTIAL_ENCRYPTION_KEY
```

This should be separate from `SECRET_KEY`, both for cryptographic domain separation and to allow egress-credential encryption keys to have an independent lifecycle and rotation policy.

### Per-user key derivation

The root key should not be used directly to encrypt every user's credentials.

Instead, derive a deterministic, cryptographically independent key for each canonical user principal using HKDF-SHA256.

Conceptually:

```text
root application key
        +
canonical principal identity
(auth_method + user_id)
        |
        v
      HKDF
        |
        v
per-user AES-256 encryption key
```

The principal identity should use the same stable/canonical identity used when constructing the user's secret-store namespace, so the same user deterministically receives the same derived key across application replicas and restarts.

For example, the derivation context could be versioned and domain-separated along the lines of:

```text
mcp-gateway-egress-credential-v1 || canonical(auth_method, user_id)
```

The `user_id` should therefore **not simply be concatenated with the root key and treated as key material**. It should be incorporated through a proper KDF such as HKDF.

This gives every user a different effective encryption key even though the deployment has one root encryption secret.

### Bind ciphertext to the credential address

The authenticated-encryption Associated Data (AAD) should additionally bind the ciphertext to its logical vault address:

```text
auth_method
user_id
provider
server_path
```

This prevents an encrypted credential copied from one user/server entry to another from being successfully decrypted in the wrong context.

## Stored representation

The vault should contain an explicitly versioned encrypted envelope rather than plaintext `StoredToken` fields.

For example:

```json
{
  "_encrypted": true,
  "version": 1,
  "algorithm": "AES-256-GCM",
  "key_id": "v1",
  "nonce": "<base64>",
  "ciphertext": "<base64>"
}
```

The exact representation is implementation-defined, but it should contain enough version/key metadata to support future algorithm changes and key rotation without ambiguity.

The plaintext `StoredToken`, including:

```text
access_token
refresh_token
client_id
scopes
expiry/status metadata
```

can be serialized and encrypted as one authenticated payload.

Encrypting the entire `StoredToken` rather than only the bearer-token fields also avoids leaking unnecessary credential metadata to the storage administrator.

## Key custody

The application-layer root encryption key must **not be stored in the same egress Secrets Manager/OpenBao trust boundary that this feature is intended to protect against**.

If an actor can retrieve both:

```text
encrypted egress credentials
+
EGRESS_CREDENTIAL_ENCRYPTION_KEY
```

the additional encryption layer provides no meaningful isolation.

Deployment documentation should therefore explicitly describe the required trust boundary for the root key.

For deployments whose threat model includes a privileged AWS account administrator, simply storing `EGRESS_CREDENTIAL_ENCRYPTION_KEY` in another Secrets Manager secret in the same account is insufficient. Such deployments need a key-custody mechanism outside that administrator's effective read/control boundary.

This feature should be documented as protecting against vault/storage compromise unless stronger external key custody is configured; it should not claim to protect against full compromise of an application runtime that can access plaintext credentials.

## Backend behavior

Application-layer encryption should be backend-independent.

Both:

```text
SECRET_STORE_BACKEND=secrets-manager
```

and:

```text
SECRET_STORE_BACKEND=openbao
```

should store only encrypted credential payloads.

The secret-store API should ideally continue exposing `StoredToken` to its callers so existing PAT, 3LO refresh, vend, delete, and connection-management logic does not need to know whether encryption is enabled.

Encryption/decryption should therefore happen in a common credential-codec/encryption layer rather than implementing two different cryptographic schemes in the AWS and OpenBao backends.

## Backward compatibility / migration

Existing installations may already contain plaintext `StoredToken` entries, so an upgrade path is required.

Suggested behavior:

1. New writes are always application-encrypted once the feature is enabled.
2. Reads recognize both the legacy plaintext representation and the new versioned encrypted representation during migration.
3. A legacy plaintext entry can be migrated to encrypted form either:

   * on successful read ("read repair"), or
   * through an explicit migration command/job.
4. After migration, deployments should have an option to reject legacy plaintext entries entirely.
5. The encrypted format should include a key/version identifier so encryption-key rotation can support old-key decrypt/new-key encrypt without forcing every user to reconnect their MCP account.

## Acceptance criteria

* PAT credentials are encrypted by the application before being persisted.
* 3LO OAuth access tokens and refresh tokens are encrypted by the application before being persisted.
* AWS Secrets Manager contains ciphertext rather than usable `StoredToken` credential values.
* OpenBao contains ciphertext rather than usable `StoredToken` credential values.
* AES-256-GCM or an equivalent authenticated-encryption primitive is used with a unique random nonce for each encryption.
* A dedicated deployment-level root encryption key is used.
* A separate deterministic encryption key is derived for each `(auth_method, user_id)` principal using HKDF or an equivalent KDF.
* The full `(auth_method, user_id, provider, server_path)` credential address is cryptographically bound using AEAD associated data.
* The root encryption key is never persisted alongside the encrypted egress credentials.
* The encrypted payload format is versioned and supports key rotation.
* Existing plaintext entries have a documented migration path.
* AWS Secrets Manager inline and overflow-sharded layouts continue to work.
* OpenBao per-connection KV storage continues to work.
* PAT create/read/update/delete tests cover encrypted persistence.
* 3LO token create/read/refresh/delete tests cover encrypted persistence.
* Cross-user ciphertext substitution fails authentication/decryption.
* Cross-server/provider ciphertext substitution fails authentication/decryption.
* Incorrect or unavailable encryption keys fail closed and never cause credentials to be returned or overwritten with plaintext.
* Sensitive plaintext or cryptographic material is never emitted in application logs or exception messages.

## Security benefit

This creates two independent controls around egress credentials:

```text
                   application-held root key
                             |
                             v
User credential ---> per-user AEAD encryption
                             |
                             v
                       ciphertext only
                             |
                  +----------+----------+
                  |                     |
          AWS Secrets Manager        OpenBao
          encryption/ACLs            encryption/ACLs
```

Compromise or privileged read access to the backing secret store alone would therefore expose encrypted blobs rather than immediately usable upstream MCP credentials.

This is particularly valuable because these credentials can include long-lived PATs and OAuth refresh tokens and can therefore represent substantially more durable authority than a short-lived gateway session token.

omrishiv:
Fixes #1665

Encrypt per-user egress credentials with AES-256-GCM before they are handed to the secret-store backend (AWS Secrets Manager / OpenBao), so the vault holds a versioned ciphertext envelope instead of usable PAT / OAuth access / refresh tokens. Gated by a new dedicated root key, EGRESS_CREDENTIAL_ENCRYPTION_KEY; empty preserves the legacy plaintext behavior.

Crypto (registry/secrets/credential_codec.py):
- Per-(auth_method, user_id) key derived from the root key via HKDF-SHA256 (the root key is never used directly), keyed on the same canonical principal that namespaces the vault.
- AEAD associated data binds the full (auth_method, user_id, provider, server_path) address AND the envelope version/algorithm/key_id, so a ciphertext copied to another user/server/provider fails authentication and the metadata cannot be tampered to steer decryption.
- Fresh random 96-bit nonce per write; versioned envelope {_encrypted, version, algorithm, key_id, nonce, ciphertext}.

Fail-closed: an envelope seen without a key, a wrong key, tamper, unsupported version/algorithm/key_id, or malformed payload all raise; credentials are never returned or re-persisted as plaintext, and no plaintext/key material is logged.

Backward compatibility / migration:
- Both backends encode on write and decode on read; reads recognize legacy plaintext and the new envelope.
- Legacy plaintext entries are transparently re-encrypted on first read (read-repair), non-blocking, using compare-and-set so a migration never clobbers a concurrent token refresh: Secrets Manager under the principal mutation lease, OpenBao via KV v2 cas.
- EGRESS_CREDENTIAL_ENCRYPTION_REQUIRE_ENCRYPTED (strict mode) rejects any lingering plaintext once migration is complete.

Key custody & rotation: the root key must live outside the secret-store trust boundary it protects. Rotation is a destructive cutover today (single active key); the envelope carries a key_id as the seam for a future multi-key keyring.

Deployment surfaces wired for the key: docker-compose (x3), Helm registry chart (secret), Terraform ECS (Secrets Manager valueFrom + IAM). CDK left untouched (it has no egress vars yet). Docs and .env.example updated.


aarora79:
Follow-up #1694 addresses the review items deferred from this PR: it surfaces the encryption key (masked) and the `EGRESS_CREDENTIAL_ENCRYPTION_REQUIRE_ENCRYPTED` toggle in the System Config UI (`CONFIG_GROUPS`), and adds an operator nudge for when the vault is enabled but the key is unset — a startup WARNING, an `mcpgw_registry_recommended_config_not_set` gauge, and a UI badge, all driven by a single recommended-config registry.

PR #1676, opened 2026-08-21
fix(1532): keep empty inputSchema arrays as [] (lua fix + OpenResty cjson runtime)

### LINKED ISSUE #1532: Virtual server tools/list corrupts empty JSON-Schema `required: []` into `{}`, breaking strict MCP clients

aarora79:
Closes #1532. Supersedes #1646 (its Lua commit is cherry-picked here with @ebarkhordar's authorship preserved).

## Background

An MCP backend that advertises `"required": []` on a tool reached clients through a virtual server as `"required": {}`. lua-cjson cannot distinguish an empty array from an empty object (both decode to a bare Lua table that re-encodes as `{}`), and the virtual router does a decode/re-encode round trip on the `tools/list` aggregation path. The direct per-server proxy path is a byte passthrough, so only the virtual path corrupted the schema.

## Why one PR

The complete fix has two interdependent halves; neither works alone:

1. **Lua** (`docker/lua/virtual_router.lua`, from #1646): `_tag_empty_schema_arrays` tags empty array-valued JSON-Schema keywords (`required`, `enum`, `allOf`, ...) with `cjson.empty_array_mt` so they survive the round trip as `[]`. Map-valued keywords (`properties`, `$defs`, ...) are descended into without matching member names, so a property literally named `required` stays an object.
2. **Runtime** (this PR): `cjson.empty_array_mt` is an OpenResty lua-cjson extension. The registry image shipped Debian `lua-cjson` 2.1.0, where it is `nil`, so `setmetatable(value, empty_array_mt)` was a no-op and the empty array still serialized as `{}` in production. CI passed only because `lua-test.yml` runs in `openresty/openresty:alpine`, which has the extension.

Verified end to end: created a virtual server over a backend advertising `required: []`, ran `initialize` + `tools/list` through the virtual endpoint. On the old image it returned `"required": {}`; after this change it returns `"required": []`, and empty objects like `"properties": {}` correctly stay `{}`.

## Changes

- `docker/Dockerfile.registry`: add a `cjson-builder` stage that builds OpenResty's lua-cjson against the nginx-lua LuaJIT, drop the Debian `lua-cjson` package, and `COPY` the built `cjson.so` into the runtime image. A build-time assertion fails the image build if the shipped cjson lacks `empty_array_mt` (or if an empty array-tagged table does not encode as `[]`).
- `docker/nginx_rev_proxy_http_and_https.conf`, `docker/nginx_rev_proxy_http_only.conf`: pin `lua_package_cpath` so `require "cjson"` resolves to the OpenResty build.
- `docker/lua/virtual_router.lua` + `tests/lua/test_virtual_router.lua` (from #1646): the fix and its unit tests, plus a guard that fails the suite if run against a cjson without `empty_array_mt`.
- `tests/integration/test_virtual_empty_schema_e2e.py`: live-stack E2E asserting `required` serializes as `[]` through a virtual server. Skips unless `MCP_GATEWAY_E2E_URL` / `MCP_GATEWAY_E2E_TOKEN` are set.

## Scope note

Switching to OpenResty's cjson helps beyond this issue: `_as_json_array` already relies on `empty_array_mt`, so on the Debian runtime any empty array the virtual router emits (empty `tools`/`resources`/`prompts` lists, not just schema fields) had the same problem. This change corrects that whole class of serialization and aligns the shipped runtime with what CI already tests.

The registry image is shared across Docker Compose, ECS (Terraform), and EKS (Helm), so this one image change covers all three deployment surfaces.

## Testing

- Full Lua suite passes in `openresty/openresty:alpine` (incl. the schema-array cases and the new env guard).
- Rebuilt the registry image with the runtime change and re-ran the E2E on a live stack: `tools/list` returns `"required": []`; `"properties": {}` stays `{}`; no lua/cjson errors; metrics lua unaffected.
- New integration test collects and skips cleanly with no live stack; `ruff` clean.

No configuration parameters changed.

### DIFF
--- a/tests/integration/test_virtual_empty_schema_e2e.py
+++ b/tests/integration/test_virtual_empty_schema_e2e.py
@@ -1,0 +1,138 @@
+"""End-to-end check that empty inputSchema arrays survive as ``[]`` (issue #1532).
+
+This is a live-stack test, not an in-process one. The empty-array serialization
+bug lived in the nginx + lua virtual router, which the FastAPI-app integration
+tests cannot exercise, and it only reproduced on the shipped runtime (Debian
+nginx + Debian lua-cjson, where ``cjson.empty_array_mt`` is nil). So this test
+drives a real gateway over HTTP.
+
+It skips unless a gateway URL and bearer token are provided, so it is a no-op in
+the standard unit/integration CI. Point it at a running stack to exercise it:
+
+    export MCP_GATEWAY_E2E_URL=http://localhost
+    export MCP_GATEWAY_E2E_TOKEN="$(python3 -c \
+      "import json;print(json.load(open('.token'))['tokens']['access_token'])")"
+    uv run pytest tests/integration/test_virtual_empty_schema_e2e.py -v
+
+The durable, always-on guardrail for the same property is the build-time
+assertion in ``docker/Dockerfile.registry`` (the image build fails if the shipped
+cjson lacks ``empty_array_mt``); this test adds an HTTP-level confirmation when a
+stack is available.
+"""
+
+import json
+import os
+import uuid
+
+import httpx
+import pytest
+
+_GATEWAY_URL = os.environ.get("MCP_GATEWAY_E2E_URL")
+_TOKEN = os.environ.get("MCP_GATEWAY_E2E_TOKEN")
+
+pytestmark = pytest.mark.skipif(
+    not (_GATEWAY_URL and _TOKEN),
+    reason="Set MCP_GATEWAY_E2E_URL and MCP_GATEWAY_E2E_TOKEN to run the live E2E.",
+)
+
+_MCP_HEADERS = {
+    "Content-Type": "application/json",
+    "Accept": "application/json, text/event-stream",
+}
+
+
+def _auth_headers() -> dict[str, str]:
+    return {"Authorization": f"Bearer {_TOKEN}"}
+
+
+def _parse_mcp_body(text: str) -> dict:
+    """Return the JSON object from a plain or SSE-framed MCP response body."""
+    data_lines = [line[len("data:") :].strip() for line in text.splitlines() if line.startswith("data:")]
+    return json.loads(data_lines[-1] if data_lines else text)
+
+
+def _find_backend_with_empty_required(client: httpx.Client) -> tuple[str, str] | None:
+    """Find a (server_path, tool_name) whose tool advertises ``required: []``."""
+    resp = client.get("/api/servers", params={"limit": 200}, headers=_auth_headers())
+    resp.raise_for_status()
+    for server in resp.json().get("servers", []):
+        for tool in server.get("tool_list", []):
+            schema = tool.get("schema") or tool.get("inputSchema") or {}
+            if isinstance(schema.get("required"), list) and not schema["required"]:
+                return server["path"], tool["name"]
+    return None
+
+
+def test_empty_required_serializes_as_array_through_virtual_server() -> None:
+    base = _GATEWAY_URL.rstrip("/")
+    vs_path = f"/virtual/e2e-empty-required-{uuid.uuid4().hex[:8]}"
+
+    with httpx.Client(base_url=base, timeout=30.0) as client:
+        backend = _find_backend_with_empty_required(client)
+        if backend is None:
+            pytest.skip("No registered backend advertises a tool with required: [].")
+        backend_path, tool_name = backend
+
+        create_body = {
+            "path": vs_path,
+            "server_name": "E2E empty-required (issue 1532)",
+            "description": "Transient VS asserting empty required stays [].",
+            "tool_mappings": [{"tool_name": tool_name, "backend_server_path": backend_path}],
+            "required_scopes": [],
+            "tool_scope_overrides": [],
+            "tags": ["e2e", "issue-1532"],
+            "supported_transports": ["streamable-http"],
+            "is_enabled": True,
+        }
+
+        try:
+            created = client.post("/api/virtual-servers", json=create_body, headers=_auth_headers())
+            assert created.status_code == 201, created.text
+
+            toggled = client.post(
+                f"/api/virtual-servers{vs_path}/toggle",
+                json={"enabled": True},
+                headers=_auth_headers(),
+            )
+            assert toggled.status_code == 200, toggled.text
+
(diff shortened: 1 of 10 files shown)


aarora79:
@dheerajoruganty can you please review these changes.

tests/integration/test_virtual_empty_schema_e2e.py
```diff
@@ -0,0 +1,142 @@
...
+            session_id = init.headers.get("Mcp-Session-Id")
+            assert session_id, f"no Mcp-Session-Id from initialize: {init.text}"
+
+            listed = client.post(
+                mcp_url,
+                headers={**_auth_headers(), **_MCP_HEADERS, "Mcp-Session-Id": session_id},
+                content=json.dumps(
+                    {"jsonrpc": "2.0", "id": 2, "method": "tools/list", "params": {}}
+                ),
+            )
```
omrishiv: There's a potential for a race condition here between enabling the server, nginx reloading it, and then running through these steps. it might be good to make sure to wait for each step to be ready

omrishiv (approved):
Looks good, I deployed it and ran it through E2E with a live server as well, just a comment on the integration test

tests/integration/test_virtual_empty_schema_e2e.py
aarora79: Good catch. Fixed in a5b7809d: after enabling the virtual server the test now polls `initialize` until it returns an `Mcp-Session-Id` (so it waits out the nginx reload / mapping write instead of racing it), then retries `tools/list` until backend enrichment yields tools. Both are bounded by a 30s readiness timeout that fails with the last response. Thanks for deploying and running it live.

aarora79:
While testing this branch on ECS we hit an unrelated latent bug and fixed it here too (commit d8d438f5):

- After a Secrets Manager rotation, the DocumentDB password contained an RFC 3986 reserved character. The hand-built Mongo connection URIs did not URL-encode credentials, so auth/registry tasks crash-looped on startup. Now `quote_plus`-escaped in both entrypoint readiness loops and in `registry/utils/mongodb_connection.py` (username/password + IAM branches).
- Also fixed the `vs-toggle` CLI: `registry_client.py` was calling non-existent `/enable` and `/disable` paths (405); it now POSTs to `/toggle` with `{"enabled": ...}`.

Both verified against the live ECS deployment.

ebarkhordar:
Thanks for carrying this through and for the OpenResty runtime change and E2E test on top of it. Good to see it land on main.

Issue #1461, opened 2026-07-13
Egress OBO: Keycloak RFC 8693 token-exchange path (Phase 4 follow-up to #1269)


aarora79:
Follow-up to #1269 (which shipped the `token-exchange` / `obo_exchange` egress mode with the **Microsoft Entra `jwt-bearer`** provider path). This issue tracks the remaining provider path.

## Problem

The OBO egress mode (`egress_auth_mode = "obo_exchange"`) is implemented and working for Entra, but the **Keycloak RFC 8693 token-exchange** path is a deliberate Phase-4 stub: `auth_server/egress_obo.py` `_keycloak_exchange_body()` raises `OboUnsupportedIdpError("Keycloak OBO token-exchange (RFC 8693) is not yet implemented; Entra (jwt-bearer) ships first. Tracked as Phase 4.")`.

## Proposed solution

Implement the Keycloak branch of `obo_exchange`:

- RFC 8693 token exchange: `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`, the user's ingress token as `subject_token` (with `subject_token_type=...:access_token`), and the backend as the `audience` (bare target client id, not Entra's `scope=api://.../.default`).
- Reuse the auth-server's configured Keycloak token endpoint and client credentials.
- Same fail-closed contract as the Entra path: strip the ingress token, exchange, inject; on failure return an error with no relay and no client-credentials fallback.
- Reject a backend audience the realm does not own (no federation guessing).

## Acceptance criteria

- [ ] Keycloak branch of `_keycloak_exchange_body` / `obo_exchange` implemented (no longer raises).
- [ ] Against a same-realm Keycloak backend, the exchanged token preserves the user identity (delegated, not client-only).
- [ ] Exchange failure returns an error with no relay and no client-credentials fallback.
- [ ] The egress path logs no token material.

## Out of scope

Public SaaS backends (served by `vault-oauth`); the `vault-pat` and custom-header modes (#1268).

ruspg:
Hi! We run the gateway with Keycloak as the IdP and picked this up — the draft PR is up: #1678. Flagging the approach here for visibility; happy to reshape anything.

**Implementation** (`_keycloak_exchange_body` in `auth_server/egress_obo.py`):

- `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`
- `subject_token` + `subject_token_type=urn:ietf:params:oauth:token-type:access_token`
[...]

Issue #1269, opened 2026-06-19
Egress: token-exchange (OBO) for same-trust-domain backends


aarora79:
Part of #1265 (Egress Token Credentials Broker). **Sub-issue 4 of 5. Depends on sub-issue 3 (the `egress_auth_mode` setting). No vault dependency.**

## Problem

When the gateway IdP and the backend resource live in the **same trust domain** (same Entra tenant, same Keycloak realm, or an explicitly federated pair), the gateway can exchange the user's ingress token for a backend-scoped token, preserving the user identity cryptographically. This is the cleanest egress mode where it applies, but it only reaches same-trust-domain backends (e.g. Outlook / Microsoft 365); public SaaS does not federate with our IdP.

## Proposed solution

Add the `token-exchange` egress mode:

- Entra: `jwt-bearer` grant with the ingress token as the assertion, requesting the backend app audience. The backend may do its own second hop to Microsoft Graph.
- Keycloak: RFC 8693 token exchange with the ingress token as the subject token and the backend as the audience.
- Reuse the auth-server's already-configured IdP token endpoint and client credentials.
- A new per-server setting carries the target audience.
- Fail closed: strip the ingress token, exchange, inject. On exchange failure, return an error and do not relay the ingress token. Never silently fall back to client-credentials (which would drop the user identity).
- No vault; nothing is stored per user. An exchanged token may be briefly cached, default off, capped well below its expiry.

## Acceptance criteria

- [ ] `token-exchange` mode added with a per-server target-audience setting.
- [ ] Against a same-tenant Entra backend, the exchanged token preserves the user identity (verified as a delegated token, not app-only).
- [ ] Exchange failure returns an error with no relay and no client-credentials fallback.
- [ ] A backend whose audience the IdP does not own is rejected (no federation guessing).
- [ ] A small Entra-wired MCP server is delivered as the reusable integration-test fixture.
- [ ] The egress path logs no token material.

## Out of scope

Public SaaS backends (GitHub, Slack, Atlassian), which cannot use token exchange and are served by `vault-pat` or `vault-oauth`.

## Notes

The exchange must use an async HTTP client in the async egress handler. The assertion must be the original IdP-issued token, not a gateway-internal token; confirm the original bearer is available at the exchange point.

aarora79:
Decision: build now but inert behind a default-off flag; do not enable in any deployment until a same-tenant M365 target exists.

aarora79:
[...]

Issue #1666, opened 2026-08-19
Egress OBO: optional caching of exchanged tokens in the configured SecretStore


zoltan-fedor:
## Summary

Add an optional cache for Microsoft Entra OBO (`obo_exchange`) tokens so deployments can choose to reuse a still-valid exchanged token rather than performing a new token exchange against Entra for every MCP request.

The existing behavior should remain available and should preferably remain the default.

## Current behavior

The current OBO implementation is deliberately stateless.

`auth_server/egress_obo.py` documents the following security invariant:

```python
# The minted token embeds the user's ``sub``; it is exchanged PER REQUEST and is
# NEVER cached or reused across users. This module holds no cache.
```

Each request to an MCP server configured with:

```text
egress_auth_mode=obo_exchange
```

therefore performs an OBO token exchange against the configured IdP before forwarding the request to the MCP server.

For Entra this means that, conceptually:

```text
MCP request
    |
    v
Gateway
    |
    +----> Entra token endpoint
    |        OBO exchange
    |<---- access token
    |
    +----> MCP server
```

This has a strong security property: the gateway does not retain the resulting OBO token and every request receives a freshly exchanged token.

It also means that every MCP invocation incurs an additional network request to Entra.

## Motivation

Some deployments may prefer to trade a small amount of the current security/isolation property for reduced token-exchange overhead.

For workloads involving frequent MCP calls, performing an Entra OBO exchange for every individual MCP request can introduce:

* additional request latency;
* additional dependency on Entra availability on the MCP request hot path;
* increased calls to the Entra token endpoint;
* additional operational load and possible throttling considerations.

The gateway already has a pluggable per-user credential `SecretStore` supporting AWS Secrets Manager and OpenBao.

It would be useful to optionally cache short-lived OBO access tokens there and reuse them until they approach expiry.

This assumes that, for a given deployment, reading a cached credential from the configured SecretStore is operationally preferable to performing another OBO exchange with Entra.

That may not be true for every deployment, which is why this behavior should be optional.

## Historical context

The original OBO feature request (#1269) already anticipated this possibility:

> An exchanged token may be briefly cached, default off, capped well below its expiry.

The implementation shipped in #1420 chose the stronger stateless model and explicitly introduced a per-request/no-cache invariant.

This request proposes implementing the optional caching mode originally contemplated by #1269 without changing the existing stateless mode.

## Proposed behavior

Introduce an optional OBO token cache.

For example:

```text
[...]

PR #1678, opened 2026-08-23
feat(egress): implement Keycloak RFC 8693 token-exchange for OBO

ruspg:
Refs #1461 (see *Not covered* below — this does not close it on its own)

## What

Implements the Keycloak half of egress OBO token exchange (Phase 4 follow-up to #1269): replaces the `not yet implemented` stub in `_keycloak_exchange_body` with a real RFC 8693 standard-token-exchange request.

- `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`
- `subject_token` + `subject_token_type=urn:ietf:params:oauth:token-type:access_token`
- `requested_token_type` **pinned** to `access_token` rather than left to the server default — see *Version semantics*
- `audience` = the bare target client id (registration-time validation already constrains audience shape upstream)
- the gateway's own `client_id` / `client_secret` as form fields
- `scope` sent only when explicit scopes are requested; omitting it makes Keycloak apply the **requesting** client's default client scopes — the gateway's own, not the audience client's

### Error mapping

`_map_token_error` now takes the IdP family so a provider-specific remediation hint cannot reach an operator running the other IdP:

| code | mapping | note |
|---|---|---|
| `invalid_grant`, `invalid_token` | `OboReauthRequired` | Keycloak never answers `invalid_grant` for token-exchange; legacy exchange reports an unusable `subject_token` as `invalid_token`, so the most likely runtime failure — the ingress JWT expiring mid-flight — was landing in the generic bucket |
| `unsupported_grant_type` | `OboConfigError` | the feature is not enabled on the server |
| `access_denied` | `OboConfigError` — **Keycloak only** | names both authorization models rather than asserting one; see *Version semantics*. Deliberately not applied to Entra, which returns this for denied consent and for conditional-access blocks, and where reclassifying would move an already-released path out of the `exchange_failed` audit bucket |
| `invalid_client`, `invalid_scope`, `unauthorized_client` | `OboConfigError` | unchanged |
| `invalid_request` | **deliberately unclassified** | standard exchange answers it for an expired `subject_token`, for the client's exchange toggle being off, and for an unplaceable audience — two config, one re-auth. The code alone cannot separate them, and calling an expired token a config problem would stop the caller retrying with a fresh one. `error_description` is the only discriminator, so it is now logged |

The IdP's `error_description` is now logged (truncated; it carries configuration text, never a token) — without it the overloaded codes above are undiagnosable.

Also in this PR: a blank `target_audience` is rejected before any credential leaves the process, and `resp.json()` on the success path is wrapped so a 200 with a broken body stays an `OboExchangeError` instead of escaping as a 500.

## Version semantics

This matters more than it looks, because the repo ships three Keycloak majors across four surfaces:

| surface | Keycloak | `KC_FEATURES` |
|---|---|---|
| `docker-compose*.yml`, `terraform/aws-ecs` | 25.0 | `token-exchange,admin-api` |
| `infra/` (CDK) | 26.2.4 | `token-exchange` |
| `charts/` (Helm) | 26.3.3 | *(absent)* |

Legacy exchange (≤ 26.1) and standard exchange (26.2+) differ in ways this code path touches directly:

- **default `requested_token_type`** — legacy `refresh_token`, standard `access_token`. Unpinned, the legacy path made Keycloak mint a refresh token on every per-request exchange that the code reads past and discards. Now pinned.
- **what `access_denied` means** — legacy: the *target* client lacks the token-exchange permission. Standard: the *gateway's own* client is outside the subject token's audience. The message names both.
- **on 26.2+ both are enabled** (`token-exchange` is the v1 key; v2 is on by default and takes precedence), so the same request is served by different implementations depending on the client's Standard-token-exchange toggle.

The live verification was done on Keycloak 25 — i.e. **legacy** exchange. The standard-exchange path is reasoned from the docs and source, not exercised.

## Tests

26 unit tests, all passing. Beyond body shape and error mapping:

- the POST target URL is asserted — previously it was captured and never checked, so a regression redirecting the gateway's `client_secret` and the user's JWT to another endpoint would have passed every assertion in the file;
- scope inputs use shapes that survive `ServerInfo._validate_egress_auth`, which binds each scope's resource prefix to `target_audience` — bare OIDC names like `profile` are rejected at registration and asserting on them proves nothing;
- an Entra `access_denied` test pins that the new branch does **not** reclassify that path.

## Design question — resolved by #1707

`obo_exchange` routes the token POST through `guarded_async_client` with `CREDENTIALED_OAUTH_PROFILE` (HTTPS-only, deliberately empty allowlist), so an in-cluster IdP on a private address — including the shipped compose default `KEYCLOAK_URL=http://keycloak:8080` — was blocked before the body was ever built. The question this section originally raised (trust the configured IdP `token_url`, or add an opt-in allowlist knob) has since been answered upstream: #1707 added `EGRESS_OAUTH_TRUSTED_IDP_HOSTS`, an operator-named allowlist of exact IdP hostnames, and 7d2f85b6 propagates it to the auth-server, which is the process that performs the exchange.

This PR keeps the guard untouched and documents the two prerequisites that remain for a self-hosted Keycloak: the IdP host must be listed in `EGRESS_OAUTH_TRUSTED_IDP_HOSTS`, and TLS stays required (`require_https=True` on the profile; an internal CA works through the process trust store, there is no per-hop CA-bundle setting). In-cluster non-TLS OBO remains out of scope for now, per review.

## Not covered by this PR

Listed so the tracking issue is not closed by mistake:

1. **No integration test** proving the exchanged token is delegated with the subject's `sub` preserved. #1269's checklist asks for that, plus a negative test for an audience the IdP does not own and a regression test that `client_credentials` never appears on the OBO path. None ship here.
2. **Design doc updated here; the operator doc is not.** `docs/design/egress-auth-design.md` now states that Keycloak RFC 8693 is implemented (status line, mode table, provider paragraph) and gains a "Keycloak (self-hosted) as the OBO identity provider" subsection with the request shape and the `EGRESS_OAUTH_TRUSTED_IDP_HOSTS` / TLS prerequisites. `docs/obo-token-exchange.md` still has no Keycloak setup section, no `KEYCLOAK_*` row in the env table, and a failure-triage table of `AADSTS*` codes only; the ops notes below belong there — follow-up.
3. **`extra_audiences` is a no-op on Keycloak.** `auth_server/server.py` computes per-server OBO audiences and passes them to `validate_token`, and `_server_advertises_per_server_prm()` returns true for OBO on *any* provider — but `KeycloakProvider.validate_token(**kwargs)` swallows the argument and its allowlist is hardcoded. Harmless while Keycloak ignores RFC 8707 `resource`, but an operator who adds an audience mapper for the advertised per-server resource gets a fail-closed `Invalid audience` with no knob. Not touched here; happy to fix in a follow-up if you want it separate.
4. **`egress_oauth.scopes` is effectively unusable on Keycloak.** `_obo_scope_mismatches_target` requires a scope's resource prefix to equal `target_audience`, which is an Entra convention; every scope a Keycloak operator would want (`profile`, `email`, …) is rejected at registration, and the shapes that survive are not Keycloak client-scope names. The happy path (empty scopes → target's defaults) works, and the UI exposes no scopes field, so this is latent rather than breaking.
5. **The Helm chart never enables token exchange** (`KC_FEATURES` appears in no chart), so that surface needs a manual step.

## Ops notes (for `docs/obo-token-exchange.md`)

- legacy exchange (Keycloak ≤ 26.1) needs the target client's token-exchange permission via fine-grained admin permissions, otherwise the token endpoint answers `access_denied — Client not allowed to exchange`;
- standard exchange (26.2+) instead needs the gateway's own client inside the subject token's `aud`;
- the exchange uses the gateway IdP provider's `client_id` — the **web** client, not `m2m_client_id` — so whichever permission applies must be granted there;
- the subject token's `iss` must match the token endpoint's host, which matters when the gateway and the caller reach the IdP through different front doors.

### DIFF
--- a/auth_server/egress_obo.py
+++ b/auth_server/egress_obo.py
@@ -111,29 +111,99 @@ def _keycloak_exchange_body(
     target_audience: str,
     scopes: list[str],
 ) -> dict[str, str]:
-    """Build the Keycloak RFC 8693 token-exchange request body.
+    """Build the Keycloak RFC 8693 token-exchange (OBO) request body.
 
-    Phase 4 (follow-on). Keycloak uses ``subject_token``/``audience`` (the bare
-    target client id), NOT Entra's ``assertion``/``scope=api://.../.default``.
+    Keycloak's standard token exchange authenticates the gateway's own client
+    via ``client_id``/``client_secret`` form fields, carries the caller's
+    ingress access token as ``subject_token`` (typed as an access_token), and
+    names the target as ``audience`` — the bare target client id, not an
+    https URL and not Entra's ``assertion``/``scope=api://.../.default``
+    convention.
+
+    ``requested_token_type`` is pinned rather than left to the server default,
+    because that default differs by Keycloak generation: legacy exchange
+    (<= 26.1) defaults to ``refresh_token`` and would mint a refresh token this
+    code reads past and discards on every request, while standard exchange
+    (26.2+) defaults to ``access_token``. RFC 8693 §2.1 makes the parameter
+    OPTIONAL with a server-chosen default, which is exactly why relying on it
+    is an interop hazard.
+
+    ``scope`` is sent only when explicit scopes are requested; omitting it
+    makes Keycloak apply the **requesting** client's default client scopes —
+    the gateway's own client, not the audience client.
     """
-    raise OboUnsupportedIdpError(
-        "Keycloak OBO token-exchange (RFC 8693) is not yet implemented; "
-        "Entra (jwt-bearer) ships first. Tracked as Phase 4."
-    )
+    body: dict[str, str] = {
+        "grant_type": _RFC8693_TOKEN_EXCHANGE_GRANT,
+        "client_id": client_id,
+        "client_secret": client_secret,
+        "subject_token": subject_token,
+        "subject_token_type": _RFC8693_ACCESS_TOKEN_TYPE,
+        "requested_token_type": _RFC8693_ACCESS_TOKEN_TYPE,
+        "audience": target_audience,
+    }
+    if scopes:
+        body["scope"] = " ".join(scopes)
+    return body
 
 
-def _map_token_error(status_code: int, payload: dict) -> OboExchangeError:
-    """Map an IdP token-endpoint error response to a typed exception."""
+def _map_token_error(status_code: int, payload: dict, kind: str = "") -> OboExchangeError:
+    """Map an IdP token-endpoint error response to a typed exception.
+
+    ``kind`` is the IdP family (``entra``/``keycloak``). It exists so a
+    provider-specific remediation hint never reaches an operator running the
+    other IdP: the codes overlap but their causes and fixes do not. Codes whose
+    meaning is provider-independent stay in the shared branches below.
+    """
     err = (payload.get("error") or "").strip()
     if err == "interaction_required":
         return OboConsentRequired("IdP requires consent")
-    if err == "invalid_grant":
+    if err in ("invalid_grant", "invalid_token"):
         # invalid_grant spans both user-fixable (expired/no-permission) and
-        # config (gateway not granted access) cases; surface as re-auth with the
-        # IdP detail so the agent/user sees the actual reason.
+        # config (gateway not granted access) cases; re-auth is the safer of the
+        # two, since retrying with a fresh token is cheap and a config problem
+        # simply fails again. The IdP's error_description is logged, not
+        # returned: it is operator diagnostics, not something the calling agent
+        # can act on.
+        # Keycloak never answers invalid_grant for token-exchange: legacy
+        # exchange reports an expired or unusable subject_token as
+        # invalid_token, which is the same user-fixable situation.
         return OboReauthRequired("IdP rejected the user assertion")
+    if err == "unsupported_grant_type":
+        # The token-exchange grant is not enabled on the server at all — on
+        # Keycloak <= 26.1 that means KC_FEATURES lacks token-exchange.
+        return OboConfigError(
+            f"IdP does not support the token-exchange grant "
+            f"(unsupported_grant_type, status={status_code})"
+        )
+    if err == "access_denied" and kind == "keycloak":
+        # Keycloak answers access_denied when the exchange is not permitted,
+        # and what "permitted" means depends on the server generation, so name
+        # both rather than asserting one: legacy exchange (<= 26.1) wants the
+        # token-exchange permission on the TARGET client, while standard
+        # exchange (26.2+) wants the gateway's own client inside the subject
+        # token's audience. Both are operator-actionable configuration.
+        #
+        # Deliberately NOT shared with Entra: Entra returns access_denied for
+        # denied consent and for conditional-access blocks, which are a
+        # different fix and, for CA, not operator configuration at all.
+        # Reclassifying it there would also move an already-released code path
+        # from the exchange_failed audit bucket into config_error.
+        return OboConfigError(
+            f"Keycloak denied the exchange (access_denied, status={status_code}): "
+            "grant the target client's token-exchange permission (legacy exchange), "
+            "or place the gateway client inside the subject token's audience "
+            "(standard exchange, Keycloak 26.2+)"
+        )
     if err in ("invalid_client", "invalid_scope", "unauthorized_client"):
         return OboConfigError(f"IdP rejected exchange configuration ({err})")
+    # invalid_request is deliberately NOT classified. Keycloak standard exchange
+    # (26.2+) answers it for at least three unrelated situations: an expired
+    # subject_token, the client's standard-token-exchange toggle being off, and
+    # a requested audience that cannot be placed in the token. Two are operator
+    # config and one is user re-auth, and the code alone cannot tell them apart
+    # — only error_description can, which is why it is logged at the call site.
+    # Guessing here would be worse than the generic error: classifying an
+    # expired token as config would stop the caller retrying with a fresh one.
     return OboExchangeError(
         f"IdP token exchange failed (status={status_code}, error={err or 'unknown'})"
     )
@@ -170,7 +240,23 @@ async def obo_exchange(
     token_url = getattr(idp_provider, "token_url", "") or ""
     if not token_url or not client_id or not client_secret:
         raise OboConfigError("gateway IdP credentials/token_url not configured for OBO exchange")
+    if not target_audience.strip():
+        # Registration enforces a non-empty audience, but this is the last hop
+        # before the gateway's client_secret and the user's raw JWT leave the
+        # process, and a check that can be reached with the value missing is
+        # equivalent to no check. Entra would send scope="/.default" and
+        # Keycloak a blank audience field; neither should ever be attempted.
+        raise OboConfigError("obo target_audience missing")
 
+    # CREDENTIALED_OAUTH_PROFILE enforces TLS (require_https=True): self-hosted
+    # IdPs (Keycloak is the default self-managed IdP) are supported over https
+    # only. EGRESS_OAUTH_TRUSTED_IDP_HOSTS (#1707) relaxes the public-address
+    # requirement for the named IdP hosts, not the TLS requirement, so the
+    # shipped in-cluster http://keycloak:8080 default cannot serve OBO as-is.
+    # An internal CA works via the process trust store (SSL_CERT_FILE); there
+    # is no per-hop CA-bundle setting. In-cluster non-TLS OBO is deliberately
+    # out of scope pending an explicit, operator-gated design (see
+    # docs/design/egress-auth-design.md).
     try:
         validate_url(
             token_url,
@@ -233,12 +319,29 @@ async def obo_exchange(
             payload = resp.json()
         except ValueError:
             payload = {}
+        # error_description is the only way to tell apart the situations that
+        # share an error code (notably invalid_request on standard exchange).
+        # It carries IdP configuration text, never a token or a secret, and is
+        # truncated so a verbose IdP cannot flood the log.
+        description = str(payload.get("error_description") or "")[:200]
         logger.warning(
-            f"obo_exchange: IdP token exchange failed status={resp.status_code} error={payload.get('error') or 'unknown'}",
+            "obo_exchange: IdP token exchange failed status=%s error=%s description=%s",
+            resp.status_code,
+            payload.get("error") or "unknown",
+            description or "-",
         )
-        raise _map_token_error(resp.status_code, payload)
+        raise _map_token_error(resp.status_code, payload, kind)
 
-    access_token = resp.json().get("access_token")
+    try:
+        success_payload = resp.json()
+    except ValueError as exc:
+        # The error path already tolerates a non-JSON body; the success path
+        # must too, or a 200 with a broken body escapes as an unhandled
+        # exception and the caller returns 500 instead of a typed JSON-RPC
+        # failure.
+        raise OboExchangeError("IdP returned 200 with a non-JSON body") from exc
+
+    access_token = success_payload.get("access_token")
     if not access_token:
         raise OboExchangeError("IdP returned 200 but no access_token")
     return access_token
--- a/tests/auth_server/unit/test_egress_obo.py
+++ b/tests/auth_server/unit/test_egress_obo.py
@@ -3,8 +3,8 @@
 Covers:
 - Entra jwt-bearer request body shape (grant_type, assertion, scope, on_behalf_of).
 - .default scope synthesis vs explicit scopes.
+- Keycloak RFC 8693 request body shape (subject_token, audience, client auth).
 - IdP error-code -> typed exception mapping.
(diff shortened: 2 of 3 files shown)


omrishiv (commented):
@ruspg thank you for this PR. I believe #1707 is related as it is needed for internal keycloak installs to be able to use this feature. Can you update `docs/design/egress-auth-design.md` to mention that Keycloak is now supported for OBO. 

Additionally, we might want to think about whether we want to support in-cluster, non tls OBO. The `CREDENTIALED_OAUTH_PROFILE` sets `require_tls=True`. we may want a way to override this, but for now, can we document this in the comment since keycloak is the default self-managed idp

ruspg:
Thank you for the review, and for pointing out #1707 — that is exactly the missing piece for self-hosted Keycloak deployments, glad to see it merged (and propagated to the auth-server in 7d2f85b6).

Both items are now in the PR (amended into the single commit, per the one-commit convention):

1. `docs/design/egress-auth-design.md` — the `token-exchange` (OBO) section now states that Keycloak RFC 8693 is implemented (table row + provider-support paragraph) and gains a short "Keycloak (self-hosted) as the OBO identity provider" subsection: the request shape the gateway sends, and the prerequisites for an in-cluster Keycloak — the host must be named in `EGRESS_OAUTH_TRUSTED_IDP_HOSTS` on the auth-server (#1707), and TLS is required.
2. A comment at the `CREDENTIALED_OAUTH_PROFILE` usage in `auth_server/egress_obo.py` documenting the TLS requirement (`require_https=True` on the profile): trusted-host naming relaxes the public-address requirement but not TLS, so the shipped `http://keycloak:8080` default cannot serve OBO until Keycloak is fronted with https. One thing I made explicit while writing it up: there is no application-level CA-bundle setting for this hop (`KEYCLOAK_CA_BUNDLE` is nginx-only), so an internal CA has to come in via the auth-server's process trust store (`SSL_CERT_FILE`). The doc says the same.

On in-cluster non-TLS OBO: I kept an explicit override intentionally out of scope for this PR. If there is appetite to revisit (e.g. an operator-gated, per-host opt-out, possibly together with a per-hop CA-bundle setting), I am happy to open a follow-up issue to discuss the design rather than fold it in here.

omrishiv:
@ruspg thank you for the update and apologies for my delay, I was out last week. There is a conflict that needs to be resolved before we can merge if you can update. thank you

ruspg:
Thanks, and sorry for the slow turnaround on my side too. I merged current `main` into the branch. The only conflict was the status line at the top of `docs/design/egress-auth-design.md`: it now keeps your generic-proxy wording and marks Keycloak RFC 8693 as implemented. The auth-server test suite and pre-commit pass locally on the merged tree. Since this is a fork PR, the workflow runs will need your approval again.

auth_server/egress_obo.py
```diff
@@ -233,12 +319,29 @@ async def obo_exchange(
...
+    try:
+        success_payload = resp.json()
+    except ValueError as exc:
+        # The error path already tolerates a non-JSON body; the success path
+        # must too, or a 200 with a broken body escapes as an unhandled
+        # exception and the caller returns 500 instead of a typed JSON-RPC
+        # failure.
+        raise OboExchangeError("IdP returned 200 with a non-JSON body") from exc
+
+    access_token = success_payload.get("access_token")
```
omrishiv, line 482: `[]`, `null`, `"oops"`, `3` are all valid json that will fail here

auth_server/egress_obo.py
```diff
@@ -233,12 +319,29 @@ async def obo_exchange(
             payload = resp.json()
         except ValueError:
             payload = {}
+        # error_description is the only way to tell apart the situations that
+        # share an error code (notably invalid_request on standard exchange).
+        # It carries IdP configuration text, never a token or a secret, and is
+        # truncated so a verbose IdP cannot flood the log.
+        description = str(payload.get("error_description") or "")[:200]
```
omrishiv: [], null, "oops", 3 are all valid json that will fail here

tests/auth_server/unit/test_egress_obo.py
```diff
@@ -158,13 +171,176 @@ async def test_invalid_client_maps_to_config(self, monkeypatch):
...
+        # refresh_token and would mint a credential this code discards.
+        assert body["requested_token_type"] == "urn:ietf:params:oauth:token-type:access_token"
+        # Keycloak takes the bare target client id as audience, never an https URL
+        # and never Entra's assertion/requested_token_use convention.
+        assert body["audience"] == "finance-mcp-server"
+        assert body["client_id"] == "gw-client"
+        assert body["client_secret"] == "gw-secret"
+        assert "assertion" not in body
+        assert "requested_token_use" not in body
+        # No explicit scopes -> omit the field entirely so Keycloak applies
```
omrishiv: I think this comment contradicts the docstring in `auth_server/egress_obo.py`

auth_server/egress_obo.py
omrishiv: Can we also strip `\r\n\t`

auth_server/egress_obo.py
```diff
@@ -111,29 +111,99 @@ def _keycloak_exchange_body(
...
+        "audience": target_audience,
+    }
+    if scopes:
+        body["scope"] = " ".join(scopes)
+    return body
 
 
-def _map_token_error(status_code: int, payload: dict) -> OboExchangeError:
-    """Map an IdP token-endpoint error response to a typed exception."""
+def _map_token_error(status_code: int, payload: dict, kind: str = "") -> OboExchangeError:
```
omrishiv: can we make `kind` required?

omrishiv (changes requested):
Hi @ruspg thanks so much for resolving the conflict, just a few more comments:

- All the tests prove entra OBO works, can we add a test for keycloak obo?

auth_server/egress_obo.py
ruspg: Done in db7b5baa: the success path now requires a JSON object — anything else is an `OboExchangeError` — and `access_token` must be a non-empty string. Covered by `TestMalformedIdpResponses` with `[]`, `null`, `"oops"` and `3`.

auth_server/egress_obo.py
ruspg: Done in db7b5baa: on the error path a non-object body is treated like a non-JSON one (no error fields), and a non-string `error` no longer breaks `_map_token_error`. Same test class.

ruspg: Done in db7b5baa: `_sanitize_idp_text` turns every non-printable character (CR, LF, TAB, the other C0/C1 controls, U+2028/U+2029, bidi overrides) into a space before truncating. While there I also narrowed what is logged to the error code and status, per the OAuth error-body rule in `docs/SECURITY_GUIDELINES.md`; the one exception is Keycloak's `invalid_request`, where `error_description` is the only way to tell its causes apart.

auth_server/egress_obo.py
ruspg: Done in db7b5baa.

tests/auth_server/unit/test_egress_obo.py
ruspg: Good catch — each was half right. Per Keycloak's token-exchange docs, legacy (V1) exchange applies the audience client's scopes and standard (V2, 26.2+) exchange the requesting client's. The docstring now describes both and the test comment points to it (db7b5baa).

ruspg:
@omrishiv thanks for the review — all five points are addressed (details in the threads), and I've rebased the branch onto current `main` as a linear series. A correction to my previous comment: I had merged `main` into the branch, which this repo's rebase-only merge setting doesn't allow; that merge commit is gone. (The branch was force-pushed twice today; the second push only corrected comments, docs, commit messages and one log message.)

On top of the original commit:
- `db7b5baa` fix(egress): harden OBO token-exchange response handling — your inline points.
- `72d71c1f` fix(egress): refuse exchanged Keycloak tokens that the gateway would accept — found on a closer review of the exchange path. Keycloak's legacy token exchange mints the token with the audience client's default scopes, and the shipped realm puts the `mcp-gateway` audience mapper on `basic`, so the token we forward upstream could also be accepted by the gateway itself. The engine now verifies the exchanged token (JWKS signature, issuer, expiry, target audience, `sub`) and refuses one whose `aud` contains a gateway audience.
- `a811dd92` test(egress): cover Keycloak OBO end to end through mcp_proxy — the Keycloak twin of `test_obo_integration_real_exchange_only_idp_mocked`, plus the refusal path.
- `5ea4594a` docs(egress): correct Keycloak OBO TLS, issuer and audience prerequisites.

One consequence I'd like your view on: with the shipped realm, OBO is now refused wherever legacy exchange serves the request — the Keycloak 25.0 that compose and Terraform ship, and by default the CDK's 26.2.4 with `token-exchange` enabled — unless the operator sets up standard exchange on Keycloak 26.2+ (the gateway client's Standard token exchange switch, that client in the subject token's `aud`, and the target audience available to it — an optional client scope named like the target keeps that audience out of ingress tokens) or gives the target client a copy of `basic` without the audience mapper. I derived this from Keycloak's source and docs and `init-keycloak.sh`, not from a live run. If you'd rather handle it differently — for example in the realm setup — I'm happy to adjust.

I've updated the PR description to match. The auth-server tests (1124) and pre-commit on all files (with the pytest-fast hook skipped — it needs MongoDB) pass locally. Could you approve the workflow runs on the new head when you have a moment? Thank you!

ruspg:
Follow-up on the consequence above — I've now checked it live on the Keycloak 25.0 the compose file ships (legacy exchange, with `admin-fine-grained-authz` enabled temporarily so the token-exchange permission could be granted). With the shipped realm, the token exchanged for a test target client carried `aud: ["mcp-gateway", "account", "<target>"]`, and the new check (`exchanged_token_gateway_audiences()`) reported `mcp-gateway`, so the engine refuses it. After replacing `basic` on the target client with a copy without the `mcp-gateway` audience mapper (keeping `sub`, `auth_time` and `groups`), the exchanged token carried `aud: ["account", "<target>"]` with the same `sub`, and the check reported no gateway audience. The gateway client's own `basic` was unchanged in both runs, so the audience comes from the target client's default scopes, as expected for legacy exchange.

auth_server/providers/keycloak.py
```diff
@@ -223,17 +208,101 @@ def validate_id_token(
...
+        the internal URL (e.g. http://keycloak:8080/realms/mcp-gateway) and the
+        localhost URL used in development.
+        """
+        return [
+            self.external_realm_url,
+            self.realm_url,
+            f"http://localhost:8080/realms/{self.realm}",
+        ]
+
+    def _gateway_audiences(self) -> list[str]:
```
omrishiv, line 282: can we rename this to `accepted_audiences` matching auth_server/providers/entra.py:159

auth_server/egress_obo.py
```diff
@@ -111,34 +120,205 @@ def _keycloak_exchange_body(
...
+        # denied consent and for conditional-access blocks, which are a
+        # different fix and, for CA, not operator configuration at all.
+        # Reclassifying it there would also move an already-released code path
+        # from the exchange_failed audit bucket into config_error.
+        return OboConfigError(
+            f"Keycloak denied the exchange (access_denied, status={status_code}): "
+            "grant the target client's token-exchange permission (legacy exchange), "
+            "or place the gateway client inside the subject token's audience "
+            "(standard exchange, Keycloak 26.2+)"
+        )
```
omrishiv, line 245:
```suggestion
        logger.error((
            f"Keycloak denied the exchange (access_denied, status={status_code}): "
            "grant the target client's token-exchange permission (legacy exchange), "
            "or place the gateway client inside the subject token's audience "
            "(standard exchange, Keycloak 26.2+)"
        )
        return OboConfigError("Keycloak did not permit the OBO exchange for this server")
```

auth_server/egress_obo.py
```diff
@@ -111,34 +120,205 @@ def _keycloak_exchange_body(
...
+    Args:
+        idp_provider: The gateway's Keycloak provider.
+        token: The access token returned by the exchange.
+        target_audience: The audience the exchange was requested for.
+
+    Raises:
+        OboConfigError: The token would also be accepted by this gateway.
+        OboExchangeError: The token failed verification.
+    """
+    # obo_exchange already refused a Keycloak provider without this method
```
omrishiv, line 290:
```suggestion
    jwks_url = getattr(idp_provider, "jwks_url", "") or ""
    try:
        validate_url(jwks_url, profile=CREDENTIALED_OAUTH_PROFILE)
    except UrlValidationError as exc:
        logger.error("obo_exchange: realm JWKS endpoint blocked by security policy")
        raise OboExchangeError("IdP JWKS endpoint blocked by security policy") from exc
    # obo_exchange already refused a Keycloak provider without this method
```

omrishiv (commented):
Hi @ruspg, thank you for addressing the issues. I'm leaving a few small suggestions, but we should be good to merge.

As a follow up, and I think this would also benefit #1666, we should look to provide a cache for the tokens. I have that self-assigned to me for entra, but will look to see how we extend it to Keycloak

PR #1685, opened 2026-08-25
fix(auth): strip ingress proxy-context headers on the egress hop

### LINKED ISSUE #1625: MCP proxy forwards ingress `X-Forwarded-Proto` to upstream servers, causing HTTPS redirect loops behind TLS-terminating proxies
### Description

The MCP proxy forwards ingress proxy-context headers such as `X-Forwarded-Proto` to the upstream MCP server.

In deployments where TLS terminates before the Gateway nginx/auth-server hop, the request reaching the auth-server can contain:

```http
X-Forwarded-Proto: http
```

The auth-server then makes a new outbound **HTTPS** request to the configured MCP upstream but preserves the stale ingress `X-Forwarded-Proto: http` header.

An upstream ingress/reverse proxy that trusts `X-Forwarded-Proto` can interpret this as an HTTP request and respond with an HTTPS redirect, even though the actual Gateway-to-upstream connection is already HTTPS.

This results in a redirect loop / unusable MCP route.

### Relevant code

In `auth_server/server.py`, `_forward_headers()` documents that it strips:

> hop-by-hop and proxy-hint headers so httpx can set them correctly for the connection

However, it currently does not strip standard proxy-context headers such as:

```text
Forwarded
X-Forwarded-Proto
X-Forwarded-Host
X-Forwarded-Port
X-Forwarded-Prefix
X-Original-URI
```

The resulting `forward_headers` mapping is then passed directly to the outbound `httpx.AsyncClient.stream(...)` request.

This means headers describing the **client → Gateway** transport are reused on the separate **Gateway → MCP upstream** transport.

### Reproduction

Environment:

* Gateway MCP endpoint is exposed over HTTPS.
* TLS is terminated before an internal Gateway hop.
* Upstream MCP server is configured as an HTTPS URL.
* Upstream ingress respects `X-Forwarded-Proto`.

Configured upstream:

```text
https://[domain]
```

MCP endpoint:

```text
https://[domain]/mcp
```

A direct HTTPS request succeeds:

```text
POST https://[domain]/mcp
→ 200 OK
```

A direct HTTPS request containing the stale proxy header reproduces the failure:

```http
POST https://[domain]/mcp
X-Forwarded-Proto: http
```

Result:

```text
HTTP 308 Permanent Redirect
Location: https://[domain]/mcp
```

The response body is:

```text
Permanent Redirect
```

The same request through the Gateway returns:

```text
HTTP 308
Content-Type: application/json
Content-Length: 18

Permanent Redirect
```

The Gateway auth-server logs confirm that it is actually connecting to the correct HTTPS URL:

```text
HTTP Request: POST https://[domain]/mcp "HTTP/1.1 308 Permanent Redirect"
```

A direct HTTPX request from inside the same auth-server pod to the same HTTPS endpoint succeeds with `200`, with all combinations of `verify=True/False` and `trust_env=True/False`.

The issue is also reproducible with egress authentication disabled, so it is unrelated to OBO or other egress-auth modes.

### Root cause

Ingress proxy headers describe the transport and routing context of:

```text
client → Gateway
```

They should not be blindly copied onto the separate request:

```text
Gateway → upstream MCP server
```

In particular, propagating:

```http
X-Forwarded-Proto: http
```

onto an outbound HTTPS connection gives the upstream contradictory transport information.

An ingress/load balancer in front of the MCP server may trust that header and issue an HTTP→HTTPS redirect.

### Recommended fix

Strip ingress proxy-context/routing headers when constructing the MCP upstream request.

For example, introduce a dedicated set:

```python
_PROXY_CONTEXT_HEADERS = frozenset(
    {
        "forwarded",
        "x-forwarded-host",
        "x-forwarded-port",
        "x-forwarded-proto",
        "x-forwarded-prefix",
        "x-original-uri",
        "x-original-url",
    }
)
```

and update `_forward_headers()`:

```python
if lower in _PROXY_CONTEXT_HEADERS:
    continue
```

`X-Forwarded-For` should be considered separately. There may be deployments that intentionally propagate the original client IP to an upstream server, so removing it could be a compatibility change. The routing/scheme headers above should not be propagated because they describe a different HTTP hop.

The Gateway should also **not** simply force:

```http
X-Forwarded-Proto: https
```

because HTTP upstreams are valid configurations. Dropping the stale ingress header allows the actual Gateway→upstream connection or the upstream's own trusted reverse proxy to establish the correct scheme context.

### Expected behavior

For an upstream configured as:

```text
https://example.internal/mcp
```

an ingress request containing:

```http
X-Forwarded-Proto: http
X-Forwarded-Host: gateway.example.com
X-Forwarded-Prefix: /my-server
```

should result in an upstream request that does **not** contain those ingress routing-context headers.

Normal application headers and MCP headers should continue to be forwarded according to the existing policy.

### Suggested tests

Add unit tests for `_forward_headers()` covering at least:

```text
X-Forwarded-Proto
X-Forwarded-Host
X-Forwarded-Port
X-Forwarded-Prefix
Forwarded
X-Original-URI
X-Original-URL
```

Verify they are removed case-insensitively.

Also verify existing behavior remains unchanged for:

* normal application headers;
* `Mcp-Session-Id`;
* `Accept`;
* `Content-Type`;
* ingress authentication stripping;
* the special internal Authorization relay behavior;
* OBO/vaulted egress token injection.

An integration/regression test should simulate:

```text
client request:
X-Forwarded-Proto: http

configured upstream:
https://upstream.example/mcp
```

and assert that the upstream does not receive `X-Forwarded-Proto: http`.

### Impact

This prevents MCP servers behind HTTPS-enforcing ingresses/load balancers from being used through the Gateway in common TLS-termination topologies.

It also avoids leaking frontend proxy-routing metadata across the Gateway's egress trust boundary.

AmirF194:
`_forward_headers()` copies the incoming request's headers onto the outbound
call to the upstream MCP server, stripping hop-by-hop headers and client
auth. It never stripped proxy-context headers: `X-Forwarded-Proto`,
`X-Forwarded-Host`, `X-Forwarded-Port`, `X-Forwarded-Prefix`, `Forwarded`,
`X-Original-URI`, `X-Original-URL`. Those describe the client to Gateway
hop, and the Gateway to upstream hop is a separate connection with its own
scheme and host. When TLS terminates in front of the Gateway, the request
can arrive carrying a stale `X-Forwarded-Proto: http` even though the
Gateway's own outbound call to the upstream is HTTPS. An upstream ingress
that trusts that header sees an HTTP request on an HTTPS connection and
answers with an HTTP-to-HTTPS redirect, which loops through the Gateway
indefinitely.

## Fix

Added `_PROXY_CONTEXT_HEADERS` and strip it in `_forward_headers`, next to
the existing hop-by-hop and auth-header stripping. `X-Forwarded-For` is
left untouched, since some deployments rely on it to see the real client
IP on the upstream side.

## Verification

- New regression tests in `TestForwardHeadersProxyContextStrip` fail
  against the pre-fix `_forward_headers` (each proxy-context header leaks
  through) and pass after the fix, including the issue's own repro shape
  (`X-Forwarded-Proto: http` with an HTTPS-configured upstream) and a
  check that `X-Forwarded-For` is still forwarded.
- Full `tests/auth_server/` suite (755 tests) passes.
- `ruff check`, `ruff format --check`, and the repo's `pre-commit` hooks
  (bandit, mypy, detect-secrets) pass on the changed files.
- Not verified: an end-to-end run against a real TLS-terminating ingress
  in front of both the Gateway and the upstream; the fix and tests are at
  the header-forwarding unit level where the leak originates.

Fixes #1625

### DIFF
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -6299,13 +6299,30 @@ async def _read_bounded(
 # servers that need an upstream credential use the egress vault.
 _INTERNAL_INGRESS_RELAY_SERVERS: frozenset[str] = frozenset({"airegistry-tools"})
 
+# Proxy-context headers describe the client -> Gateway hop (scheme, host, port,
+# path prefix as seen by nginx/an ingress LB) and must not be copied onto the
+# separate Gateway -> upstream MCP server hop; an upstream that trusts them can
+# see a stale scheme and issue a redirect loop. X-Forwarded-For is deliberately
+# excluded: some deployments rely on it to see the real client IP upstream.
+_PROXY_CONTEXT_HEADERS: frozenset[str] = frozenset(
+    {
+        "forwarded",
+        "x-forwarded-host",
+        "x-forwarded-port",
+        "x-forwarded-proto",
+        "x-forwarded-prefix",
+        "x-original-uri",
+        "x-original-url",
+    }
+)
+
 
 def _forward_headers(
     incoming: dict[str, str],
     relay_authorization: bool = False,
 ) -> dict[str, str]:
     """Copy incoming request headers to the upstream, stripping hop-by-hop and
-    proxy-hint headers so httpx can set them correctly for the connection.
+    proxy-context headers so httpx can set them correctly for the connection.
 
     Ingress-auth policy (issue #1266): Cookie is ALWAYS stripped (never
     forwarded to any upstream). Authorization and X-Authorization are also
@@ -6322,6 +6339,8 @@ def _forward_headers(
             continue
         if lower == "x-upstream-url":
             # Never leak this internal routing header to the upstream.
+            continue
+        if lower in _PROXY_CONTEXT_HEADERS:
             continue
         if lower in ("x-body", "x-body-uninspectable"):
             # Gateway-internal body-capture headers set by capture_body.lua for
--- a/tests/auth_server/unit/test_server.py
+++ b/tests/auth_server/unit/test_server.py
@@ -5031,6 +5031,57 @@ class TestForwardHeadersIngressStrip:
         assert "authorization" not in {k.lower() for k in out}
 
 
+class TestForwardHeadersProxyContextStrip:
+    """Ingress proxy-context headers (issue #1625) describe the client -> Gateway
+    hop and must not leak onto the separate Gateway -> upstream MCP server hop,
+    or an upstream that trusts them can see a stale scheme and redirect-loop.
+    """
+
+    def _forward(self, incoming):
+        from auth_server.server import _forward_headers
+
+        return _forward_headers(incoming)
+
+    def test_strips_each_proxy_context_header(self):
+        for name in (
+            "X-Forwarded-Proto",
+            "X-Forwarded-Host",
+            "X-Forwarded-Port",
+            "X-Forwarded-Prefix",
+            "Forwarded",
+            "X-Original-URI",
+            "X-Original-URL",
+        ):
+            out = self._forward({name: "x", "Accept": "y"})
+            assert name.lower() not in {k.lower() for k in out}, name
+            assert out.get("Accept") == "y"
+
+    def test_case_insensitive(self):
+        out = self._forward({"x-forwarded-proto": "http", "x-original-url": "u"})
+        assert "x-forwarded-proto" not in {k.lower() for k in out}
+        assert "x-original-url" not in {k.lower() for k in out}
+
+    def test_stale_ingress_scheme_does_not_reach_the_egress_hop(self):
+        """The issue's own repro: an HTTPS ingress request that arrives carrying
+        a stale X-Forwarded-Proto: http must not forward it to an HTTPS upstream."""
+        out = self._forward(
+            {
+                "X-Forwarded-Proto": "http",
+                "X-Forwarded-Host": "gateway.example.com",
+                "X-Forwarded-Prefix": "/my-server",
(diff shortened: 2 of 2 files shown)


auth_server/server.py
```diff
@@ -6299,13 +6299,30 @@ async def _read_bounded(
 # servers that need an upstream credential use the egress vault.
 _INTERNAL_INGRESS_RELAY_SERVERS: frozenset[str] = frozenset({"airegistry-tools"})
 
+# Proxy-context headers describe the client -> Gateway hop (scheme, host, port,
+# path prefix as seen by nginx/an ingress LB) and must not be copied onto the
+# separate Gateway -> upstream MCP server hop; an upstream that trusts them can
+# see a stale scheme and issue a redirect loop. X-Forwarded-For is deliberately
+# excluded: some deployments rely on it to see the real client IP upstream.
+_PROXY_CONTEXT_HEADERS: frozenset[str] = frozenset(
```
omrishiv, line 6307: Some others that can be added
```
x-forwarded-ssl,
x-forwarded-scheme,
x-cloudfront-forwarded-proto
```

omrishiv (approved):
Hi @AmirF194 , thank you for this. this looks great; we can add a follow up to add some others listed below

AmirF194:
Thanks for the quick review! Good call on those extra headers, I will look at picking that up as a follow-up.

cccs-eric:
Thanks for you both.  Will benefit from this fix.

PR #1657, opened 2026-08-19
fix(keycloak): pass KEYCLOAK_EXTERNAL_URL to the prebuilt registry, warn when it cannot work

Evolver-sweden:
Refs #1649

## Correction to the issue first

I filed #1649, and part of it was wrong. `docker-compose.yml` already passes `KEYCLOAK_EXTERNAL_URL` to the registry service, and has since d8543fe ("fix: Add KEYCLOAK_EXTERNAL_URL to registry service", #681, March 2026). The Podman stack and the Helm chart (via `oauthProviderSecretName`) carry it too.

Our deployment was built from a `docker-compose.yml` carried over from a 1.0.x release and never re-synced, so it predates that fix. The symptom was real, the diagnosis of *our* environment was not. Apologies for the noise.

The report's nginx observation is also not a bug: `registry/core/nginx_service.py` builds the Keycloak upstream from `KEYCLOAK_URL`, but that hop is container-to-container and must stay internal. The issue itself said as much.

Two things do remain, and this PR addresses them.

## 1. `docker-compose.prebuilt.yml` still omits it

The registry service there does not receive `KEYCLOAK_EXTERNAL_URL`. That is the quick-start path with prebuilt images, so it is the file a new operator is most likely to run.

Both the registry and the auth server build OAuth discovery documents from this value. Without it, `_create_keycloak_provider` falls back to `KEYCLOAK_URL`:

```python
# auth_server/providers/factory.py
keycloak_external_url = os.environ.get("KEYCLOAK_EXTERNAL_URL", keycloak_url)
```

so the service advertises the internal Docker DNS name. The response is a well-formed metadata document, nothing errors server-side, and external MCP clients fail discovery with a generic OAuth error.

Added the variable, plus a compose parity test covering every discovery-serving service in every compose file, so a third instance cannot land silently. There have now been two (#1378 for Podman, this one for prebuilt).

## 2. The fallback is silent, and a parity test cannot catch every case

A compose parity test does not help the case we actually hit: an operator running a compose file carried over from an older release. Nothing in the repo can assert on a file that is not in the repo. The only place that can catch it is the process that reads the value at startup.

So `_create_keycloak_provider` now says what is wrong instead of leaving it to be inferred from a client-side error:

| Configured value | Level |
|---|---|
| `KEYCLOAK_EXTERNAL_URL` unset | WARNING |
| single-label host (`keycloak`, i.e. container DNS) | WARNING |
| private, link-local or reserved address | WARNING |
[...]

Issue #1649, opened 2026-08-18
Registry does not receive KEYCLOAK_EXTERNAL_URL, so authorization server metadata and generated nginx config use the internal Docker URL


Evolver-sweden:
Body

Possible duplicate: #1378 reports the same symptom for docker-compose.podman.yml. That issue is closed. We hit it on the standard docker-compose.yml path with the registry service, so this may be a second instance of the same root cause rather than a regression. Happy to have this closed as a duplicate if the fix was meant to cover both.

Summary

With AUTH_PROVIDER=keycloak and KEYCLOAK_EXTERNAL_URL set in .env, the variable does not reach the registry container. As a result:

GET /.well-known/oauth-authorization-server advertises http://keycloak:8080, the internal Docker DNS name, which no external client can resolve or reach.
The generated nginx configuration is also built from the internal URL.

External MCP clients perform OAuth discovery, receive an unreachable issuer, and cannot complete the flow.

Environment
Version: 1.28.0, Docker Compose deployment
AUTH_PROVIDER=keycloak, Keycloak brokering an upstream OIDC IdP
KEYCLOAK_EXTERNAL_URL=https://gateway.example.com present in .env
TLS terminated by the bundled nginx
Steps to reproduce
Set AUTH_PROVIDER=keycloak and KEYCLOAK_EXTERNAL_URL=https://gateway.example.com in .env.
Start the stack.
curl -s https://gateway.example.com/.well-known/oauth-authorization-server
Expected

Metadata endpoints point at https://gateway.example.com.

Actual

Metadata advertises http://keycloak:8080. The registry log confirms which value it used when generating nginx config:

Keycloak hostname is 'keycloak', using netloc instead: keycloak
Using Keycloak configuration from KEYCLOAK_URL 'http://keycloak:8080': http://keycloak:8080
Keycloak upstream scheme is 'http' (not https); nginx will proxy the IdP token/JWKS/discovery traffic in plain

The auth-server, by contrast, does know the external URL:

Initializing Keycloak provider for realm 'mcp-gateway' at http://keycloak:8080 (external: https://gateway.example.com)

So the value is available to one service and not the other. Confirmed absent from the registry container's environment.

Suggested fix

[...]

PR #1688, opened 2026-08-26
fix(keycloak): rewrite every internal URL in the discovery document

Evolver-sweden:
Refs #1649

This is the second half of #1649. [#1657](https://github.com/agentic-community/mcp-gateway-registry/pull/1657) passed `KEYCLOAK_EXTERNAL_URL` to the registry service, which fixed the primary endpoints. The issue stayed open because the URL rewriting itself is incomplete, as I reported in [the follow-up comment](https://github.com/agentic-community/mcp-gateway-registry/issues/1649#issuecomment-5305118445) after verifying on 1.29.0. This closes that half.

## What was wrong

`authorization_server_metadata` rewrote the internal Keycloak URL onto the external host for **ten named fields**. Everything not on that list kept the internal Docker URL, and the nested `mtls_endpoint_aliases` object was not walked at all, so every key inside it leaked.

Still internal on a deployment with `KEYCLOAK_EXTERNAL_URL` correctly set:

| Field | |
|---|---|
| `check_session_iframe` | OIDC session management |
| `backchannel_authentication_endpoint` | CIBA |
| `pushed_authorization_request_endpoint` | PAR |
| `mtls_endpoint_aliases.*` | all eight nested endpoints |

The primary endpoints — `issuer`, `authorization_endpoint`, `token_endpoint`, `jwks_uri`, `registration_endpoint` — were correct, so the document looks right at a glance. A client that uses PAR, CIBA, mTLS client auth or OIDC session management is handed an unreachable internal host and fails in a way that is hard to attribute back to discovery.

An allowlist also means every field a future Keycloak release adds starts out broken by default, and nobody finds out until a client happens to use it.

## The fix

Walk the whole document. A value is rewritten because it points at the internal host, not because someone remembered to list its field name, so new fields and nested objects are covered without maintenance.

Two things fall out of doing it properly:

**The match is anchored at a URL boundary.** The old code used a bare `startswith`, so an internal base of `http://keycloak:8080` also matched `http://keycloak:80801` — a different port on the same host — and rewrote it. Only `/`, `?`, `#` or end-of-string now count as a boundary. This was a pre-existing defect; it is fixed here because the recursive walk would otherwise have spread it to more fields.

**New containers are built rather than mutating in place.** `_get_openid_configuration` is `lru_cache`d, so rewriting a nested dict in place would poison the cache for every later caller — including the internal-URL consumers that must keep pointing at the cluster host. A test pins this.

## Verification

Ten new tests in `TestKeycloakMetadataRewritesEveryUrl`. The primary one asserts the invariant over the whole document rather than over named fields:

```python
leaked = [s for s in self._all_strings(metadata) if self.INTERNAL in s]
assert not leaked
```

That is deliberately the shape of the fix: naming fields in the test would reintroduce the same blind spot the code had. The remaining tests cover each field from the report individually, the nested `mtls_endpoint_aliases`, non-URL values passing through untouched, the boundary case, a bare origin with no path, cache integrity, and the unchanged passthrough when no external URL is configured.

**Six of the ten fail against the pre-fix tree.** The four that pass do so correctly: non-URL values were never touched, a bare origin was already rewritten by the old `startswith` + `replace`, the old shallow copy did not corrupt the cache, and the passthrough path is unchanged.

Full suites: `tests/auth_server/unit/providers/test_keycloak.py` 53 passed, `tests/auth_server/unit/providers/` 263 passed. `pre-commit run --files <changed files>` passes with `SKIP=pytest-fast`, matching `.github/workflows/pre-commit.yaml`; Bandit and MyPy both cover `keycloak.py`.

## Scope

Keycloak only. It is the only provider that brokers an internal/external split — Cognito, Entra, Okta, Auth0 and PingFederate build their documents from a single cloud-facing host, so there is no internal URL to rewrite. I grepped for the same allowlist pattern elsewhere and found none.

Marked `Refs` rather than `Closes` so you can decide whether #1649 is now fully addressed; between this and #1657 I believe it is, but that call is yours.

### DIFF
--- a/auth_server/providers/keycloak.py
+++ b/auth_server/providers/keycloak.py
@@ -21,6 +21,11 @@ JWT_AUDIENCE = os.environ.get("JWT_AUDIENCE", "mcp-registry")
 # raises if it is missing rather than silently using a known-bad value.
 SECRET_KEY = os.environ.get("SECRET_KEY")
 
+# Characters that can legally follow an origin in a URL. Used to anchor the
+# internal-URL rewrite at a boundary so an internal base of
+# "http://keycloak:8080" cannot also match "http://keycloak:80801".
+_URL_BOUNDARY_CHARS: frozenset[str] = frozenset({"/", "?", "#"})
+
 
 logging.basicConfig(
     level=logging.INFO,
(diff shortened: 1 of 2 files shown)


omrishiv (approved):
@Evolver-sweden thank you so much for taking care of this, the fix looks good and appreciate the added tests

PR #1692, opened 2026-08-26
[Phase 3] RFC 8707 resource enforcement + gateway token proxy (IdP-signed MCP tokens)

### LINKED ISSUE #991: [Phase 3] RFC 8707 resource parameter enforcement + token proxy (closes #889)
# Sub-issue E: RFC 8707 `resource` parameter enforcement + token proxy (IdP-signed tokens)

**Parent:** #988
**Labels:** `enhancement`, `oauth`, `mcp-spec`, `authentication`, `credentials-management`, `subtask`
**Phase:** 3
**Depends on:** A
**Blocks:** B, C, D, H, closes #889 as part of acceptance
**Absorbs:** #614 Phase 2-3

---

## Summary

Stand up the `/oauth/token` proxy in the Registry Gateway, enforce [RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707) `resource` parameter semantics end-to-end, and switch MCP-data-plane token validation from the current self-signed 8h JWT to IdP-signed tokens. This is the phase that makes refresh and revocation actually work, and it closes #889 as a side-effect.

## Background

Today's auth flow:

- Operator clicks Connect in the UI → auth server mints a self-signed JWT with a hardcoded 8h TTL, no refresh token.
- Operator pastes it into `.mcp.json` / `mcp.json`.
- Assistant uses the Bearer token until expiry; there is no refresh and no revocation propagation from the upstream IdP.
- `MAX_TOKEN_LIFETIME_HOURS=24` is defined but never consulted (#889).
- `expires_in_hours` on the registration request is ignored.

Problems:

1. Revocation in the upstream IdP has no effect on the Registry until the 8h timer elapses.
2. Refresh is manual (re-paste).
3. Tokens are not scoped to a specific MCP server; any token hitting the gateway can reach any path the user is entitled to, with no resource-parameter check.
4. `#889` TTL enforcement can't be cleanly implemented on top of the self-signed path without also fixing the proxy.

### In scope

- **`POST /oauth/token`** as a proxy for `authorization_code` and `refresh_token` grants to the configured IdP:
  - Preserves PKCE `code_verifier` on the way through.
  - Preserves `resource` parameter on the way through (RFC 8707).
  - Enforces `resource` matches a Registry-registered MCP server canonical URI or a Registry-own resource.
  - On success, returns the IdP-signed `access_token` and `refresh_token` to the assistant; the gateway does not re-mint.
- **Inbound token validation** on MCP-facing paths:
  - Validate signature via the configured IdP's JWKS (cached with key rotation support).
  - Validate `aud` claim: MUST equal the canonical URI of the MCP server being called (i.e., the path the request is going to).
  - Validate `iss` claim against provider config.
  - Validate `exp` / `nbf`.
  - On any failure, 401 with `WWW-Authenticate` (sub-issue A's middleware takes care of the header).
- **Retire self-signed 8h tokens** on the MCP data plane. The web-UI login flow can keep its own session cookies; MCP data plane moves to IdP-signed.
- **Per-server `expires_in_hours` at registration**: add a field on the `register` model, enforce `MAX_TOKEN_LIFETIME_HOURS` as the ceiling, honor lower values per server. Closes #889.
- **Revocation observability**: add a metric for `aud` mismatches and signature failures so operators can see revocation events working in real time.
- **Tests**:
  - Unit: `aud` mismatch, `iss` mismatch, expired, wrong signature, missing `resource` on `/authorize`, `resource` not in registered set, TTL clamp against `MAX_TOKEN_LIFETIME_HOURS`.
  - Integration: end-to-end PKCE against a local IdP (Keycloak in docker-compose), revocation in IdP → next request 401 within JWKS cache refresh interval.

### Out of scope

- CIMD consumer behavior on `/authorize` → sub-issue C.
- Token exchange (RFC 8693) grant type → sub-issue D.
- Entra v1 scope format nuances → sub-issue F.
- Domain B backend token handling → declined, see #988 §"What we considered and declined".

## Design notes

- Keep the proxy thin. It forwards to the configured IdP's `/token` endpoint and returns the IdP's response verbatim, except for:
  - Rejecting `resource` values that don't map to a Registry-registered server.
  - Clamping `expires_in` down to the per-server / registry max.
- `aud` enforcement is the single most important hardening gain here. Today the gateway accepts any signed JWT regardless of audience; after this lands, a token for `/server-a/mcp` fails against `/server-b/mcp`.
- JWKS cache TTL balances revocation latency vs. IdP load. Default to 5 minutes; operator-configurable.
- The switch off the self-signed path is a breaking change for any existing deployment that is still relying on the UI-pasted JWT. Ship behind a feature flag for one release; flip default in the next.

## Acceptance criteria

- [ ] `POST /oauth/token` proxies `authorization_code` and `refresh_token` grants to the configured IdP, preserving PKCE and `resource`.
- [ ] `resource` parameter is required on `/authorize` and `/token` for MCP-facing flows; rejected if not present or not in registered set.
- [ ] MCP-facing paths validate `aud` matches the path; mismatch returns 401 + `WWW-Authenticate`.
- [ ] JWKS cache honors key rotation (kid match + re-fetch on unknown kid).
- [ ] `expires_in_hours` in registration honored; `MAX_TOKEN_LIFETIME_HOURS` enforced as ceiling (closes #889).
- [ ] Metrics for: successful exchanges, `aud` mismatch count, signature failure count, TTL clamp count.
- [ ] Self-signed 8h token path retired on the MCP data plane behind a feature flag; default flipped in the release after this one.
- [ ] Integration test using Keycloak: authorization code + refresh + revocation all behave correctly.
- [ ] Docs page in [docs/](docs/) describing the new token model and the migration from self-signed tokens.

## Risks and open questions

- **Breaking change for operators.** The UI-paste workflow is going away for MCP paths. Communicate via release notes, one full release behind a flag before flipping default.
- **IdP quirks.** Entra v1 does not always include all the fields we want in JWTs. Sub-issue F explicitly handles the scope format; coordinate so scope mapping for Entra is unified.
- **JWKS downtime.** If the IdP's JWKS endpoint is down, valid tokens start failing. Mitigation: cache persists across JWKS fetch failures with a bounded staleness window; alert on sustained JWKS fetch failures.
- **`aud` format inconsistencies across IdPs.** Some IdPs use the resource URI, others use a client-id-like opaque string. Normalize in provider-specific adapters, not in the core validation.

## References

- [RFC 8707 Resource Indicators](https://datatracker.ietf.org/doc/html/rfc8707)
- [MCP 2025-06-18 authorization spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization)
- Issue #889 (TTL enforcement)
- Issue #614 (umbrella tactical)
- [auth_server/providers/entra.py](auth_server/providers/entra.py) — dual-audience pattern already present
- Recommendation doc §6 Phase 3: [.scratchpad/coding-assistant-oauth/recommendation-2026-05-04.md](.scratchpad/coding-assistant-oauth/recommendation-2026-05-04.md)

omrishiv:
# [Phase 3] RFC 8707 resource enforcement + gateway token proxy (IdP-signed MCP tokens)

Fixes: #991

## Summary

Stands up a `/oauth/token` proxy in the Registry Gateway, enforces [RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707) `resource` semantics end-to-end, and moves MCP-data-plane token validation from the self-signed 8h JWT to **IdP-signed tokens** — so refresh and revocation actually work. Gated behind two **default-off** flags; with both off, behavior is unchanged.

## Motivation

Today the Connect flow mints a self-signed 8h JWT with no refresh and no revocation propagation: upstream IdP revocation has no effect until the timer elapses, refresh is manual re-paste, tokens aren't scoped to a specific MCP server, and `MAX_TOKEN_LIFETIME_HOURS` / registration `expires_in_hours` are ignored. This phase fixes all four.

## What changed

**Feature flags (both default-off, fail-closed interlock)**
- `MCP_TOKEN_PROXY_ENABLED` — enables the gateway `/oauth/token` proxy.
- `MCP_IDP_SIGNED_TOKENS` — switches MCP data-plane validation to IdP-signed tokens. Requires the proxy to be enabled; the combination is validated at startup.

**auth-server**
- `POST /oauth/token`: proxies `authorization_code` and `refresh_token` grants to the configured IdP over the SSRF/DNS-rebinding-safe client, preserving PKCE `code_verifier` and the RFC 8707 `resource`. `resource` is **required and matched against the registered server set** when IdP-signed tokens are on (lenient otherwise). Only standard token fields are returned; upstream cookies/headers are never relayed. Responses carry `Cache-Control: no-store`.
- `/validate`: when IdP-signed tokens are on, rejects the gateway self-signed token on MCP data-plane paths at a single chokepoint, requires the per-server `aud` for servers advertising a per-server resource (keyed on egress mode via a cached internal metadata lookup), and enforces a per-server access-token lifetime ceiling. Non-MCP, session-cookie, federation, A2A, and browser-UI paths are unaffected.

**registry**
- Internal per-server metadata endpoint (egress mode + lifetime ceiling), reachable only by the auth-server; fails closed (serves last-known-good on outage, treats 404 as a definite negative, empty only on cold-cache outage).
- Repoints the advertised `token_endpoint` at the gateway when the proxy is enabled — **`authorization_endpoint`, `issuer`, and `jwks_uri` stay the IdP's**.
- Moves the per-server-resource predicate into a shared, dependency-free helper consumed by both services.

**providers**
- New shared, hardened JWKS cache (`auth_server/providers/jwks_cache.py`) backing all IdP providers (Keycloak, Entra, Cognito, Okta, Auth0, PingFederate): TTL cache keyed by `jwks_url`, coalesced fetch, coalesced + negative-cached unknown-`kid` refetch behind a cooldown (blunts a `kid`-spray JWKS DoS), and bounded stale-on-failure fallback. RS256 pinning and all signature/issuer/audience checks stay in the providers' `jwt.decode`.

**nginx**
- Exact-match `/oauth/token` location added to every server block in both reverse-proxy configs, proxying to the auth-server and forwarding the request body, not gated by `auth_request`. (Without it the advertised gateway token endpoint fell through to the registry catch-all and 404'd.)

**infra**
- Both flags wired across Docker Compose, Helm, and Terraform, with chart env assertions kept in sync.

## Security notes
- Default-off; enabling `MCP_IDP_SIGNED_TOKENS` without the proxy fails closed at startup.
- Token proxy strips upstream Set-Cookie/headers and returns only standard token fields; `no-store` on every `/oauth/token` response (including errors).
- `/validate` closes the session-cookie bypass: a non-bearer caller can no longer skip token checks on MCP paths.
- Per-server metadata lookup fails closed so per-server audience enforcement is never silently dropped.

## Testing
Unit + integration cover: proxy route (PKCE + `resource` passthrough/enforcement), self-signed rejection on MCP paths, `aud`/`iss`/`exp`/`nbf` failures, TTL clamp against `MAX_TOKEN_LIFETIME_HOURS`, the startup interlock, the shared JWKS cache (cooldown/bound/stale fallback), the internal metadata endpoint (fail-closed), register validation, the token-endpoint repoint, and an nginx routing guard asserting each `/oauth/token` block proxies to the auth-server, forwards the body, and is un-gated.


aarora79:
Thanks for the thorough work here — the analysis and implementation are solid. After review we've decided to solve the token-revocation problem differently: a gateway-side **per-user revocation denylist** (an admin kill-switch keyed on the token's `sub`, with a TTL), which gives instant, provider-agnostic revocation without the IdP-signed + DCR client-onboarding burden — a real blocker on no-DCR IdPs like Entra. Closing in favor of that approach; a new design issue will capture it. Appreciate the effort on #991.

PR #1697, opened 2026-08-27
Remove `/api/internal/*` from the public listeners

omrishiv:
### Summary

Completes the network isolation started in #1655. That PR moved the egress-token vend onto the dedicated `:8091` internal listener, but the public `location /api/` catch-all still proxied **every** `/api/internal/*` path — the vend, the virtual-server session store, and the service-management routes — straight to the app on the internet-facing 8080/8443 listeners. The app-level gates held (401/403), but the network reachability the internal listener was meant to remove was still there via the public `/api/` fallthrough.

This PR blocks the whole `/api/internal/` prefix on the public listeners and migrates the one remaining public caller to a user-authenticated endpoint.

### What changed

**nginx** (`docker/nginx_rev_proxy_http_and_https.conf`, `docker/nginx_rev_proxy_http_only.conf`)
- `return 404` for `^~ {{ROOT_PATH}}/api/internal/` in every public server block (8080 ×1 in http-only, 8080 + 8443 in http+https). `^~` outranks the plain `/api/` prefix.
- Removed the dead `/api/internal/register` key from the registration rate-limit classifier map + updated the surrounding comments.
- **Unaffected:** the egress vend (`/_egress_internal/egress-token` on `:8091`) and the session store (`/_internal/sessions/`, `internal;`) both `proxy_pass` directly to `127.0.0.1:7860`, so those upstream requests never re-enter public location matching.

**Operator scripts** (`cli/service_mgmt.sh`, `terraform/aws-ecs/scripts/service_mgmt.sh`)
- The failed-scan auto-disable now calls the public, user-authenticated `POST /api/servers/toggle` (`path` + `new_state=false`) with the gateway JWT already loaded from `.oauth-tokens/ingress.json`, instead of minting a SECRET_KEY service token for the internal `POST /api/internal/toggle`.
- `verify_csrf_token_flexible` skips CSRF for Bearer clients, so no CSRF token is needed.
- This keeps the operator scripts working over the public ALB from anywhere (no Service Connect / NetworkPolicy dependency).

**Tests** (`tests/unit/core/test_nginx_service.py`)
- New `test_internal_routes_blocked_on_public_listeners`: asserts the per-server 404 blocks, that no listener proxies the `/api/internal/` prefix, that the egress vend location survives, and that the dead register map key is gone.
- Fixed `test_conf_declares_rate_limit_zones` (dropped `internal/register` from the register-classifier tuple).

**Docs** (`docs/api-reference.md`)
- The Internal Admin Endpoints section now states the routes are not publicly reachable (public listeners 404 the prefix; tooling uses `/api/servers/*`; the vend + sessions keep their dedicated internal paths) and corrects the auth line to the single-use `SECRET_KEY`-signed service JWT.

### Testing

- `pytest tests/unit/core/test_nginx_service.py` — 118 passed.
- Both templates: brace-balanced; `bash -n` clean on both scripts.
- Live-verified on EKS deployment:
  - Public ALB: `/health` → 200; `POST /api/internal/{egress-token,toggle,register}`, `GET /api/internal/{sessions/x,list-groups}` → **404**.
  - auth-server → `registry:8091/_egress_internal/egress-token` → **401** (reachable + gated = working); `registry:8000/api/internal/egress-token` → 404 (public path blocked).

### Deployment notes

- **Depends on the #1655 Helm wiring being applied.** The auth-server must reach the vend on the internal listener, which requires:
  - `registry.egressAuth.enabled=true` → renders the `:8091` Service port + `egress-internal` containerPort (+ the `registry-egress-internal` NetworkPolicy), and
  - `auth-server` `EGRESS_REGISTRY_INTERNAL_URL=http://registry:8091` (chart default).
  If the auth-server is still pointed at `registry:8000` (the public `:8080` listener), the vend will 404 after this change. Verified/fixed live on `mcp-sub` by adding the `:8091` Service port and repointing the auth-server env.

### Backward compatibility

- The `/api/internal/*` FastAPI routes are unchanged and still served on the app loopback (gated by `validate_internal_auth`); they are simply no longer reachable through nginx on the public listeners. Service management moves to the `/api/servers/*` user-authenticated siblings, which already existed.
- No changes to the egress vend or session-store request paths.

### DIFF
--- a/cli/service_mgmt.sh
+++ b/cli/service_mgmt.sh
@@ -542,43 +542,39 @@ except Exception as e:
         echo ""
         echo "====Disabling the server===="
 
-        # Generate JWT token for internal auth using shared SECRET_KEY
-        if [ -z "$SECRET_KEY" ]; then
-            print_error "SECRET_KEY not set in environment - cannot disable server"
+        # Disable the just-registered (auto-enabled) server via the public,
+        # user-authenticated toggle endpoint, reusing the same gateway JWT the
+        # rest of this script uses (the ingress token from credentials-provider).
+        local token_file="$PROJECT_ROOT/.oauth-tokens/ingress.json"
+        local auth_token
+        auth_token=$(python3 -c "import json; d=json.load(open('$token_file')); print(d.get('access_token') or d.get('tokens', {}).get('access_token') or '')" 2>/dev/null)
+
+        if [ -z "$auth_token" ]; then
+            print_error "No gateway token at $token_file - cannot disable server (run credentials-provider/generate_creds.sh)"
         else
-            local auth_token
-            auth_token=$(python3 -c "
-from registry.auth.internal import generate_internal_token
-print(generate_internal_token(subject='cli-service-mgmt', purpose='toggle-service'))
-" 2>/dev/null)
+            # Set the server to disabled (false); it was auto-enabled at registration.
+            print_info "Calling toggle endpoint with: ${GATEWAY_URL}/api/servers/toggle"
+            print_info "Service path: $service_path"
 
-            if [ -z "$auth_token" ]; then
-                print_error "Failed to generate auth token - cannot disable server"
+            output=$(curl -s -w "\nHTTP_STATUS:%{http_code}" -X POST "${GATEWAY_URL}/api/servers/toggle" \
+                -H "Authorization: Bearer $auth_token" \
+                --data-urlencode "path=$service_path" \
+                --data-urlencode "new_state=false" 2>&1)
+
+            # Extract HTTP status code from response
+            http_status=$(echo "$output" | grep "HTTP_STATUS:" | cut -d':' -f2)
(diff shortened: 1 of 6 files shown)


aarora79:
Follow-up addressed in #1712: the review's fail-closed and doc items are done there rather than in this PR.

- The failed-scan auto-disable now fails closed. Both operator scripts (`cli/` and `terraform/aws-ecs/scripts/`) `exit 1` on a missing token or a non-200 `/api/servers/toggle`, so a server that fails its security scan is never silently left enabled.
- Documented the #1655 deploy dependency in `docs/egress-credential-vault.md` (auth-server must point `EGRESS_REGISTRY_INTERNAL_URL` at `registry:8091`, or the vend 404s after this change).
- Noted the `ROOT_PATH` bare-path case in the nginx `/api/internal/` block comments.

Live-verified on a running stack: `/api/internal/*` returns 404 on the public listener; `/api/servers/toggle` returns 401 without a valid token (fail closed, exit 1) and 200 with an admin token (exit 0).

PR #1704, opened 2026-08-31
Docs: optional TOA verify for registry diligence

dulrajnr:
## Summary

Docs-only. Optional offline `toa-verify` before approving a registry entry / enable.

- `docs/toa-optional-register-gate.md`
- `examples/toa-after-register.yml`
- Docs index pointer

Does not change gateway/registry runtime. No AgentStatus account required to verify.

## Test plan

- [ ] Docs clearly optional / diligence-style
- [ ] Example YAML gated on `toa.json`

Made with [Cursor](https://cursor.com)


omrishiv:
Thank you for this. Ultimately, it will be up to the end operators to decide if they want to add this to their workflow

PR #1707, opened 2026-08-31
feat(url-guard): allow operator-named trusted IdP hosts for egress OAuth token endpoints

### LINKED ISSUE #1706: Egress OAuth consent is impossible against a self-hosted IdP: credentialed-OAuth allowlist is hardcoded empty
## Summary

Since 1.29.0, per-user egress OAuth consent cannot work at all when the OAuth/OIDC provider is self-hosted on an internal network. `CREDENTIALED_OAUTH_PROFILE` requires every token endpoint to resolve exclusively to public addresses, and its allowlist is hardcoded empty with no setting able to relax it.

This blocks us from adopting 1.29.0. We are on 1.28.0 and had to roll the bump back.

## Reproduce

1. Configure the registry with an IdP whose hostname resolves to a private address (in our case a self-hosted Keycloak, RFC1918). This is the same IdP configured in `KEYCLOAK_URL`.
2. Configure a server with `egress_auth_mode: oauth_user` whose `egress_oauth` token endpoint is on that IdP.
3. Connect the account from the UI.

Consent fails with HTTP 500 and `token endpoint blocked by security policy`.

## Cause

In `registry/utils/url_guard.py`:

```python
def _credentialed_oauth_allowlist() -> _Allowlist:
    """Return the OAuth token-endpoint allowlist: deliberately empty.

    Credential-bearing token POSTs may include a client secret, refresh token,
    or user assertion. They must not inherit any operator proxy/skill bypass.
    The profile also enforces HTTPS at both structural validation and transport.
    """
    return _Allowlist()
```

The factory reads no settings whatsoever, so `SSRF_ALLOWED_HOSTS` and `SSRF_ALLOWED_CIDRS` do not reach it. That is clearly deliberate and the reasoning in the docstring is sound — a proxy-target bypass should not silently relax an endpoint that receives client secrets. But the net effect is that a self-hosted IdP is categorically unsupported for egress OAuth, with no opt-in of any kind.

I checked this is current: the factory is byte-identical at tag `1.29.0` and on `main`.

## Why we think this specific case deserves an opt-in

The gateway **already** makes credentialed calls to this exact same IdP for its own authentication, via `KEYCLOAK_URL` / `KEYCLOAK_CA_BUNDLE`. So the guard blocks egress OAuth to the very IdP the product requires the operator to configure and already trusts elsewhere.

The IdP is also categorically different from the other endpoints this guard protects. A federation peer or a `proxy_pass_url` is registrant-supplied and therefore attacker-influenceable, which is a strong argument for allowing no bypass at all. An IdP token endpoint is operator-configured infrastructure, named in static config by the same person who deploys the gateway.

Self-hosted Keycloak on an internal network, and Entra behind Private Link, both seem like mainstream deployment shapes for the per-user egress feature, so we suspect we are not the only ones who will hit this.

## Secondary issue: the error is near-undiagnosable

1.29.0 reports this as a generic `token endpoint blocked by security policy`. There is nothing in the message or logs pointing at the SSRF guard, the resolved address, or the fact that no setting can influence the outcome. It took reading `url_guard.py` to work out what was happening. Even if the policy decision stands, including the denial reason and the resolved category would save operators a lot of time.

## Proposed fix

PR to follow, adding `EGRESS_OAUTH_TRUSTED_IDP_HOSTS`: a dedicated, hosts-only, exact-match, empty-by-default allowlist that follows the existing `_skill_allowlist` / `_proxy_allowlist` pattern while staying deliberately narrower than both.

Design intent is to preserve every stated invariant of the current profile:

- it reads **only** its own setting, so `ssrf_allowed_hosts` / `github_extra_hosts` still cannot re-permit a token endpoint — the "must not inherit any operator proxy/skill bypass" property in the docstring is kept
- hosts only, no CIDRs, no wildcards, so each IdP must be named exactly
- `require_https` unchanged, and the hard-denied categories (cloud/workload credential, metadata, link-local, reserved, multicast) are untouched, since they are evaluated before the hostname relaxation
- empty by default, so no existing deployment changes behaviour

Happy to take this in a different direction if you would prefer something else — e.g. scoping the trust to the already-configured `KEYCLOAK_URL` host automatically rather than adding a new setting, or gating it behind an explicit "I accept a private IdP" flag. Wanted to raise it with a working patch rather than just a complaint.

go-faustino:
Fixes #1706.

## Problem

`CREDENTIALED_OAUTH_PROFILE` requires every egress OAuth token endpoint to resolve exclusively to a public address, and its allowlist is hardcoded empty:

```python
def _credentialed_oauth_allowlist() -> _Allowlist:
    """Return the OAuth token-endpoint allowlist: deliberately empty."""
    return _Allowlist()
```

The factory reads no settings, so `SSRF_ALLOWED_HOSTS` / `SSRF_ALLOWED_CIDRS` do not reach it. A self-hosted IdP is therefore categorically unsupported for per-user egress consent: connecting an account against an internal Keycloak fails with `token endpoint blocked by security policy` and nothing an operator can configure will change it. This blocks us from adopting 1.29.0.

The reasoning in the existing docstring is sound and I have tried not to weaken it. A `proxy_pass_url` or a federation peer is registrant-supplied and therefore attacker-influenceable, which justifies allowing no bypass at all. An IdP token endpoint is different in kind: it is operator-configured infrastructure named in static config, and the gateway is **already required** to make credentialed calls to that same IdP for its own authentication via `KEYCLOAK_URL` / `KEYCLOAK_CA_BUNDLE`. The guard currently blocks egress OAuth to the very IdP the product makes you configure.

## Change

Adds `EGRESS_OAUTH_TRUSTED_IDP_HOSTS`, following the existing `_skill_allowlist` / `_proxy_allowlist` pattern and deliberately narrower than both:

```python
@lru_cache(maxsize=1)
def _credentialed_oauth_allowlist() -> _Allowlist:
    """Return the OAuth token-endpoint allowlist: trusted IdP hosts only."""
    return _Allowlist(hosts=_parse_hosts(_get_settings().egress_oauth_trusted_idp_hosts))
```

Every invariant the profile documents today is preserved:

| Property | Status |
|---|---|
| Does not inherit proxy/skill bypass | **Kept.** Reads only its own setting; `ssrf_allowed_hosts`, `ssrf_allowed_cidrs` and `github_extra_hosts` still cannot re-permit a token endpoint. Covered by two tests. |
| CIDR ranges | **Not supported**, unlike the proxy profile. Hosts only, matched exactly, so no subdomain or parent inference and no wildcards. |
| `require_https` | **Unchanged**, at both structural validation and transport. |
| Metadata / credential / link-local / reserved / multicast | **Still hard-denied.** These are evaluated in `_ip_denial_reason` before the hostname relaxation, so naming a host cannot turn it into a credential-endpoint bypass. |
| DNS-rebinding pinning | **Unchanged.** A trusted host is still resolved, classified and pinned to a validated answer. |
| Existing deployments | **Unaffected.** Defaults to empty, which reproduces today's behaviour exactly. |

Also adds `@lru_cache(maxsize=1)`, consistent with the other settings-reading factories, and updates the module docstring's profile description.

### A note on the auth-server process

`_egress_upstream_allowlist` carries a comment about deliberately touching no settings so it stays importable where registry `Settings` is not built. That constraint does not apply here: `auth_server/egress_obo.py` is the only non-registry consumer of this profile, and the auth-server already imports `registry.core.config.settings` (`auth_server/server.py`). The standalone tooling the lazy-binding comment refers to uses the default empty-allowlist profiles, not this one. Worth a second opinion from a maintainer, since I am reading this from outside.

## Files

- `registry/utils/url_guard.py` — the factory, cache decorator, docstring
- `registry/core/config.py` — the new setting, empty by default
- `tests/unit/utils/test_url_guard.py` — tests plus the new `cache_clear()` in the autouse reset fixture
- `.env.example`, `docs/unified-parameter-reference.md` — operator documentation

I did not wire the setting into the Helm chart or Terraform. It is not a chart-reserved name, so `extraEnv` already covers Helm without changes. Glad to add explicit plumbing if you would prefer it as a first-class chart value.

## Tests

Added to `TestCredentialedOAuthProfile`:

- empty by default, and `require_https` still true
- the setting is read from its own field, lower-cased and whitespace-stripped, and contributes no CIDRs
- a trusted IdP resolving to RFC1918 is accepted (the actual bug)
- an untrusted private host is still rejected
- exact match only: `evil.idp.example.com`, `example.com` and `idp.example.com.evil.net` are all rejected when `idp.example.com` is trusted
- the metadata endpoint stays denied even for a trusted host
- `http://` is still rejected for a trusted host
- the transport pins to the resolved private IP and preserves the `Host` header
- `github_extra_hosts` cannot relax a token target (new)
- `ssrf_allowed_hosts` / `ssrf_allowed_cidrs` cannot relax a token target (existing test, retained)

Results:

- `tests/unit/utils/test_url_guard.py` — 156 passed
- `tests/unit/utils` + `tests/unit/core` — 1467 passed, no regressions
- `ruff format --check` clean on all touched files; `ruff check` introduces no new findings (verified by diffing against the same files on `main`)

The pre-existing `test_profile_has_empty_allowlist_and_requires_https` is renamed to `test_profile_is_empty_by_default_and_requires_https` and now patches settings explicitly, since the assertion is about the default rather than about the allowlist being unconditional.

Note for reviewers: the `_settings()` test helper is a `MagicMock`, so the new attribute had to be added there — otherwise any existing test that patches settings and exercises this profile would receive a `MagicMock` where a string is expected.

## Open to direction

If you would rather not add a setting, two alternatives I would be equally happy to implement: derive the trust from the already-configured `KEYCLOAK_URL` host automatically, or gate it behind an explicit opt-in flag acknowledging a private IdP. I went with an explicit named-hosts list because it is the least magical and the easiest to audit.

Separately, and not addressed here: the failure surfaces as a generic `token endpoint blocked by security policy` with nothing pointing at the SSRF guard or the resolved address, which made this genuinely hard to diagnose. Happy to open a follow-up PR improving that message if useful.


omrishiv (changes requested):
Hi @go-faustino thanks so much for the PR and sorry about the issue its caused you. Two things hopefully you can address and it should be good to go:

1) `registry/api/config_routes.py` can you add this to the `CONFIG_GROUPS` please?
2) can you propagate this through to docker/helm/terraform?

With these two, we should be good to merge; thank you!

go-faustino:
Thanks @omrishiv — both addressed in `52de9d69`, and the branch is rebased onto current `main` (it was 12 commits behind; clean rebase, no conflicts).

**1) `CONFIG_GROUPS`**

Added to the existing `egress_credential_vault` group, right after `egress_obo_allowed_audiences`:

```python
("egress_oauth_trusted_idp_hosts", "Trusted IdP Hosts", False),
```

Marked non-sensitive since the value is hostnames only. I left a short comment noting that entries still receive client secrets and refresh tokens, so it's worth an operator's attention even though it isn't masked.

**2) Propagation**

Followed the `EGRESS_OBO_ALLOWED_AUDIENCES` precedent in each target:

- **Docker** — `docker-compose.yml`, `.podman.yml`, `.prebuilt.yml`
- **Helm** (`charts/registry`) — `values.yaml` (documented, defaults `""`), `configmap-egress.yaml`, `reserved-env-names.txt`, plus two `helm unittest` cases
- **Terraform** (`terraform/aws-ecs`) — root and module `variables.tf`, the registry task env in `ecs-services.tf`, the module pass-through in `main.tf`, and `terraform.tfvars.example`

The ConfigMap key is emitted conditionally, so an unset value renders no key at all rather than an empty string. Verified both ways:

```
# set
EGRESS_OAUTH_TRUSTED_IDP_HOSTS: "keycloak.internal.example.com"
# unset -> key absent
```

**One judgement call worth flagging**

I moved the `docs/unified-parameter-reference.md` row out of the SSRF/a2a table and into the egress-vault table (Group 31), because it now has real `helm`/`terraform` names and the old row said `n/a (set via extraEnv)` in both columns, which this PR makes untrue. The SSRF section keeps a one-line pointer, since *why* a token endpoint is deliberately not covered by `SSRF_ALLOWED_HOSTS` is most useful to a reader standing in that section. Happy to put it back if you'd rather keep it adjacent to the SSRF settings.

**Verification**

- 264 Python tests pass (`test_url_guard.py`, `test_config.py`, `test_config_export.py`)
- 87 chart tests pass across all 11 suites, including the 2 new cases
- `scripts/validate-extra-env.sh` passes
- `terraform fmt -check` clean on all four files I touched

Default remains empty on every path, so there's no behaviour change on upgrade — an operator has to name a host to permit anything.

omrishiv (approved):
@go-faustino thank you for taking care of the changes. I'll open a follow up PR to add the `EGRESS_OAUTH_TRUSTED_IDP_HOSTS` to the `auth-server`. As it is, this should work for 3LO, but there might be an issue with OBO where we need to also set this. With #1678 keycloak will support OBO, so we should add this.

Issue #1695, opened 2026-08-27
1.28.0: egress vend keys on username/email while consent writes the OIDC sub, so all bearer clients get an empty tools/list


anjosluc:
## Summary

`1.28.0` changed the egress **consent-write** path to key the per-user token vault on the canonical OIDC `sub` (#1491), but the **vend-read** path still resolves to the username/email for bearer callers. The two halves therefore disagree, and every per-user egress server returns an empty `tools/list` to IDE clients (Cursor, Claude) after upgrading from `1.27.1`.

On `1.27.1` both sides resolved to the email, so they agreed and egress worked. The fix corrected one side only, which converts a consistent scheme into a divergent one — it presents as a regression introduced by the fix.

## Impact

- Every `egress_auth_mode: oauth_user` server returns `EMPTY tools/list` for bearer/IDE clients.
- Reconnecting never helps: consent faithfully rewrites the OIDC-sub bucket, which nothing reads.
- Ingress auth is unaffected and looks healthy (`Token validation successful`, `Access granted`), so it does not resemble an auth failure.
- No log line is emitted on the miss — `get_valid_token` returns `None` silently on a store miss, and neither the `non-per-user auth_method` nor the `vend REFUSED` branch fires.

IdP: Keycloak. Deployment: separate `auth-server` and `registry` containers, image `1.28.0`.

## Evidence

Consent writes the OIDC-sub bucket (verified directly in the vault):

```
oauth2 / d1135345-…-992406af3a7f / custom / /trino-dev   status=active  updated=12:31:34Z
```

26 seconds later the same server misses:

```
mcp_proxy: egress server=trino-dev/mcp has no token; returning EMPTY tools/list
```

Replaying the store lookup in-process shows the data is findable and which key shape fails:

```
auth_method=oauth2    server_path=/trino-dev   -> HIT   (status=active)
auth_method=oauth2    server_path=trino-dev    -> MISS
auth_method=keycloak  server_path=/trino-dev   -> MISS
```

Writing a copy of the same token under the **email** key made the tools appear immediately, confirming the vend resolves to the email:

```
oauth2 / <user>@<domain> / custom / /trino-dev    -> tools/list now populated
```

The confirmed OIDC `sub` in Keycloak for that user is `d1135345-…`, i.e. identical to the bucket consent wrote — so this is not an identity mismatch, it is a resolution fallback.

## Mechanism

`auth_server/internal_request_token.py` stamps the claim but tolerates an empty value:

[...]

PR #1709, opened 2026-08-31
fix(egress): stop resolving the egress vault id across identifier namespaces

go-faustino:
A user completes the browser consent for a per-user egress MCP server, the consent succeeds, and then every per-user egress server returns an empty `tools/list` to their IDE client. Nothing is logged. The vend simply returns no token and the caller sees zero tools, so there is no signal anywhere pointing at the cause.

The cause is that two code paths resolve the per-user egress vault key through different identifier namespaces, so one human ends up with two parallel vault identities. The consent-write path resolves the OIDC `sub`. The vend path, when the caller is a gateway-minted self-signed token that carries no `egress_user` claim, falls through to that token's `sub` — which is the login username by construction, and on Keycloak the login username is the user's email address. The token is written to one bucket and read from another, and the miss is indistinguishable from "never consented".

This is issue #1695. That issue was closed as completed with no linked fix; a maintainer may well have closed it alongside related work, but the code on `main` still exhibits the problem, which is what this change addresses.

This builds on two already-merged PRs rather than reverting either of them. #1491 established that the vault keys on the OIDC `sub` across both consent and vend. #1530 established that the `egress_user` claim is only trusted from a `self_signed` token this gateway minted itself, so an external issuer cannot inject a vault key. Both are correct and both are preserved here — the trust gate from #1530 is untouched. The remaining gap is what happens when the gate applies but the claim is absent: resolution kept walking down a chain whose remaining entries are, for that specific token type, the login username.

## The resolution chain

Before, in `_canonical_egress_user`:

```
egress_user (self_signed only) -> data.sub -> data.subject -> sub -> username -> ""
```

After:

```
self_signed:      egress_user -> ""
everything else:  data.sub -> data.subject -> sub -> username -> ""
```

The second line is byte-for-byte what it was. Only the `self_signed` case changes, and only when the claim is missing. A correctly minted self-signed token still resolves to its `egress_user`, and non-OIDC callers with only a `username` still key on the username exactly as documented.

The vend in `registry/api/egress_auth_routes.py` changes from

```python
sub = claims.get("egress_user") or claims.get("sub") or ""
```

to reading only `egress_user`, and then, after the existing per-user check, refusing an empty value with a warning that names the server and the reason.

## On the non-per-user `sub` fallback

The fallback is gone entirely rather than retained for non-per-user callers. Reading the surrounding code, `sub` is only ever consumed on per-user branches: the very next check returns `consent_required=True` for any caller whose `auth_method` is not per-user, before `sub` is used for anything. A non-per-user fallback would therefore be dead code that only invites a future reader to reintroduce the namespace crossing.

## Deliberate behaviour change

Tokens minted before the `egress_user` claim existed will now be refused instead of silently keyed on the wrong bucket. This is intentional. Those tokens have a short TTL — the UI states 8 hours — so the blast radius is bounded to one token lifetime after upgrade, and the next mint carries the claim, which makes the condition self-healing. Crucially, the refusal is now visible: it logs a warning naming the server, where previously the same situation produced an empty tool list and complete silence. Trading one bounded token lifetime of visible, self-correcting refusals for the removal of a silent permanent misroute is the better failure mode, so this does not add a configuration flag to opt out.

## What this does not change

`registry/api/egress_auth_routes.py` around line 173, in `_resolve_target_principal`, has a third instance of the same pattern: `user_context.get("egress_user") or user_context.get("username") or ""`. It is left alone deliberately. That line is on the PAT submission and consent principal path, not the vend, and it resolves against a session-derived `user_context` rather than an mcp-proxy token, so the reasoning about which namespace the remaining candidate belongs to is different and needs its own analysis. Changing it speculatively risks breaking PAT flows for a defect that has not been demonstrated there. It is worth a follow-up on its own.

## Tests

`tests/auth_server/unit/test_egress_mint.py` gains five cases in the existing `TestCanonicalEgressUserPaths` suite: a self-signed token with no `egress_user` but a username-shaped `data.sub` must resolve to `""`; the same with `data.subject`, top-level `sub` and `username` present, since none of those is an OIDC sub on that token type either; a regression guard that a self-signed token with a valid `egress_user` still resolves to it; a regression guard that non-self-signed OIDC callers still resolve through `data.sub` and `data.subject`; and a regression guard that a non-OIDC caller with only a `username` still keys on the username.

`tests/unit/egress_auth/test_internal_egress_token_route.py` gains a case asserting that a per-user caller with no `egress_user` claim gets `consent_required=True` with no token, no service call, and a warning naming the server, using `caplog`.

The shared `_claims()` helper in `test_internal_egress_token_route.py` and `test_pat_vend.py` now includes `egress_user` alongside `sub`, so those fixtures reflect the shape of a token the current auth-server actually mints. Those suites test upstream binding, path normalisation and PAT vending, not the removed fallback, and all their assertions are unchanged.

Counts from a local run with the repository's pytest:

- `tests/auth_server/unit/test_egress_mint.py` plus `tests/unit/egress_auth/`: 307 passed before, 313 passed after.
- Three of the six new cases fail against the unmodified source and pass with it; the other three are regression guards that pass in both states, which is their purpose.
- Full run of `tests/auth_server/` and `tests/unit/`: 7344 passed, 1 failed, 8 skipped. The single failure is `tests/unit/audit/test_routes.py::TestAdminOnlyAccess::test_rejects_non_admin_users`, a Hypothesis `DeadlineExceeded` flake (the test took 5008 ms against a 200 ms deadline, reported as `FlakyFailure` after it did not reproduce on the retry). It is unrelated to this change, touches no egress code, and passes on its own both with and without these edits.

`docs/egress-credential-vault.md` is updated to record both fail-closed points in the section describing the canonical `user_id`.


go-faustino:
Field evidence that this is still worth merging, and that the failure mode it prevents is now considerably worse than when it was written.

We hit this on a production-adjacent deployment running **1.30.0**, and both halves of the namespace crossing described above are still present on that tag: the auth-server's `_canonical_egress_user` still walks the fallback chain past `egress_user`, and the registry's vend still does

```python
sub = claims.get("egress_user") or claims.get("sub") or ""
```

so the two-bucket split reproduces exactly as described. Our Keycloak login username is the user's email, so the consent write lands in the OIDC-`sub` bucket and the vend reads the email bucket. Closing #1695 did not fix it.

## 1.30.0 destination binding makes this permanent instead of silent

The PR description argues the deliberate behaviour change is acceptable because the exposure is bounded to one ~8h token lifetime and self-heals on the next mint. That argument holds, but the *counterfactual* has changed, and strongly in this PR's favour.

1.30.0 added destination binding: `get_valid_token` refuses a credential whose `bound_upstreams` does not cover the server's registered upstream. A credential written before 1.30.0 has no such field, so it deserialises to an empty list and is refused permanently, regardless of `status`.

Compose that with the two-bucket split and you get a state that a user **cannot** recover from through the UI:

1. The bucket the vend reads holds a stale pre-1.30.0 credential: `active`, empty bound set, permanently refused.
2. The user disconnects and reconnects in Connected Accounts. That writes a correctly-bound credential — into the *other* bucket.
3. The vend still reads the first bucket. Still refused. Still zero tools.

We verified this end to end: a full disconnect/reconnect cycle does not fix it, and it is not provider-specific — we reproduced it on two unrelated per-user egress servers. So on 1.30.0 the defect this PR fixes is no longer "silent zero tools until the token rotates". It is a **permanent outage that survives the only remediation a user has available**, because re-consent operates on a bucket that is never read.

That inverts the risk comparison in the "Deliberate behaviour change" section. Refusing pre-`egress_user` tokens for one bounded, visible, self-healing token lifetime is clearly the better failure mode than what 1.30.0 currently does, and destination binding removes the "it self-heals eventually" escape hatch that made the status quo tolerable.

## Why the `bound_upstreams` fix alone is not sufficient

Worth noting for anyone triaging these together: healing the binding does **not** substitute for this PR. Re-consent produces a correctly-bound credential every time — it just deposits it in the wrong bucket. As long as the write and the read resolve through different identifier namespaces, no amount of binding correctness is reachable by the vend. Fixing the namespace crossing is the load-bearing change; the binding behaviour merely converts the symptom from silent to permanent.

We are carrying a CronJob that mirrors credentials across both buckets as a stopgap, and we have just had to teach it that an empty `bound_upstreams` means "dead, replace me" rather than "already working, skip" — otherwise it refuses to overwrite the stale copy and the outage persists indefinitely. That is a lot of machinery to work around one fallback chain, and it retires entirely if this merges.

Happy to test a rebase of this branch against current `main` on our deployment and report back, if that is useful for getting it reviewed.

go-faustino:
CI is red on this head (`3caa6ee7`), but the failure is not from this change.

`Test (Python 3.14)` reports `1 failed, 8448 passed, 57 skipped`. The single failure is:

```
FAILED tests/unit/core/test_telemetry.py::TestLockAcquisition::test_acquire_lock_file_based_always_succeeds - assert False is True
```

Every other check passes, including `Auth Server Tests (Python 3.14)` — the job that covers `_canonical_egress_user`. This branch touches `auth_server/server.py`, `registry/api/egress_auth_routes.py`, `docs/egress-credential-vault.md` and three egress test files; nothing under `registry/core/telemetry.py` or its tests.

The test is order-dependent on shared CI state. It patches only `registry.core.telemetry.settings`, and `_acquire_telemetry_lock` never branches on `storage_backend` — it goes straight to `get_documentdb_client()`. The two sibling tests in the same class mock that client; this one does not, so it issues a real `find_one_and_update` against the CI MongoDB service container. With `upsert=False` and a 60s staleness window, that returns a document only when `_id: "telemetry_config"` already exists *and* `last_startup_sent_at` is absent, null, or older than the cutoff — which depends on what the other xdist workers did to the shared database first. The assertion is an unconditional `is True`, so the outcome is a race.

That also explains why it is usually green: with no reachable MongoDB the call raises and the `except` branch returns `True`. Run the test on its own and it passes (0.3s locally, no mongod listening). #1766, based on the same `main` commit `d22270f1`, had this exact job pass.

I do not have rerun rights here. I am happy to either push a content-identical amend to trigger a fresh run, or open a separate PR that mocks `get_documentdb_client` in that test (or seeds the document / uses `upsert=True`), if you would rather see the flake fixed first.

For the record on this branch: it is rebased onto `d22270f1`, single commit, no merge commits, and currently mergeable.

registry/api/egress_auth_routes.py
```diff
@@ -688,6 +692,20 @@ async def vend_egress_token(
         logger.info("egress vend: non-per-user auth_method %r -> consent", auth_method)
         return EgressTokenResponse(consent_required=True)
 
+    # A per-user principal with no resolved vault id: refuse rather than guess.
+    # Reachable for tokens minted before the ``egress_user`` claim existed; those
+    # expire within one short token TTL, and the next mint carries the claim, so
+    # this is self-healing. Log it -- the previous cross-namespace fallback made
+    # this failure completely invisible.
+    if not sub:
```
omrishiv: I think this needs to move below the egress_mode=obo block

registry/api/egress_auth_routes.py
```diff
@@ -677,9 +677,13 @@ async def vend_egress_token(
...
-    # when an IdP omits preferred_username from access tokens, e.g. Entra). Fall
-    # back to ``sub`` for tokens minted before this claim existed.
-    sub = claims.get("egress_user") or claims.get("sub") or ""
+    # when an IdP omits preferred_username from access tokens, e.g. Entra). There
+    # is deliberately NO fallback to ``sub``: the mcp-proxy token's ``sub`` is the
+    # login username, a different identifier namespace from the OIDC sub the
+    # consent path wrote, so falling back would key the vault on a second identity
+    # for the same human (silent miss, 0 tools). A per-user caller without the
+    # claim fails closed and is logged below.
+    sub = claims.get("egress_user") or ""
```
omrishiv, line 686: dropping `sub` here means auth0, github, google won't have access to the egress_user. we may need to update these providers a proper mapping

omrishiv (changes requested):
Hi @go-faustino, thank you for the PR. I think this might introduce an issue with some of the providers as written. Please take a look

registry/api/egress_auth_routes.py
go-faustino: Good catch, thanks. The guard ran before the mode was known, so it refused `obo_exchange` too, and that mode never touches the vault. Fixed in 3cbd8064.

I didn't move it literally below the OBO block, because the `pat` branch sits above that block and does key the vault on the caller id, so it would have lost the guard. The check now runs after the server lookup and upstream binding check, just before `pat`, and skips `obo_exchange`:

```python
if not sub and egress_mode != "obo_exchange":
```

Tests cover both sides: an OBO directive still vends when there's no `egress_user` (`test_obo_directive_vends_without_egress_user_claim`), and a `pat` vend with no `egress_user` is still refused (`test_missing_egress_user_claim_consents_without_vend`).

registry/api/egress_auth_routes.py
go-faustino: You're right. Those three providers never persisted a session subject, so before this PR both consent and vend fell back to the username. With the fallback gone, their bearer tokens would all have been refused. Fixed in 3cbd8064 by giving each one a proper subject:

- **Auth0**: `extract_user_from_tokens` never copied the id_token `sub`. It now returns it as `subject`.
- **GitHub / Google**: their userinfo endpoints (`/user`, `oauth2/v2/userinfo`) have no `sub` field. `map_user_info` now reads the subject from a new per-provider `subject_claim` (default `"sub"`, so every other provider behaves as before), and both set it to `"id"`. GitHub's numeric id survives login renames, and Google's v2 `id` is the OIDC sub. The value is stringified, since GitHub's is an integer.
- If a login still yields no subject (for example, a custom provider in `oauth2_providers.yml` without `sub`), the callback now logs a warning, so the problem shows up at login instead of as 0 tools.

One migration cost: Auth0/GitHub/Google users who consented before this change have grants keyed on the username. They'll need to sign in again and reconnect each egress server once. I documented this in `docs/egress-credential-vault.md` and added `subject_claim` to the provider table in `docs/configuration.md`. I preferred this over keeping a username fallback for those providers, because that fallback is the cross-namespace keying this PR removes.

go-faustino:
@omrishiv thanks for the review. Both points are addressed in 3cbd8064, with replies in each thread: the vend guard no longer blocks `obo_exchange`, and Auth0, GitHub, and Google now persist a proper session subject. The one trade-off is that users of those three providers reconnect each egress server once; the PR description covers it. Ready for another look when you have time.

omrishiv (approved):
@go-faustino thank you for addressing the PR comments, I think it's good to merge as is, but would ask if you're able to update the comment in https://github.com/agentic-community/mcp-gateway-registry/blob/main/tests/auth_server/unit/test_token_mint_session_reconcile.py#L112-L113

go-faustino:
@omrishiv thanks for the approval. Updated that comment in 3d12441c: it now says a session with no persisted subject has no canonical egress id, so the mint omits `egress_user` and the per-user vend refuses the token instead of falling back to the username. Comment-only change; that test file passes (15 passed).

Issue #993, opened 2026-05-09
[Phase 4] CIMD consumer: accept CIMD URL as client_id on /authorize


aarora79:
# Sub-issue C: Accept CIMD URLs as `client_id` on the Registry's `/authorize` endpoint

**Parent:** #988
**Labels:** `enhancement`, `oauth`, `mcp-spec`, `authentication`, `subtask`
**Phase:** 4
**Depends on:** A, E, B
**Blocks:** none

---

## Summary

When the Registry acts as an OAuth authorization server (or as a token-proxy in front of one), accept a CIMD URL as the `client_id` parameter on `/authorize` and `/oauth/token`. Fetch the document, cache it, validate the request against it (`redirect_uris`, `grant_types`, `scope`), and issue codes / tokens against the URL-as-identity.

This is the **consumer** side of the CIMD pattern. Publisher is sub-issue B.

## Background

In the CIMD pattern the client does not pre-register with the authorization server. On the first `/authorize` request, it presents its metadata URL as `client_id`. The AS fetches the document, caches it, and from that point treats the URL as the client identity.

Claude Code already does this automatically. Once the Registry hosts its own MCP-facing OAuth surface (via the token proxy from sub-issue E), we need to consume CIMD so we're not forcing Claude Code users to also pre-register somewhere.

### In scope

- **`/authorize` handling**: when `client_id` looks like a URL (has a scheme + host), treat it as a CIMD URL.
  - Fetch over HTTPS only. Reject `http://` outside local-dev allowlist.
  - Timeout + retry budget tightly bounded: 3s connect, 5s total, no retries on 4xx, one retry on 5xx.
  - Validate `Content-Type: application/json`.
  - Validate required fields per latest CIMD draft.
- **`redirect_uris` enforcement**: the `redirect_uri` parameter on `/authorize` MUST exactly match one of the `redirect_uris` in the fetched document. No fuzzy matching.
- **`grant_types` enforcement**: subsequent `/oauth/token` requests for this `client_id` MUST use a grant_type listed in the document.
- **`scope` enforcement**: requested scopes MUST be a subset of the scopes the document declares (intersection with the Registry's own `scopes_supported`).
- **Cache**: TTL from response `Cache-Control: max-age` (bounded 60s min, 86400s max). LRU cache of documents, ~1000 entries. Key by canonical URL.
[...]

Issue #992, opened 2026-05-09
[Phase 4] Publish CIMD (Client ID Metadata Document) for AI Registry Tools MCP client


aarora79:
# Sub-issue B: Publish a CIMD (Client ID Metadata Document) for the AI Registry Tools MCP client

**Parent:** #988
**Labels:** `enhancement`, `oauth`, `mcp-spec`, `authentication`, `subtask`
**Phase:** 4
**Depends on:** A, E
**Blocks:** C

---

## Summary

When the Registry itself acts as an MCP client (calling out to other CIMD-aware authorization servers on behalf of AI Registry Tools), it needs a stable, publicly-fetchable Client ID Metadata Document describing itself. Publish it at a canonical URL and treat that URL as the Registry's `client_id` when it authenticates to external IdPs that have adopted CIMD.

## Background

The November 2025 MCP authorization spec update ([Aaron Parecki](https://aaronparecki.com/2025/11/25/1/mcp-authorization-spec-update)) retires RFC 7591 Dynamic Client Registration in favor of CIMD. In CIMD, a client publishes `https://example-app.com/client.json` describing itself; the authorization server fetches that document on the first auth request, caches it, and treats the URL as the `client_id`. No per-AS registration, no database growth, no DCR.

Claude Code v2.1.64+ already auto-discovers CIMD. The Registry should publish its own so it can play the client role against external IdPs that have adopted the pattern (most relevant once our AI Registry Tools server starts calling external MCP servers on behalf of users, e.g. through a token-exchange chain).

This is the **publisher** side. The consumer side (accepting an incoming CIMD URL as `client_id` on our `/authorize`) is sub-issue C.

### In scope

- **`GET /.well-known/mcp-client-metadata`** (final path TBD based on spec landing point; track the latest CIMD draft at [draft-parecki-oauth-client-id-metadata-document](https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/)) returning:
  - `client_name`: human-readable, e.g. "AI Registry Tools".
  - `client_uri`: Registry homepage.
  - `logo_uri`: static asset served by the Registry.
  - `redirect_uris`: list of allowed callbacks for the AI Registry Tools client flow.
  - `scope`: space-separated list of scopes the client requests.
  - `grant_types`: at least `authorization_code`, `refresh_token`, and (once D lands) `urn:ietf:params:oauth:grant-type:token-exchange`.
[...]

Issue #988, opened 2026-05-09
Umbrella: Coding-Assistant OAuth Integration (catalog + gateway modes)


aarora79:
# Umbrella: Coding-Assistant OAuth Integration (catalog mode + gateway mode)

**Labels:** `enhancement`, `oauth`, `mcp-spec`, `authentication`, `architecture`, `feature-request`

**Linked existing issues:** #256 (spec umbrella), #614 (tactical PKCE + metadata), #889 (TTL), #936 (to be closed as superseded), #895, #896

**Owner:** @aarora79

**Status:** Proposed

---

## Summary

Two distinct modes for how a coding assistant interacts with MCP servers relative to the Registry. No middle ground, no federation layer.

### Catalog mode (Registry is discovery-only; Registry is NOT in the data path)

AI Registry Tools (`intelligent_tool_finder`, `list_services`, `list_agents`, `list_skills`) returns the **direct** URL of an external MCP server (e.g. `https://public.cloudflare.com/mcp`). The coding assistant adds that external URL to its own `mcp.json`. Any OAuth or auth negotiation happens directly between the coding assistant and that external server's IdP. The Registry is never in the data path, does not proxy, does not mediate auth, takes on no IdP-compatibility burden. Cloudflare's auth is Cloudflare's problem.

Catalog mode is existing AI Registry Tools behavior; the only umbrella-relevant change is confirming that `intelligent_tool_finder` returns external URLs (not rewritten-through-gateway URLs) for external servers.

### Gateway mode (Registry-routed paths; Registry IS in the data path)

The coding assistant points at a Registry-gateway URL (e.g. `https://mcpgateway.example/ai-registry-tools/mcp` or `https://mcpgateway.example/cloudflare-api/mcp`). The gateway is on every request. It enforces identity + entitlement on ingress via `X-Authorization`. Backend auth on egress has two sub-cases:

1. **Backend under the Registry's trust domain** (today's pattern): gateway injects the backend's credential on egress. One header in `mcp.json`: the ingress `X-Authorization`. The user doesn't see or manage any backend token. This is what today's `cloudflare-docs`-style registrations already do.
[...]

PR #1711, opened 2026-08-31
feat(cimd): standalone Client ID Metadata Document publisher

omrishiv:
**Refs:** #992 (publisher slice — see *Not covered* below; does **not** close it)
**Parent epic:** #988 (Phase 4 CIMD)

## What

Publishes the registry's own **Client ID Metadata Document (CIMD)** so the registry can identify itself as an OAuth *client* (by URL) to external CIMD-aware IdPs, without pre-registration or DCR. Per the [CIMD draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-client-id-metadata-document/), the document's URL **is** the `client_id`.

```
GET /oauth/client-metadata.json      # public, unauthenticated, cacheable
```

Opt-in (`CIMD_PUBLISHER_ENABLED`, default off → 404). Returns `client_id`, `client_name`, `client_uri`, `redirect_uris`, `grant_types`, `response_types`, `token_endpoint_auth_method: none`, `scope`, and optional `logo_uri`/`contacts`, with `Cache-Control: public, max-age=<CIMD_CACHE_TTL>`.

## Why this is a standalone branch

The original CIMD work was built on top of #1692 (the gateway token proxy), which was **closed**. The CIMD *publisher* has no functional dependency on the token proxy — it references no token-proxy symbols. This branch is the publisher rebased cleanly onto `main`, so #992's publisher can land independently of the closed token-proxy effort. (The token-proxy-coupled bits from the original branch — the `TestCimdClientIdPassthrough` tests and the `python-multipart` dep — were intentionally dropped; they belong with a token proxy, not the publisher.)

## Design decisions

- **Path `/oauth/client-metadata.json`, not `/.well-known/…`.** A CIMD is a *client*-published resource at an operator-chosen public path — RFC 8615 `.well-known` is for authority/server documents,. `/oauth/client-metadata.json` matches the MCP spec's own example and the AT-Protocol convention. #992 explicitly left the path "TBD, track the draft," so this satisfies that note rather than contradicting it.
- **Root-mounted, rides the public catch-all.** The route is served by a root-mounted `cimd_router`; the gateway nginx catch-all `location /` proxies to the registry with **no `auth_request`**, so the AS fetches it unauthenticated with **zero nginx changes**.
- **`client_id`/`redirect_uris` derive from the egress OAuth callback base** (`settings.egress_oauth_callback_base` = `EGRESS_OAUTH_CALLBACK_BASE_URL` or `REGISTRY_URL`), not `registry_url` alone — so a deployment whose external callback host differs from `registry_url` still publishes a `client_id` and redirect that share the origin the IdP redirects back to. A single canonical `build_cimd_client_id_url` guarantees the published URL and any future outbound `client_id` can't drift.
- **Kept the safe defaults:** opt-in gate, `enforce_https` on the `client_id`, and `public, max-age` caching.

## Config surface (three-surface parity)

All 7 registry-consumed, non-secret params are wired first-class:

| Env | Setting | Helm (`registry.app.*`) | Terraform var | Type / default |
|---|---|---|---|---|
| `CIMD_PUBLISHER_ENABLED` | `cimd_publisher_enabled` | `cimdPublisherEnabled` | `cimd_publisher_enabled` | bool / `false` |
| `CIMD_CACHE_TTL` | `cimd_cache_ttl` | `cimdCacheTtl` | `cimd_cache_ttl` | int / `3600` |
| `CIMD_CLIENT_NAME` | `cimd_client_name` | `cimdClientName` | `cimd_client_name` | str / `"AI Registry Tools"` |
| `CIMD_REDIRECT_URIS` | `cimd_redirect_uris` | `cimdRedirectUris` | `cimd_redirect_uris` | str CSV / `""` |
| `CIMD_SCOPE` | `cimd_scope` | `cimdScope` | `cimd_scope` | str / `""` |
| `CIMD_LOGO_URI` | `cimd_logo_uri` | `cimdLogoUri` | `cimd_logo_uri` | str / `""` |
| `CIMD_CONTACTS` | `cimd_contacts` | `cimdContacts` | `cimd_contacts` | str CSV / `""` |

- **System Config UI:** new `CIMD Publisher` group in `config_routes.py` `CONFIG_GROUPS` (all non-sensitive).
- **Docs:** row block in `docs/unified-parameter-reference.md`.
- **Docker:** `.env.example` + all three compose files (typed `${VAR:-false}`/`${VAR:-3600}` defaults so an empty value never breaks Pydantic).
- **Helm:** `charts/registry` values + deployment (bool/int via `toString | quote`) + `reserved-env-names.txt` (chart-managed, not extraEnv-shadowable).
- **Terraform:** root + module variables, module passthrough, `ecs-services.tf` env (bool/int via `tostring()`), `terraform.tfvars.example`.

## Testing

- `tests/integration/test_cimd_publisher.py` — document shape, required fields, field-order stability, cache header, **`client_id` == served URL parity**, **callback-base precedence over `registry_url`**, disabled→404, and a root-mount/no-prefix invariant that reads `main.py` so a future rename/prefix can't silently break parity. **6 passed.**
- **Helm** `helm template charts/registry` renders `CIMD_PUBLISHER_ENABLED: "true"`, `CIMD_CACHE_TTL: "1200"` (bool/int coercion correct).
- **Terraform** `terraform fmt -check` clean on all `.tf` files.
- `ruff check`/`format`, pre-commit (syntax, fast tests) clean.

## Not covered by this PR (so #992 is not closed by mistake)

1. **Outbound `client_id` wiring** — nothing today sends the registry's CIMD URL as its `client_id` on an outbound `/authorize` (egress uses per-server operator credentials). There is no such flow to wire yet; `build_cimd_client_id_url` is the single source of truth the future outbound path must reuse. #992 frames this as forward-looking ("once AI Registry Tools calls external MCP servers").
2. **Fake-AS end-to-end** (#992 acceptance) — a test of that outbound flow; can't be written until (1) exists.
3. **#993 (CIMD consumer)** — accepting a CIMD URL as `client_id` on `/authorize`/`/oauth/token` is **not** in this branch. The original approach ("Option Y": the token proxy forwards `client_id` verbatim to the upstream IdP) depended on #1692's token proxy, which was dropped here. #993 returns with the token-proxy / OAuth-discovery track.

## Files

20 files, +518 / −7. Publisher core: `registry/auth/oauth_metadata.py`, `registry/api/wellknown_routes.py`, `registry/main.py`, `registry/core/config.py`. Surface wiring + docs + tests as above.


aarora79:
Reviewed and pushed three commits to this branch (we have write access). Review: verdict **APPROVE**, no blockers outstanding.

**Rebased onto `main`.** The branch was 29 commits behind and carried a merge commit (`853b1b32`), which this repo cannot merge since it is rebase-and-merge only. Now 6 commits, no merges. Two conflicts were additive-at-the-same-position (main added `EGRESS_OAUTH_TRUSTED_IDP_HOSTS` where the CIMD block went, in both `.env.example` and the parameter reference); I kept both sides.

**`test(helm): re-derive registry env indices after the CIMD block.`** The third conflict was `extra_env_test.yaml`, and both sides were stale: main's numbering predates the CIMD block, and this branch's predates two `GATEWAY_*` vars main has since added. Rather than do the arithmetic I re-derived every index from `helm template` output, correcting 27, and moved the `extraEnv` append positions from env[49]/[50] to env[56]/[57] since the chart-managed set now runs 0-55. I also restored the 7 CIMD position assertions this branch had (taking main's side of the conflict dropped them) and added a reserved-name rejection test for `CIMD_PUBLISHER_ENABLED`, since all 7 names are in `reserved-env-names.txt` but none had an extraEnv test. 89 helm tests pass.

**`fix(cimd): quote CIMD_CLIENT_NAME in .env.example`.** This one only shows up when you build. `build_and_run.sh` does `source .env` (four call sites) and the documented flow is `cp .env.example .env`, so the unquoted `CIMD_CLIENT_NAME=AI Registry Tools` made bash read "assign CIMD_CLIENT_NAME=AI, then run the command `Registry`":

```
.env: line 1408: Registry: command not found
```

Had it survived, the variable would have held `AI`. Every other space-containing value in that file was already quoted (`REGISTRY_NAME`, `REGISTRY_ORGANIZATION_NAME`, `REGISTRY_DESCRIPTION`), so it was a single line missing the convention. The other three surfaces were unaffected: compose interpolates the default itself (verified with `docker compose config`), Helm already quotes it, and the Terraform default is an HCL string.

**`test(env): make .env.example shell-safety enforceable in CI`.** Three checks so this class cannot recur: every whitespace-containing value is quoted (names the line and key), `bash -c 'source .env.example'` exits 0 with empty stderr (catches backticks, `$(...)`, unbalanced quotes too), and values survive the round trip. That last one matters because the quiet half of this bug is a value that parses while losing its tail. Verified load-bearing: reverting the quotes fails all three, and the second reproduces the operator-visible error verbatim.

**Verified on the rebased branch:** 8561 python tests pass (4 pre-existing token-TTL failures from this host's `.env` variant, reproduced identically on `main`), 6 CIMD integration tests, 89 helm unittests, helm lint clean, ruff and bandit clean. Also built and ran it: the endpoint serves the 9-field document unauthenticated with `Cache-Control: public, max-age=3600` and `client_id` correctly derived from the callback base.

**Config parameter parity is complete**, checked per-parameter rather than from the description: all 7 across `.env.example`, three compose files, root + module Terraform + `ecs-services.tf` + tfvars example, Helm values + deployment + `reserved-env-names.txt`, `CONFIG_GROUPS`, and the unified reference. (Note for future reviewers: `CONFIG_GROUPS` keys on the snake_case setting name, so grepping for the env name there gives a false negative.)

Two non-blocking docs suggestions: `CIMD_CONTACTS` is published on a public unauthenticated endpoint, so a line recommending an alias over a person would prevent a foot-gun; and worth suggesting a low `CIMD_CACHE_TTL` during rollout, since 3600 means a config change takes up to an hour to reach a provider that cached the document.

Agreed with the decision to ship the publisher without the consumer. It settles the path, base-URL derivation, document shape and config surface, which are the expensive-to-change decisions given the URL is the identity.

aarora79:
Pushed `7323b7d3` with three of the four nice-to-haves from the review.

**Collapsed the duplicate client_id computation, and it turned out to be slightly more than tidiness.** The handler computed the URL once for `enforce_https` and again inside `build_cimd_document`. It now builds the document first and gates on `document["client_id"]`, so the URL we require HTTPS for is provably the URL we publish rather than a second computation that happens to agree. Building the dict has no side effects, so nothing is emitted before the check runs. Also drops the now-unused `build_cimd_client_id_url` import from the routes module.

**Pinned the one construction site.** `test_client_id_has_exactly_one_construction_site` guards the seam #992's outbound wiring will land on. Since the URL *is* the client_id, two sites that drift publish one value while `/authorize` sends another, and the IdP rejects it with nothing in the document to explain why. A test that only calls the helper cannot see that, so this scans `registry/**.py` for the path literal joined onto anything, outside `oauth_metadata.py` (the definition) and `wellknown_routes.py` (the route decorator, already pinned to `CIMD_PATH` by the test above it). Verified load-bearing by injecting `f"{base}/oauth/client-metadata.json"` into another module: it fails and names the file and line.

**Documented that `CIMD_CONTACTS` is public.** The endpoint is unauthenticated by design, since an IdP has to fetch it before any trust exists, so every address there is world-readable the moment the publisher is enabled. The note recommends a role alias over an individual's address. Default is empty so nothing is published unless set. Also added the missing `CIMD_CACHE_TTL` row to that settings table.

I left the fourth suggestion (advise a low `CIMD_CACHE_TTL` during rollout) out, since it is advice rather than a defect and the maintainer scoped it out.

**Verified:** 7 CIMD integration tests, 8566 python tests (the 4 token-TTL failures are this host's `.env` variant, reproduced on `main`), ruff clean. Re-ran the stack afterwards and the endpoint still serves correctly with `cache-control: public, max-age=3600`.

One small thing noticed while testing, not changed: the route is GET-only, so `curl -sI` (HEAD) returns 405. Most IdPs will GET it, so this is an observation rather than a finding, but worth knowing if anyone reaches for `curl -I` to check the cache header.

PR #1713, opened 2026-08-31
fix(auth): normalize IdP group names before matching scope mappings

### LINKED ISSUE #1689: IdP group names are never normalized, and filter_session_groups drops groups without naming them
## Summary

Two related gaps around IdP group handling. Neither is an authorization defect — access decisions are correct — but together they make a misconfigured group claim close to invisible, which cost us several days on an unrelated investigation (#1648, withdrawn as not-a-bug).

Both are the same shape as #1647: the system behaves as configured, produces no error, and gives the operator nothing to grep for.

## 1. Group names are never normalised, while server names are

Server names go through `_normalize_server_name` (`auth_server/server.py`), which strips leading and trailing slashes so a scope entry written as `/cloudflare-docs` matches a request for `cloudflare-docs`. The docstring is explicit that this exists because the UI and the URL extraction disagree on the leading slash.

Group names get no such treatment. They are compared verbatim in every place they are used:

- `get_group_mappings_bulk` — `find({"group_mappings": {"$in": unique}})`
- `get_all_mapped_group_names` — set union, exact membership
- `filter_session_groups` / `_filter_by_scope_mappings` — `[g for g in groups if g in mapped]`

This matters because **Keycloak's Group Membership protocol mapper emits full group paths by default** when *Full group path* is enabled: the claim carries `/mcp-admins`, not `mcp-admins`. A scope document seeded with `mcp-admins` then matches nothing.

The consequences are silent and compounding:

1. `get_group_mappings_bulk` returns no scope for that group, so the user resolves to fewer scopes than they should.
2. `filter_session_groups` drops the group from the session entirely, because it is not in the mapped-group union.

The user simply has less access than intended, with no error anywhere. The same applies to any IdP that emits a prefixed or differently-cased group identifier.

**Suggestion:** normalise group names at the comparison points, mirroring what `_normalize_server_name` already does for servers — at minimum strip surrounding slashes and whitespace. Doing it in one shared helper would keep the seeding path and the lookup path from drifting. If verbatim matching is deliberate, a note in `group_filter.py` and in the scope-seeding docs would at least make the trap findable.

## 2. `filter_session_groups` logs counts, never names

`auth_server/group_filter.py` intersects the user's IdP groups with the union of every scope's `group_mappings` at login, and persists only the filtered list in the session. The module docstring argues this is lossless for authorization, which is true — *provided* `group_mappings` is complete and in the matching form.

When it is not, the only trace is:

```
Group filter (scope-derived) user=<hash>: 5 -> 1
```

Counts, never names. A user can lose their most important group at login and nothing identifies it. There is no later log line either, because from that point on the session genuinely only has one group, and every downstream line faithfully reports one group.

This is what made #1648 hard: we could see that a user resolved to one scope, but nothing anywhere said which groups had been discarded on the way in, or why.

**Suggestion:** log the dropped group names when the filter removes anything. If group names are considered sensitive — the code already treats them as organisational PII in `validate_server_tool_access`, and Entra group names can encode org units — then DEBUG level, or the count at INFO plus the names at DEBUG, would still be a large improvement over counts alone.

Related: the filter runs only on the session-cookie path (`auth_server/server.py`, at the session write). A bearer-token request reads groups live from the token claim, unfiltered. That asymmetry may well be intentional, but it is not documented, and it means the same user can have different effective permissions depending on how a given request authenticated. A comment saying so would help anyone auditing "what could this user access at 09:43".

## Environment

Found on 1.29.0, Docker Compose, `AUTH_PROVIDER=keycloak` with Keycloak brokering Entra ID over SAML. Both observations are from reading `main` rather than from a reproduction, since our own incident turned out to have a different cause.

AmirF194:
## Root cause

`_normalize_server_name` (`auth_server/server.py`) strips leading/trailing
slashes so a scope entry written as `/cloudflare-docs` still matches a
request for `cloudflare-docs`. Group names never got the same treatment, so
every comparison point matched verbatim:

- `group_filter.py::_filter_by_scope_mappings` (Design C): `[g for g in groups if g in mapped]`
- `group_filter.py::filter_session_groups` (Design B): `[g for g in groups if g in allowed]`
- `server.py::map_groups_to_scopes`, which calls `get_group_mappings_bulk(groups)` (a MongoDB `$in` against unslashed stored `group_mappings`) and falls back to the same exact-match lookup against `SCOPES_CONFIG`

Keycloak's Group Membership mapper emits the full group path (`/mcp-admins`)
when "Full group path" is enabled, while scope mappings are seeded without
the leading slash. A user in that group then silently resolves to zero
scopes for it, both in the session (dropped by the login-time filter) and on
every subsequent authorization check.

## Fix

Added a small `_normalize_group_name` helper (same logic as
`_normalize_server_name`: strip surrounding slashes) at each comparison
point:

- `filter_session_groups` normalizes the incoming groups once, before either
  the Design B allowlist or the Design C scope-derived branch; the allowlist
  and the scope-mapped set are normalized too, so both sides agree.
- `map_groups_to_scopes` normalizes its `groups` argument before calling
  `get_group_mappings_bulk` and before the in-memory fallback lookup, which
  covers all five call sites of this function.

## Scope

This addresses the normalization gap only (the first half of #1689). The
issue also asks for the group filter to log dropped group *names*, not just
counts, plus a documentation note on the bearer-token/session asymmetry;
neither is touched here, so this is `Refs #1689` rather than `Fixes #1689`.

## Verification

- New regression tests in both affected modules (leading-slash IdP claim
  matching an unslashed mapping, allowlist entry, and the bulk scope query),
  confirmed failing against unmodified `auth_server/group_filter.py` and
  `auth_server/server.py` and passing against this branch, in a clean
  `python:3.14-slim` container.
- Full `tests/auth_server/` suite: 820 passed, no regressions, same
  container.
- `ruff check`, `ruff format --check`, and the repo's scoped `mypy` invocation
  (`--ignore-missing-imports --no-strict-optional` on `registry|auth_server`)
  all clean; `bandit` reports no new findings.
- Did not run the registry-wide test suite (needs a MongoDB service
  container); this diff touches only `auth_server/`, and its own tests
  exercise every changed line.

### DIFF
--- a/auth_server/group_filter.py
+++ b/auth_server/group_filter.py
@@ -57,6 +57,18 @@ if ALLOWED_IDP_GROUPS_RAW:
     logger.info("ALLOWED_IDP_GROUPS allowlist active: %d entries", len(ALLOWED_IDP_GROUPS))
 
 
+def _normalize_group_name(name: str) -> str:
+    """Strip leading/trailing slashes for group-name comparison.
+
+    Some IdPs (e.g. Keycloak's Group Membership mapper with "Full group path"
+    enabled) emit the full group path, such as '/mcp-admins', while scope
+    mappings are seeded without the leading slash. Mirrors
+    auth_server.server._normalize_server_name, which handles the identical
+    slash mismatch for server names.
+    """
+    return name.strip("/") if name else name
+
+
 async def _filter_by_scope_mappings(
     groups: list[str],
     username_hash: str,
@@ -88,6 +100,7 @@ async def _filter_by_scope_mappings(
         )
         return groups
 
+    mapped = {_normalize_group_name(g) for g in mapped}
     filtered = [g for g in groups if g in mapped]
     logger.info(
         "Group filter (scope-derived) user=%s: %d -> %d",
@@ -115,9 +128,11 @@ async def filter_session_groups(
     if not groups:
         return groups
 
+    groups = [_normalize_group_name(g) for g in groups]
+
     # Design B takes precedence when configured.
     if ALLOWED_IDP_GROUPS:
-        allowed = set(ALLOWED_IDP_GROUPS)
+        allowed = {_normalize_group_name(g) for g in ALLOWED_IDP_GROUPS}
         filtered = [g for g in groups if g in allowed]
         logger.info(
             "Group filter (allowlist) user=%s: %d -> %d",
--- a/auth_server/server.py
+++ b/auth_server/server.py
@@ -1470,6 +1470,18 @@ def safe_identity_summary(claims: dict) -> dict:
     return summary
 
 
+def _normalize_group_name(name: str) -> str:
+    """Strip leading/trailing slashes for group-name comparison.
+
+    Some IdPs (e.g. Keycloak's Group Membership mapper with "Full group path"
+    enabled) emit the full group path, such as '/mcp-admins', while scope
+    mappings are seeded without the leading slash. Mirrors
+    _normalize_server_name, which handles the identical slash mismatch for
+    server names.
+    """
+    return name.strip("/") if name else name
+
+
 async def map_groups_to_scopes(groups: list[str]) -> list[str]:
     """
     Map identity provider groups to MCP scopes by querying DocumentDB directly.
@@ -1481,6 +1493,7 @@ async def map_groups_to_scopes(groups: list[str]) -> list[str]:
         List of MCP scopes
     """
     scopes = []
+    groups = [_normalize_group_name(g) for g in groups]
 
     # Resolve all groups to scopes in a single query. Issuing one query per
     # group serialized a DB round-trip per group on every authenticated
--- a/tests/auth_server/unit/test_group_filter.py
+++ b/tests/auth_server/unit/test_group_filter.py
@@ -91,6 +91,42 @@ class TestFilterSessionGroups:
         assert result == ["a", "b"]
 
     @pytest.mark.asyncio
+    async def test_scope_derived_normalizes_leading_slash(self):
+        """Design C: a full-group-path IdP claim matches an unslashed mapping.
+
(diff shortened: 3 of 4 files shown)


AmirF194:
This reuses the existing _normalize_server_name pattern from the same file for group claims: Keycloak's full-group-path claim carries a leading slash that group_filter.py's exact string match never strips, so a correctly seeded scope mapping silently never matches. Tests cover both the slash-prefixed and bare-name cases.

auth_server/group_filter.py
```diff
@@ -57,6 +57,18 @@
     logger.info("ALLOWED_IDP_GROUPS allowlist active: %d entries", len(ALLOWED_IDP_GROUPS))
 
 
+def _normalize_group_name(name: str) -> str:
```
omrishiv, line 60: can we extract this to a helper and reuse it?

auth_server/group_filter.py
```diff
@@ -88,6 +100,7 @@ async def _filter_by_scope_mappings(
         )
         return groups
 
+    mapped = {_normalize_group_name(g) for g in mapped}
```
omrishiv, line 103: this normalizes the stored values, but the read is currently not normalized in registry/repositories/documentdb/scope_repository.py. We should normalize at write, which will fix the read as well

auth_server/server.py
omrishiv, line 4300: we might have duplicates here now due to the normalization. worth deduping this potentially

omrishiv (changes requested):
Hi @AmirF194, thank you for the PR, I left a few code comments, but also some overall comments below:

- We need the same normalization on the registry side or we'll run into inconstencies between the UI/API. This is possible to address in both places by normalizing at the provider end (auth_server/providers/keycloak)
- This might break `ALLOWED_IDP_GROUPS=/mcp-admins` for instance. 
- Can we also strip out whitespaces on the normalization?

PR #1719, opened 2026-09-02
feat(security): per-tool blocking for unsafe tools (#1623)

### LINKED ISSUE #1623: Feature request: Per-tool blocking and enable/disable toggles
# Feature request: Per-tool blocking and enable/disable toggles

## Summary

The security scanner reports per-tool findings (SAFE / HIGH / etc.), but blocking and admin toggles only work at the **server** level. When one tool is flagged on an otherwise fine server, we must disable the whole server or expose every tool.

Are there plans to add **registry-wide per-tool blocking** and **per-tool enable/disable toggles**?

## Example

On one MCP server we see mixed results:

- `show_global_status` — SAFE
- `show_global_variables` — SAFE
- `show_processlist` — **HIGH** (tool poisoning)

We want to block `show_processlist` while keeping the server and the other tools available.

Image

## Current behavior

- Scans are per-tool (`tool_results` in scan output).
- Unsafe servers can be auto-disabled via `SECURITY_SCAN_BLOCK_UNSAFE_SERVERS` / `security-pending`.
- Admins can enable/disable a **server**, not individual tools.

## Requested behavior

1. **Per-tool block** — block specific tools on an enabled server; enforce on `tools/call` and filter from `tools/list` / search. Optional auto-block on HIGH/CRITICAL rescan findings.
2. **Per-tool enable/disable** — admin toggle per tool (like the server toggle), for ops use as well as security.

## Questions

- Is this on the roadmap?
- For mixed-severity servers, should flagged tools be blocked instead of disabling the whole server?

vrindabhandari:
Closes #1623 (backend enforcement; UI to follow)

## What this does

Implements opt-in per-tool blocking per the behavior model in the issue. When `allow_unsafe_servers` is enabled, a server with HIGH/CRITICAL findings registers as **enabled** with its unsafe tools individually blocked, rather than the whole server being disabled. Default-off, so existing deployments are unaffected.

- Config flag + `ToolOverride` schema
- Repository methods for per-tool overrides
- `set_tool_blocked` service method with Mongo key validation
- Scan reconciliation: auto-blocks reapply on rescan, admin decisions survive
- `POST /toggle-tool/{service_path}` admin route
- Enforcement on `tools/call` (403 for blocked tools)

## Scope

Backend enforcement only. Agreed with @aarora79 to land this first and do the UI as a separate PR. Deliberately left for that follow-up:

- CLI methods for per-tool blocks
- `include_blocked` flagging on `GET /api/tools`
- Auto-block wiring at registration time
- Frontend per-tool toggle

## Known limitation: rename evasion

Blocks are keyed on tool name. A server that renames a blocked tool presents a name that isn't in the blocklist, so it passes both the `tools/list` filter and `tools/call` enforcement until the next scan reconciles. Auto-blocks self-heal on
rescan; admin-set blocks are the exposed case.

Raised by @Santoshkumarpuppala on the issue. His content-hash suggestion includes the tool name in the hash, which a rename also changes — so it matches on neither name nor hash. The stronger version is his second point: treat an unrecognized tool name on a server that already has scan state as unscanned and fail closed. The tool then doesn't need to match a block, it needs to have been scanned, which closes the rename path regardless of the tool body. Consistent with the fail-closed posture already in `validate_server_tool_access`.

## Testing

`TestToggleTool` and `TestReconcileSecurityBlocks` cover the route and the reconciliation matrix (auto-block applied, stale auto-block cleared, admin block survives clean rescan, admin unblock not re-blocked). Merged current main; full suite green (7389 passed, 31 skipped).


aarora79:
Rebased onto main (was 98 commits behind, branched Sep 1) and force-pushed. Nothing in the implementation changed.

Three things the rebase needed:

1. **Dropped the merge commit.** `0530678f` ("Merge remote-tracking branch 'upstream/main'") is gone. This repo allows rebase-and-merge only, so a branch carrying a merge commit cannot be merged at all, regardless of conflicts. The 7 feature commits replay cleanly.
2. **`tests/unit/api/test_server_routes.py`**: both sides appended a test class to the end of the file, `TestMetadataFieldsProjection` from main and `TestToggleTool` here. Kept both.
3. **`auth_server/server.py`**: main relocated `from registry.repositories.factory import get_scope_repository` further down the import block, and this branch had added `get_server_repository` at the old position. Folded the new name into the import where main now keeps it.

Also dropped the `uv.lock` churn (+14/-17). It removed the global `exclude-newer` pin and added `sys_platform == 'darwin'` markers to the torch dependency tree, which is what `uv run` does to the lockfile on a Mac rather than an intended dependency change. `uv.lock` is now identical to main.

Verified after the rebase: ruff clean, 551 tests pass across the four touched test files. Three failures in `TestTokenLifetimeEnforcement` are pre-existing and environment-caused, not from this branch: they reproduce identically on main when `.env` sets `MCP_TOKEN_DEFAULT_TTL_HOURS=2` while the tests assert 8.

Issue #1696, opened 2026-08-27
Duplicate-check advisory produces false positives on names sharing a common catalog suffix (e.g. "-mcp-server")


rolland80:
Title: Duplicate-check advisory produces false positives on names sharing a common catalog suffix (e.g. "-mcp-server")

Body:

## Description

`DuplicateCheckService` (`registry/services/duplicate_check_service.py`) flagged a newly-registered server, `topic-mcp-server`, as a similarity match against an existing, functionally unrelated server, `category-mcp-server`. The two servers have no overlapping capability.

## Root cause

`_build_query_text` (lines ~364-384) concatenates raw `name + description` with no boilerplate/stopword stripping before embedding. Both example names share the structural token `mcp-server`. Short, token-sparse strings are a known weakness for sentence-embedding models: shared boilerplate tokens dominate the embedding proportionally more than they would in a longer, distinctive description, inflating cosine similarity independent of actual semantic/capability overlap.

`_fetch_similarity_advisory` (lines ~295-362) then applies a single uniform `dedup_score_threshold` (default `0.7`) with no structural pre-filter (shared category/tags, minimum text richness) before the semantic step runs.

Raising the threshold alone does not fix this without also suppressing genuine near-duplicates that happen to have equally short/generic descriptions.

## Suggested fix

Strip common catalog-boilerplate tokens (`mcp`, `server`, `tool`, `skill`, `agent`, `api`, `service`, etc.) from the text passed into `_build_query_text` before embedding, so a shared structural suffix alone can't drive a false positive. Optionally, consider a minimum-text-richness check that discounts or skips the similarity comparison when either side's post-stripped text is too thin to embed meaningfully.

## Reproduction

1. Register a server named `topic-mcp-server` with any description.
2. Register a second, unrelated server named `category-mcp-server`.
3. Observe `/check-duplicates` (or the registration UI's dedup hint) surfaces the first as a similarity match against the second, despite no functional overlap.

## Environment

Observed via an internal and customized AI Registry deployment of this project, but the affected code path is fully upstream (`registry/services/duplicate_check_service.py`), not company-specific.

aarora79:
Fixed by #1721 and #1744, verified against a live registry on both MongoDB CE and DocumentDB.

[...]

PR #1721, opened 2026-09-04
fix(dedup): filter duplicate advisories on cosine similarity, not display rank

ebarkhordar:
`DuplicateCheckService` compares `dedup_score_threshold` against `relevance_score`, but under the default `rrf` fusion that value is a position in a result list, not a similarity. `_normalize_scores` min-max rescales the fused scores, and all three of its return paths hand exactly 1.0 to the highest scorer (`search_repository.py:290`). Since 1.0 always clears 0.7, the filter at `duplicate_check_service.py:342` cannot reject the top candidate whatever the registry holds.

The normalization is doing its job. `d97a24b6` added it so the UI would stop rendering RRF scores as "1-3% match"; the advisory is just reading a display number as if it were a similarity.

That makes the shared `-mcp-server` suffix incidental to this report. Against a registry holding only `payroll-mcp-server` and `weather-mcp-server`, registering `topic-mcp-server` still returns an advisory at 1.0, on true cosines of 0.2648 and 0.2313.

## Fix

`search()` now carries `similarity_score` next to `relevance_score`: the cosine of the hit against the query embedding, stamped by `_attach_similarity_scores` on the hybrid and client-side paths. `relevance_score` and the UI are untouched. The advisory ranks and filters on `similarity_score`, and a hit arriving without one is treated as not comparable rather than admitted.

This also takes the non-default linear branch's `(cosine + 1) / 2` mapping out of the threshold's meaning, where a configured 0.7 was really asking for cosine 0.4.

The existing advisory fixtures set only `relevance_score`, which is no longer the field that gates, so I renamed it there to keep them testing what they were testing.

## Verification

- Four new behavioural tests in `test_duplicate_check_service.py` fail on `main` and pass here, plus `test_search_similarity_scores.py` for the helper.
- Full `tests/unit` suite on Python 3.14: 6594 passed, 8 skipped.
- `ruff check`, `ruff format` and `mypy` clean at the pinned pre-commit versions.
- Two gaps. The matching call on the MongoDB CE client-side fallback is not covered, since that method has no test harness in the repo. And a registration with an empty description still embeds the bare name, where `topic-mcp-server` and `category-mcp-server` sit at 0.7621, so the boilerplate effect described in the report survives that case.

Stripping catalog tokens before embedding would close that second one, but it changes what gets indexed rather than how a score is read, so it seemed worth keeping separate.

Refs #1696

### DIFF
--- a/registry/core/config.py
+++ b/registry/core/config.py
@@ -1503,7 +1503,11 @@ class Settings(BaseSettings):
         default=0.7,
         ge=0.0,
         le=1.0,
-        description="Minimum semantic-search score (0..1) for an advisory match to be returned.",
+        description=(
+            "Minimum cosine similarity (0..1) between the incoming entity's "
+            "name plus description and an existing one's for an advisory "
+            "match to be returned."
+        ),
     )
     dedup_max_suggestions: int = Field(
         default=3,
--- a/registry/repositories/documentdb/search_repository.py
+++ b/registry/repositories/documentdb/search_repository.py
@@ -287,6 +287,40 @@ def _reciprocal_rank_fusion(
 SCORE_DISPLAY_FLOOR: float = 0.10
 
 
+def _attach_similarity_scores(
+    grouped_results: dict[str, list[dict[str, Any]]],
+    selected_results: list[tuple[dict, float]],
+    query_embedding: list[float] | None,
+) -> None:
+    """Stamp each hit with its raw cosine similarity to the query, in place.
+
+    ``relevance_score`` is a ranking signal: under RRF it is rank-relative,
+    and ``_normalize_scores`` maps the best hit to exactly 1.0 whatever its
+    real similarity. Callers that need an absolute "how alike are these"
+    number, rather than "what came first", read ``similarity_score``.
+
+    Hits are matched to their source document by ``path``. Entries without
+    one (tools extracted from a parent server) are left untouched, as are
+    all entries when the query could not be embedded.
+    """
+    if not query_embedding:
+        return
+
+    similarity_by_path: dict[str, float] = {}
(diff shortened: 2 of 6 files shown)


aarora79:
Verified on `94e57e7a`: full unit suite 6594 passed / 8 skipped, `ruff` and `bandit` clean, and the four new advisory tests fail against `main`, so the behaviour change is pinned. The diagnosis is right and the fix is the right shape — gating on a rescaled RRF rank was the defect firing on every registration. Merging.

Residual work we're picking up in a follow-up, tracked against #1696:

- **Query-side boilerplate stripping** plus a minimum-token guard. This is what closes the reported case: with a thin description, `topic-mcp-server` still measures 0.7621 against `category-mcp-server` and clears the gate, as you noted.
- **`DEDUP_SCORE_THRESHOLD` default 0.7 → 0.6**, in the same commit as stripping — the value is calibrated on stripped text.
- **`similarity_search_available`** should report `false` when hits come back with no similarity, so a lexical fallback is not reported as "nothing similar found".
- **The cosine is computed for every search caller**, not just the advisory. Measured 0.5-2.2 ms per `/api/search/semantic` request at `max_results=10`, and the MongoDB CE path already computes the same value at `search_repository.py:1812`. We'll reuse it rather than recompute.
- Parameter-reference row, API-reference line, one INFO log line per check, and a release note.

One correction on deferring the stripping: it does not change what gets indexed. It rewrites one string on the way to the embedder inside `/check-duplicates` and touches no document, index, or ranking — so the follow-up needs no re-index.

#1696 stays open until that lands. Thanks for the clean diagnosis and for naming your own gaps in the description — it made this review fast.

PR #1722, opened 2026-09-05
fix(servers): make PATCH a field-scoped, revision-guarded write

### LINKED ISSUE #1716: PATCH /api/servers/{path} fails and uses non-atomic full-card write on MongoDB CE
## Observed

On registry image version 1.29.0 / source revision `66d1945b6e91531512f44649e36a6bfc5c798af2`, an authenticated registry-admin M2M request to `PATCH /api/servers/{path}` returns HTTP 500 with `Failed to save server` for a description-only JSON merge patch. The card exists and is readable.

There is a second, independent correctness problem at that revision. `patch_server_endpoint` reads the complete card and merges the narrow client payload, then `server_service.update_server` reaches a Mongo repository update that performs an unconditional full-card `$set`. A concurrent egress OAuth, credential, or other card update can be overwritten even though the client omitted that field.

This blocks Git reconciliation of existing-card metadata and tool inventory. A full-card PUT has the same lost-update problem. A client-side `If-Match` check before the repository call is not race-proof unless the compare participates atomically in the repository write.

## Expected

A description-only PATCH returns 2xx and persists only that field. The same endpoint accepts `tool_list` / `num_tools` while preserving absent and concurrently updated auth, OAuth, egress, credential, and server-managed fields.

The persistence primitive must be either:

- an atomic field-scoped `$set` containing only fields present in the validated patch; or
- a repository-level compare-and-set whose version/revision predicate and write are one atomic database operation.

## Relevant source path

`registry/api/server_routes.py::patch_server_endpoint` reads and merges the card before `server_service.update_server(path, merged)`. The MongoDB CE repository then performs the write; the current route collapses a false return to the 500 and does not expose the underlying exception.

## Acceptance

- description-only PATCH succeeds as an admin identity
- tool-list growth succeeds
- the repository write is field-scoped or atomic-CAS, not read/merge plus unconditional full-card `$set`
- a concurrent update to an omitted credential/OAuth/egress field cannot be lost (persistence-level concurrency regression test)
- any `If-Match`/revision contract is enforced in the same repository operation as the write
- a repeated identical PATCH is harmless
- focused MongoDB CE route and repository regression coverage

atirna:
## Summary

I changed `PATCH /api/servers/{path}` from a read-merge-full-card-write into a field-scoped, revision-guarded write, and moved the If-Match check into the repository `update_one` itself.

What was happening (issue #1716), on the MongoDB CE backend:

- `update_server_endpoint`/`patch_server_endpoint` read the full card, merged the narrow patch, then `DocumentDBServerRepository.update()` persisted the whole merged card with an unconditional full-card `$set`
- a card stored under the slash variant of its path (readable that way because `get()` falls back to the alternate `_id`) never matched the update filter, so a description-only PATCH came back `500 Failed to save server`
- any field written between the PATCH's read and its write, a rotated backend credential, an egress OAuth config, health status, was overwritten even though the patch never mentioned that field

What changed:

- `DocumentDBServerRepository.update()` scopes its `$set` to `updated_fields` (plus `updated_at`) when given, tries the slash-variant `_id` on a miss the same way `get()` does, and takes an optional `expected_updated_at` that joins the `_id` in the same `update_one` filter, so the If-Match compare-and-set is one atomic database operation, not a client-side pre-check
- `PATCH` passes the patch's field scope plus the revision it read; a guarded write that matches nothing is a `412`, not a `500`
- `PUT` keeps its full-card replacement semantics, but its If-Match now also rides the repository write

## Testing

- before, on current `main`, against a real mongo:8.2: a description-only PATCH on a card stored under `/legacy/` read fine but `update_one` matched 0 documents and the route returned `500 Failed to save server`
- before: a credential rotation landing between the PATCH's read and write was clobbered by the stale full-card `$set` (stored `auth_credential_encrypted` reverted to the pre-rotation value)
- after: `pytest tests/integration/test_server_patch_concurrency.py tests/unit/repositories/test_documentdb_server_repository.py tests/unit/api/test_server_routes_patch.py tests/unit/api/test_server_routes_put.py`: the slash-variant card patches with 2xx, the concurrent credential survives, a stale If-Match fails atomically with nothing written, an identical replay is harmless, and clearing `proxy_pass_url` still `$unset`s the identity-url sidecar

The new integration test follows the existing `test_agent_batch_repository.py` pattern (real collection, skip when MongoDB is unreachable), so it runs in CI alongside the mongo:8.2 service the workflow already starts.

Fixes #1716

### DIFF
--- a/registry/repositories/documentdb/server_repository.py
+++ b/registry/repositories/documentdb/server_repository.py
@@ -417,8 +417,33 @@ class DocumentDBServerRepository(ServerRepositoryBase):
         self,
         path: str,
         server_info: dict[str, Any],
+        *,
+        updated_fields: list[str] | None = None,
+        expected_updated_at: str | None = None,
     ) -> bool:
-        """Update an existing server."""
+        """Update an existing server.
+
+        Args:
+            path: Server path as used by callers (with or without a
+                trailing slash).
+            server_info: Full merged server dict (callers build this by
+                layering their change over a fetched card). Only the
+                fields named in ``updated_fields`` (plus ``updated_at``)
+                are written, so concurrent writers touching other fields
+                are not overwritten (issue #1716). When omitted, every
+                field is written (legacy full-card behaviour).
+            expected_updated_at: Optional ``updated_at`` value the caller
+                read. When given, it participates in the same
+                ``update_one`` filter as the ``_id`` match, so the
+                compare-and-set is atomic: if another writer persisted a
+                change since this caller read the card, the filter
+                matches nothing and the update fails (issue #1716).
+
+        Like :meth:`get`, a card stored under the slash variant of
+        ``path`` is written through the variant that exists, so a read
+        that found the card is never followed by a write that silently
+        misses it (issue #1716).
+        """
         logger.debug(
             f"DocumentDB WRITE: Updating server at '{path}' in collection '{self._collection_name}'"
         )
@@ -429,6 +454,9 @@ class DocumentDBServerRepository(ServerRepositoryBase):
         try:
             doc = {**server_info}
             doc.pop("path", None)
+            if updated_fields is not None:
+                field_set = set(updated_fields) | {"updated_at"}
+                doc = {k: v for k, v in doc.items() if k in field_set}
             populate_normalized_identity_url(doc, ENTITY_TYPE_SERVER)
             unset_ops: dict[str, str] = {}
             # update() may carry no proxy_pass_url at all (a partial
@@ -446,10 +474,31 @@ class DocumentDBServerRepository(ServerRepositoryBase):
             update_spec: dict[str, dict[str, Any]] = {"$set": doc}
             if unset_ops:
                 update_spec["$unset"] = unset_ops
-            result = await collection.update_one({"_id": path}, update_spec)
+
+            # Try the exact _id first; on a miss, a card stored under the
+            # slash variant (readable that way via get(), issue #1716)
+            # still gets the write, guarded by the same revision
+            # predicate. Each update_one is atomic on its own; a miss on
+            # both forms means either the card is gone or the stored
+            # updated_at moved, and both must fail the write.
+            def _filter_for(id_value: str) -> dict[str, Any]:
+                if expected_updated_at is None:
+                    return {"_id": id_value}
+                return {"_id": id_value, "updated_at": expected_updated_at}
+
+            result = await collection.update_one(_filter_for(path), update_spec)
+            if result.matched_count == 0:
+                alternate_path = path.rstrip("/") if path.endswith("/") else path + "/"
+                if alternate_path != path:
+                    result = await collection.update_one(_filter_for(alternate_path), update_spec)
 
             if result.matched_count == 0:
-                logger.error(f"Server at '{path}' not found in DocumentDB")
+                if expected_updated_at is not None:
+                    logger.info(
+                        f"Server at '{path}' revision mismatch or missing; concurrent update won"
+                    )
+                else:
+                    logger.error(f"Server at '{path}' not found in DocumentDB")
                 return False
 
             logger.info(
--- a/registry/repositories/interfaces.py
+++ b/registry/repositories/interfaces.py
@@ -128,8 +128,12 @@ class ServerRepositoryBase(ABC):
         self,
         path: str,
         server_info: dict[str, Any],
+        *,
+        updated_fields: list[str] | None = None,
+        expected_updated_at: str | None = None,
     ) -> bool:
-        """Update an existing server."""
+        """Update an existing server (see the DocumentDB impl for the
+        issue #1716 field-scoping and revision-guard semantics)."""
         pass
 
     @abstractmethod
--- a/registry/services/server_service.py
+++ b/registry/services/server_service.py
@@ -231,8 +231,23 @@ class ServerService:
             "is_new_version": False,
         }
 
-    async def update_server(self, path: str, server_info: dict[str, Any]) -> bool:
+    async def update_server(
+        self,
+        path: str,
+        server_info: dict[str, Any],
+        *,
+        updated_fields: list[str] | None = None,
+        expected_updated_at: str | None = None,
+    ) -> bool:
         """Update an existing server.
+
+        Args:
+            updated_fields: When given, only these fields (plus
+                ``updated_at``) are persisted; concurrent writes to other
+                fields survive (issue #1716). None writes every field.
+            expected_updated_at: When given, the update only lands if the
+                stored ``updated_at`` still equals this value, enforced
+                atomically in the repository write (issue #1716).
 
         Raises:
             UrlValidationError: If the update sets a proxy_pass_url that fails
@@ -282,7 +297,12 @@ class ServerService:
                 # Re-enabling clears any prior refresh auto-disable.
                 server_info["proxy_disabled_reason"] = None
 
-        result = await self._repo.update(path, server_info)
+        result = await self._repo.update(
+            path,
+            server_info,
+            updated_fields=updated_fields,
+            expected_updated_at=expected_updated_at,
+        )
 
         if result:
             # Update search index
--- a/tests/integration/test_server_patch_concurrency.py
+++ b/tests/integration/test_server_patch_concurrency.py
@@ -1,0 +1,176 @@
+"""Persistence-level regression tests for PATCH /servers writes (issue #1716).
+
+Exercises DocumentDBServerRepository.update — the write primitive under
+PATCH /api/servers/{path} — against a real MongoDB: the trailing-slash
+_id asymmetry that made a description-only PATCH fail with a 500, the
+field-scoped $set that stops a metadata PATCH from clobbering a
+concurrent credential/egress write, and the atomic revision predicate
+that backs If-Match. Requires a running MongoDB (the test harness
+points DOCUMENTDB_HOST at localhost); skipped automatically if
+unreachable.
+"""
+
+import uuid
+
+import pytest
+
+from registry.repositories.documentdb.server_repository import (
+    DocumentDBServerRepository,
+)
+
+pytestmark = [pytest.mark.integration, pytest.mark.asyncio]
+
+
+def _card() -> dict:
+    return {
+        "server_name": "legacy",
+        "description": "initial",
+        "tags": ["old"],
+        "license": "MIT",
+        "deployment": "remote",
+        "proxy_pass_url": "http://upstream:9000",
+        "registered_by": "alice",
+        "id": f"asset-{uuid.uuid4().hex[:8]}",
+        "is_enabled": True,
+        "is_active": True,
+        "version": "v1.0.0",
+        "num_tools": 1,
+        "tool_list": [{"name": "search", "inputSchema": {"$schema": "x", "type": "object"}}],
+        "auth_credential_encrypted": "ENC::keepme",
+        "egress_oauth": {"provider": "github", "scopes": ["repo"]},
+        "registered_at": "2026-01-01T00:00:00",
+        "updated_at": "2026-01-01T00:00:00",
+    }
+
+
+@pytest.fixture
+async def repo():
+    """Repository backed by a real collection; cleans up cards it created."""
+    r = DocumentDBServerRepository()
+    try:
+        col = await r._get_collection()
+        await col.database.command("ping")
+    except Exception as e:
+        pytest.skip(f"MongoDB not reachable: {e}")
+    created: list[str] = []
+    r._created = created
+    yield r
+    col = await r._get_collection()
+    if created:
+        await col.delete_many({"_id": {"$in": created}})
+
+
+class TestSlashVariantWrite:
+    async def test_update_resolves_slash_variant_id(self, repo):
+        """A card stored under '/x/' must be writable as '/x' (issue #1716).
+
+        get() falls back to the slash variant when reading, so a
+        read-then-write flow on such a card resolved nothing and the
+        route returned 'Failed to save server'.
+        """
+        col = await repo._get_collection()
+        card = _card()
+        await col.insert_one({"_id": "/legacy-1716/", **card})
+        repo._created.append("/legacy-1716/")
+
+        existing = await repo.get("/legacy-1716")
+        assert existing is not None
+
+        merged = {**existing, "description": "patched"}
+        assert await repo.update("/legacy-1716", merged, updated_fields=["description"])
+
+        doc = await col.find_one({"_id": "/legacy-1716/"})
+        assert doc["description"] == "patched"
+
+
+class TestFieldScopedWrite:
+    async def test_scoped_update_preserves_concurrent_credential_write(self, repo):
+        """A PATCH must not overwrite fields another writer owns (issue #1716).
+
+        The PATCH read its card before the credential rotation landed;
+        the old full-card $set persisted the stale credential.
+        """
+        col = await repo._get_collection()
+        card = _card()
+        await col.insert_one({"_id": "/racy-1716", **card})
+        repo._created.append("/racy-1716")
+
+        stale_read = await repo.get("/racy-1716")
+        await col.update_one(
+            {"_id": "/racy-1716"},
+            {"$set": {"auth_credential_encrypted": "ENC::rotated"}},
+        )
+
+        merged = {**stale_read, "description": "patched"}
+        assert await repo.update("/racy-1716", merged, updated_fields=["description"])
+
+        doc = await col.find_one({"_id": "/racy-1716"})
+        assert doc["description"] == "patched"
+        assert doc["auth_credential_encrypted"] == "ENC::rotated"
+        assert doc["egress_oauth"] == card["egress_oauth"]
+        assert doc["tool_list"] == card["tool_list"]
+
+    async def test_repeated_identical_patch_is_harmless(self, repo):
+        col = await repo._get_collection()
+        card = _card()
+        await col.insert_one({"_id": "/idem-1716", **card})
+        repo._created.append("/idem-1716")
+
+        for _ in range(2):
+            existing = await repo.get("/idem-1716")
+            merged = {**existing, "description": "same"}
+            assert await repo.update("/idem-1716", merged, updated_fields=["description"])
+
+        doc = await col.find_one({"_id": "/idem-1716"})
+        assert doc["description"] == "same"
+        assert doc["num_tools"] == card["num_tools"]
+
+
+class TestRevisionGuardedWrite:
+    async def test_stale_revision_fails_atomically(self, repo):
+        col = await repo._get_collection()
+        card = _card()
+        await col.insert_one({"_id": "/cas-1716", **card})
+        repo._created.append("/cas-1716")
+
+        existing = await repo.get("/cas-1716")
+        # Another writer lands between our read and our write.
+        await col.update_one(
+            {"_id": "/cas-1716"},
+            {
+                "$set": {
+                    "auth_credential_encrypted": "ENC::theirs",
+                    "updated_at": "2030-01-01T00:00:00",
+                }
+            },
+        )
+
+        merged = {**existing, "description": "must not land"}
+        result = await repo.update(
+            "/cas-1716",
+            merged,
+            updated_fields=["description"],
+            expected_updated_at=existing["updated_at"],
+        )
+        assert result is False
+
+        doc = await col.find_one({"_id": "/cas-1716"})
+        assert doc["description"] == "initial"
+        assert doc["auth_credential_encrypted"] == "ENC::theirs"
+
+    async def test_fresh_revision_writes(self, repo):
+        col = await repo._get_collection()
+        card = _card()
+        await col.insert_one({"_id": "/cas2-1716", **card})
+        repo._created.append("/cas2-1716")
+
+        existing = await repo.get("/cas2-1716")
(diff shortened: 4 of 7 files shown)


registry/services/server_service.py
```diff
@@ -282,7 +297,12 @@ async def update_server(self, path: str, server_info: dict[str, Any]) -> bool:
                 # Re-enabling clears any prior refresh auto-disable.
                 server_info["proxy_disabled_reason"] = None
```
omrishiv, line 304:
```suggestion
                server_info["proxy_disabled_reason"] = None
                if updated_fields is not None:
                    updated_fields = [
                        *updated_fields,
                        "proxy_resolved_ips",
                        "proxy_target_host",
                        "proxy_disabled_reason",
                    ]
```

omrishiv (changes requested):
Hi @atirna, thanks for the PR. Please fix the failing tests and see the suggested change

atirna:
pushed the suggestion as-is, pinned proxy fields now join the write scope on field-scoped updates. the 3 failing test_server_service tests were asserting the old 2-arg update() call, updated them for the new kwargs and they pass now. the agent_service one didnt reproduce locally on 3.14 and the diff doesnt touch that path, looks like a runner flake - happy to dig in if the rerun still trips it. also rebased onto latest main since it had drifted a fair bit

registry/services/server_service.py
```diff
@@ -295,13 +317,27 @@ async def update_server(self, path: str, server_info: dict[str, Any]) -> bool:
             # and skip the clear, letting a repoint-before-enable carry the old
             # host's secret to a new host. effective_proxy_target omits that gate,
             # so a host change is caught even while the server is disabled.
+            _pre_repoint_keys = set(server_info)
             clear_upstream_headers_on_repoint(
```
omrishiv, line 321: In the registry/schemas/proxy_mixin.py: 

```
UPSTREAM_HEADER_FIELDS = (
    "custom_headers_encrypted",
    "custom_header_names",
    "custom_header_overridable_names",
    "custom_headers_updated_at",
)

def clear_upstream_headers_on_repoint(updates, existing_target, new_target) -> tuple[str, ...]:
    """... Returns the field names cleared, so a field-scoped caller can widen its write scope."""
    if _target_identity(existing_target) == _target_identity(new_target):
        return ()
    updates["custom_headers_encrypted"] = None
    updates["custom_header_names"] = []
    updates["custom_header_overridable_names"] = []
    updates["custom_headers_updated_at"] = None
    return UPSTREAM_HEADER_FIELDS
```

then this can be

```
            cleared = clear_upstream_headers_on_repoint(
                server_info,
                existing_target=effective_proxy_target("mcp_server", dict(existing or {})),
                new_target=effective_proxy_target("mcp_server", merged),
            )
            if updated_fields is not None and cleared:
                updated_fields = [*updated_fields, *cleared]
```

registry/repositories/documentdb/server_repository.py
```diff
@@ -449,10 +477,31 @@ async def update(
...
+                if expected_updated_at is None:
+                    return {"_id": id_value}
+                return {"_id": id_value, "updated_at": expected_updated_at}
+
+            result = await collection.update_one(_filter_for(path), update_spec)
             if result.matched_count == 0:
-                logger.error(f"Server at '{path}' not found in DocumentDB")
+                alternate_path = path.rstrip("/") if path.endswith("/") else path + "/"
+                if alternate_path != path:
+                    result = await collection.update_one(_filter_for(alternate_path), update_spec)
```
omrishiv, line 496: if `expected_updated_at` is set, `matched_count==0` can mean that the card exists, but that someone else wrote it first. This would then apply the $set to the other _id. Fall back only when the exact `_id` is absent, or pass the stored `_id` that `get()` already returns as `path`.

registry/repositories/interfaces.py
```diff
@@ -128,8 +128,12 @@ async def update(
         self,
         path: str,
         server_info: dict[str, Any],
+        *,
+        updated_fields: list[str] | None = None,
+        expected_updated_at: str | None = None,
     ) -> bool:
-        """Update an existing server."""
+        """Update an existing server (see the DocumentDB impl for the
```
omrishiv, line 135: please update the contract here

omrishiv (changes requested):
Hi @atirna thank you for updating the PR so quickly, a few other asks:

- Can we add tests to the PUT path? that's the explicit one handled in the PR, but there are no tests for it
- the delete, set_state, and update_fields in `server_repository` are unhandled
- There are a lot of references in the comments to this pr/issue. please thin those down a bit

registry/services/server_service.py
atirna: moved as suggested: the tuple lives in proxy_mixin beside the helper, the return carries the cleared names, and the service widens its scope from that return

registry/repositories/documentdb/server_repository.py
atirna: the fallback now probes for the exact _id first: a guarded miss on a card that exists returns False without trying the variant, so a lost race cant land the $set on the other card. regression covers it with a variant card seeded at the other _id

atirna:
pushed 9770fe35 with the rest:
- PUT revision-guard tests: 412 when the guarded write loses the race, the revision riding the write with a matching If-Match, unguarded stays full-field
- delete, set_state and update_field match either slash variant now, same as get/update
- pr/issue references in comments thinned to one

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+        "version": "v1.0.0",
+        "num_tools": 1,
+        "tool_list": [{"name": "search", "inputSchema": {"$schema": "x", "type": "object"}}],
+        "auth_credential_encrypted": "ENC::keepme",
+        "egress_oauth": {"provider": "github", "scopes": ["repo"]},
+        "registered_at": "2026-01-01T00:00:00",
+        "updated_at": "2026-01-01T00:00:00",
+    }
+
+
```
omrishiv:
```suggestion
def _path(label: str, *, trailing_slash: bool = False) -> str:
    """Unique _id per run: concurrent CI jobs share one MongoDB."""
    return f"/{label}-{uuid.uuid4().hex[:8]}{'/' if trailing_slash else ''}"


```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+class TestSlashVariantWrite:
+    async def test_update_resolves_slash_variant_id(self, repo):
+        """A card stored under '/x/' must be writable as '/x'."""
+        col = await repo._get_collection()
+        card = _card()
+        await col.insert_one({"_id": "/legacy-variant/", **card})
+        repo._created.append("/legacy-variant/")
+
+        existing = await repo.get("/legacy-variant")
```
omrishiv:
```suggestion
        stored_id = _path("legacy-variant", trailing_slash=True)
        await col.insert_one({"_id": stored_id, **_card()})
        repo._created.append(stored_id)
        requested = stored_id.rstrip("/")

        existing = await repo.get(requested)
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+        assert existing is not None
+
+        merged = {**existing, "description": "patched"}
+        assert await repo.update("/legacy-variant", merged, updated_fields=["description"])
```
omrishiv:
```suggestion
        assert await repo.update(requested, merged, updated_fields=["description"]) 
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+
+        doc = await col.find_one({"_id": "/legacy-variant/"})
```
omrishiv:
```suggestion
        doc = await col.find_one({"_id": stored_id}) 
```

registry/services/server_service.py
omrishiv:
```suggestion

        Note:
            Every branch below that writes into ``server_info`` MUST also extend
            ``updated_fields`` when it is not None. A field-scoped write persists
            only the named fields, so an unnamed mutation is silently discarded
            -- which is how a re-pinned proxy target once kept its stale IPs and
            a repointed server once kept the previous host's credential. Prefer a
            helper that returns the names it wrote (see
            ``clear_upstream_headers_on_repoint``) over restating them here.
        """
```

registry/repositories/documentdb/server_repository.py
omrishiv:
```suggestion
    @staticmethod
    def _path_variants(path: str) -> list[str]:
        """Both trailing-slash forms of a server path, the caller's form first.

        Server ``_id``s are stored inconsistently (e.g. ``/airegistry-tools/``),
        and callers may pass a bare name. Matching both forms in one query is
        equivalent to the alternate-path fallback in ``get``, without the
        second round trip.

        The caller's own spelling sorts first so a sequential resolver (``get``,
        ``update``) probes the most likely ``_id`` before its variant and wins
        deterministically when both documents exist; an ``$in`` filter ignores
        the order.
        """
        stem = "/" + path.strip("/")
        variants = [stem, stem + "/"]
        if path in variants:
            variants.remove(path)
            variants.insert(0, path)
        return variants
```

registry/repositories/documentdb/server_repository.py
```diff
@@ -449,10 +477,34 @@ async def update(
...
-            result = await collection.update_one({"_id": path}, update_spec)
 
+            # Exact _id first, then its slash variant, each with the same
+            # revision predicate; a miss on both fails the write.
+            def _filter_for(id_value: str) -> dict[str, Any]:
+                if expected_updated_at is None:
+                    return {"_id": id_value}
+                return {"_id": id_value, "updated_at": expected_updated_at}
+
+            result = await collection.update_one(_filter_for(path), update_spec)
```
omrishiv:
```suggestion
            primary, *alternates = self._path_variants(path)
            result = await collection.update_one(_filter_for(primary), update_spec)
```

registry/repositories/documentdb/server_repository.py
```diff
@@ -449,10 +477,34 @@ async def update(
...
             if result.matched_count == 0:
```
omrishiv:
```suggestion
            if result.matched_count == 0 and alternates:
```

registry/repositories/documentdb/server_repository.py
```diff
@@ -449,10 +477,34 @@ async def update(
...
+                alternate_path = path.rstrip("/") if path.endswith("/") else path + "/"
+                if alternate_path != path:
+                    if expected_updated_at is not None and await collection.find_one(
+                        {"_id": path}, {"_id": 1}
+                    ):
+                        # The exact card exists with a moved revision: a lost
+                        # race, never a variant miss.
+                        logger.info(f"Server at '{path}' revision mismatch; concurrent update won")
+                        return False
+                    result = await collection.update_one(_filter_for(alternate_path), update_spec)
```
omrishiv:
```suggestion
                if expected_updated_at is not None and await collection.find_one(
                    {"_id": primary}, {"_id": 1}
                ):
                    # The card exists with a moved revision: a lost race, never a
                    # variant miss. Writing another _id would retarget the write.
                    logger.info(f"Server at '{path}' revision mismatch; concurrent update won")
                    return False
                result = await collection.update_one(_filter_for(alternates[0]), update_spec)
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+        merged = {**stale_read, "description": "patched"}
+        assert await repo.update("/racy-scope", merged, updated_fields=["description"])
+
+        doc = await col.find_one({"_id": "/racy-scope"})
+        assert doc["description"] == "patched"
+        assert doc["auth_credential_encrypted"] == "ENC::rotated"
+        assert doc["egress_oauth"] == card["egress_oauth"]
+        assert doc["tool_list"] == card["tool_list"]
```
omrishiv:
```suggestion
        col = await repo._get_collection()
        card = _card()
        path = _path("racy-scope")
        await col.insert_one({"_id": path, **card})
        repo._created.append(path)

        stale_read = await repo.get(path)
        await col.update_one(
            {"_id": path},
            {"$set": {"auth_credential_encrypted": "ENC::rotated"}},
        )

        merged = {**stale_read, "description": "patched"}
        assert await repo.update(path, merged, updated_fields=["description"])

        doc = await col.find_one({"_id": path})
        assert doc["description"] == "patched"
        assert doc["auth_credential_encrypted"] == "ENC::rotated"
        assert doc["egress_oauth"] == card["egress_oauth"]
        assert doc["tool_list"] == card["tool_list"]
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+        repo._created.append("/idem-scope")
+
+        for _ in range(2):
+            existing = await repo.get("/idem-scope")
+            merged = {**existing, "description": "same"}
+            assert await repo.update("/idem-scope", merged, updated_fields=["description"])
+
+        doc = await col.find_one({"_id": "/idem-scope"})
+        assert doc["description"] == "same"
+        assert doc["num_tools"] == card["num_tools"]
```
omrishiv:
```suggestion
        col = await repo._get_collection()
        card = _card()
        path = _path("idem-scope")
        await col.insert_one({"_id": path, **card})
        repo._created.append(path)

        for _ in range(2):
            existing = await repo.get(path)
            merged = {**existing, "description": "same"}
            assert await repo.update(path, merged, updated_fields=["description"])

        doc = await col.find_one({"_id": path})
        assert doc["description"] == "same"
        assert doc["num_tools"] == card["num_tools"]
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+            "/cas-race",
+            merged,
+            updated_fields=["description"],
+            expected_updated_at=existing["updated_at"],
+        )
+        assert result is False
+
+        doc = await col.find_one({"_id": "/cas-race"})
+        assert doc["description"] == "initial"
+        assert doc["auth_credential_encrypted"] == "ENC::theirs"
```
omrishiv:
```suggestion
        col = await repo._get_collection()
        card = _card()
        path = _path("cas-race")
        await col.insert_one({"_id": path, **card})
        repo._created.append(path)

        existing = await repo.get(path)
        # Another writer lands between our read and our write.
        await col.update_one(
            {"_id": path},
            {
                "$set": {
                    "auth_credential_encrypted": "ENC::theirs",
                    "updated_at": "2030-01-01T00:00:00",
                }
            },
        )

        merged = {**existing, "description": "must not land"}
        result = await repo.update(
            path,
            merged,
            updated_fields=["description"],
            expected_updated_at=existing["updated_at"],
        )
        assert result is False

        doc = await col.find_one({"_id": path})
        assert doc["description"] == "initial"
        assert doc["auth_credential_encrypted"] == "ENC::theirs"
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+        existing = await repo.get("/cas-race-2")
+        merged = {**existing, "description": "landed"}
+        assert await repo.update(
+            "/cas-race-2",
+            merged,
+            updated_fields=["description"],
+            expected_updated_at=existing["updated_at"],
+        )
+        doc = await col.find_one({"_id": "/cas-race-2"})
+        assert doc["description"] == "landed"
```
omrishiv: `card` is unused here once the id is generated, so this drops it too.

```suggestion
        col = await repo._get_collection()
        path = _path("cas-fresh")
        await col.insert_one({"_id": path, **_card()})
        repo._created.append(path)

        existing = await repo.get(path)
        merged = {**existing, "description": "landed"}
        assert await repo.update(
            path,
            merged,
            updated_fields=["description"],
            expected_updated_at=existing["updated_at"],
        )
        doc = await col.find_one({"_id": path})
        assert doc["description"] == "landed"
```

tests/integration/test_server_patch_concurrency.py
```diff
@@ -0,0 +1,198 @@
...
+            updated_fields=["description"],
+            expected_updated_at=existing["updated_at"],
+        )
+        assert result is False
+
+        doc = await col.find_one({"_id": "/cas-variant-race"})
+        assert doc["description"] == "initial"
+        variant = await col.find_one({"_id": "/cas-variant-race/"})
+        assert variant["description"] == "initial"
+        assert variant["server_name"] == "variant occupant"
```
omrishiv:
```suggestion
        col = await repo._get_collection()
        path = _path("cas-variant-race")
        variant_path = path + "/"
        await col.insert_one({"_id": path, **_card()})
        await col.insert_one({"_id": variant_path, **_card(), "server_name": "variant occupant"})
        repo._created.extend([path, variant_path])

        existing = await repo.get(path)
        # Another writer lands on the exact card between our read and write.
        await col.update_one(
            {"_id": path},
            {"$set": {"updated_at": "2030-01-01T00:00:00"}},
        )

        merged = {**existing, "description": "must not land"}
        result = await repo.update(
            path,
            merged,
            updated_fields=["description"],
            expected_updated_at=existing["updated_at"],
        )
        assert result is False

        doc = await col.find_one({"_id": path})
        assert doc["description"] == "initial"
        variant = await col.find_one({"_id": variant_path})
        assert variant["description"] == "initial"
        assert variant["server_name"] == "variant occupant"
```

registry/repositories/documentdb/server_repository.py
omrishiv:
```suggestion
                if expected_updated_at is not None and await collection.find_one(
                    {"_id": primary}, {"_id": 1}
                ):
                    # The card exists with a moved revision: a lost race, never a
                    # variant miss. Writing another _id would retarget the write.
                    logger.info(f"Server at '{path}' revision mismatch; concurrent update won")
                    return False
                result = await collection.update_one(_filter_for(alternates[0]), update_spec)
```

tests/integration/test_server_patch_concurrency.py
omrishiv:
```suggestion
        stored_id = _path("legacy-variant", trailing_slash=True)
        await col.insert_one({"_id": stored_id, **_card()})
        repo._created.append(stored_id)
        requested = stored_id.rstrip("/")

        existing = await repo.get(requested)
```

registry/repositories/documentdb/server_repository.py
omrishiv: `get()` is the last method still building its own variant pair, so it disagrees with `delete` / `set_state` / `update_field` on non-canonical input:

| `path` | `_path_variants(path)` | what `get()` tries today |
|---|---|---|
| `/a` | `['/a', '/a/']` | `['/a', '/a/']` |
| `/a/` | `['/a', '/a/']` | `['/a/', '/a']` |
| `a` | `['/a', '/a/']` | `['a', 'a/']` |
| `//a//` | `['/a', '/a/']` | `['//a//', '//a']` |

`delete("a")` finds the card stored at `/a`; `get("a")` does not. The routes normalise via `_normalize_server_path`, but the other ~16 `update_server` callers (federation sync, health service, startup, peer federation) build paths from peer payloads and stored data.

Sequential rather than one `$in` query: `$in` would halve the round trips on a miss but lets Mongo choose when both variant documents exist, and `get()` feeding `update()` is exactly where that choice must be deterministic. Round trips are otherwise unchanged — 1 on a hit at the caller's `_id`, 2 on a variant hit or a miss. `enumerate` + `if index` preserves the existing "Trying alternate path" debug line that a plain loop drops.

```suggestion
    async def get(
        self,
        path: str,
    ) -> dict[str, Any] | None:
        """Get server by path.

        Matches either trailing-slash form of the stored ``_id``, the caller's
        spelling first.
        """
        logger.debug(
            f"DocumentDB READ: Getting server with path='{path}' from collection '{self._collection_name}'"
        )
        collection = await self._get_collection()

        try:
            server_info = None
            for index, candidate in enumerate(self._path_variants(path)):
                if index:
                    logger.debug(f"DocumentDB READ: Trying alternate path '{candidate}'")
                server_info = await collection.find_one({"_id": candidate})
                if server_info:
                    break
```

omrishiv: Two methods now answer "which `_id` holds this card" differently: `delete` / `set_state` / `update_field` / the tool-override methods go through `_path_variants()`, which normalises the stem, while `get()` and `update()` build their own pair, which does not. Adding the ordering guarantee here lets both of those route through this helper (see the `get()` comment at L101 and the `update()` ones at L488-499) without losing "the caller's own `_id` is tried first", which is what keeps a duplicate-variant card from making a guarded write non-deterministic.

`$in` call-sites are unaffected — Mongo ignores the order of an `$in` list.

```suggestion
    @staticmethod
    def _path_variants(path: str) -> list[str]:
        """Both trailing-slash forms of a server path, the caller's form first.

        Server ``_id``s are stored inconsistently (e.g. ``/airegistry-tools/``),
        and callers may pass a bare name. Matching both forms in one query is
        equivalent to the alternate-path fallback in ``get``, without the
        second round trip.

        The caller's own spelling sorts first so a sequential resolver (``get``,
        ``update``) probes the most likely ``_id`` before its variant and wins
        deterministically when both documents exist; an ``$in`` filter ignores
        the order.
        """
        stem = "/" + path.strip("/")
        variants = [stem, stem + "/"]
        if path in variants:
            variants.remove(path)
            variants.insert(0, path)
        return variants
```

omrishiv (commented):
Thank you @atirna, just a few small comments that should help round out this PR

PR #1744, opened 2026-09-10
fix(dedup): strip catalog boilerplate before embedding, and report degraded similarity

### LINKED ISSUE #1696: Duplicate-check advisory produces false positives on names sharing a common catalog suffix (e.g. "-mcp-server")

aarora79:
Follow-up to #1721, which fixed the dominant defect behind #1696 — the advisory gated on a min-max normalized RRF rank, so the top hit of any search cleared the threshold. This closes the case the issue actually reported, plus the two gaps #1721 named in its own description.

## The problem #1721 left

With a thin description, `topic-mcp-server` measures **0.7621** cosine against `category-mcp-server` on the shared `-mcp-server` suffix alone, which clears the 0.7 default. Sentence models weight tokens roughly by how much of the string they fill, and two of three words in that name are boilerplate every registry carries.

## Changes

**Strip catalog boilerplate from the query before it is embedded.** New `registry/utils/text_normalize.py` drops `mcp`, `server`, `tool`, `skill`, `agent`, `api`, `service` and their plurals. The reported pair falls from 0.76 to **0.47**. Stripping is query-side only, so no stored embedding is rebuilt — measurement on `all-MiniLM-L6-v2` showed query-only stripping separates unrelated pairs (max 0.3717) from true duplicates (min 0.7074) as cleanly as stripping both sides.

`gateway` and `registry` are deliberately absent from the list: stripping them dropped two genuine near-duplicates from 0.6230 and 0.6810 to 0.4633 and 0.4518, turning both into misses.

**Refuse a query left with fewer than two distinct tokens.** After stripping, `topic-mcp-server` is the single token `topic`, which cannot clear a usable threshold against any described entry (`topic` scores 0.2603 against its own entry). The embedding call is skipped rather than spent. Those registrations stay covered by the exact-URL check, which is the strong duplicate signal.

**Report `similarity_search_available: false` when hits carry no similarity at all.** That state means the search backend fell back to keyword-only matching because the embedder was unreachable. The frontend keys its degraded-mode notice off that flag (`useDuplicateCheck.ts:138`), so an empty advisory would otherwise read to the user as "nothing similar exists". One scored hit is enough to prove the embedder answered; unscored hits are still dropped.

**`DEDUP_SCORE_THRESHOLD` 0.7 → 0.6**, calibrated on `all-MiniLM-L6-v2` where unrelated pairs measure 0.07-0.37 and true duplicates 0.71-0.87. It ships in the same commit as stripping on purpose: 0.6 was calibrated on stripped text, and lowering the gate while the text still carries `-mcp-server` would admit more noise, not less.

**One INFO line per check** recording candidates, scored count, top cosine, threshold, and advised count. The advisory now says nothing most days, and without this line an operator cannot tell "the threshold is working" from "the embedder is down".

## Configuration surface

The three `DEDUP_*` variables were absent from all three compose files, so on Docker `DEDUP_SCORE_THRESHOLD` in `.env` never reached the container and the registry ran on code defaults. They are wired now, which repairs the three-surface parity invariant (theory §2.5). The new default is set on Docker, Terraform (root + module + tfvars example), and Helm (both values files).

Group 20a of `docs/unified-parameter-reference.md` documented six parameters that do not exist in `Settings` (`DEDUP_ENABLED`, `DEDUP_CANDIDATE_POOL_SIZE`, three `DEDUP_WEIGHT_*`, `DEDUP_MAX_TEXT_CHARS`), an endpoint that does not exist (`POST /servers/similar`), and a weight-sum validation rule that is not in the code. Rewritten to the three parameters that exist, with the cosine semantics stated.

No new parameter is added: the boilerplate list and the two-token floor are module constants, so there is no fourth surface to wire.

## Verification

- `pytest tests/unit -n 8`: **6839 passed, 8 skipped**. The one local failure, `test_mcp_token_ttl_defaults`, is environmental — this workspace's `.env:501` sets `MCP_TOKEN_DEFAULT_TTL_HOURS=2` while the test asserts the code default of 8. It fails the same way on `main`.
- `ruff check` and `ruff format --check` clean at the pinned v0.11.13.
- New: 13 cases in `tests/unit/utils/test_text_normalize.py`, 7 in the service suite covering the stripped query, the reported pair never reaching the backend, an all-boilerplate name, a two-token name that still searches, and the three availability states.
- Three existing tests were updated rather than worked around: two in `TestQueryComposition` asserted the raw unstripped query, and two fixtures used names (`My Server`, `New Server`) that are single-token after stripping and so no longer reach the search call. They now carry real signal, since they test threshold, cap, and exclusion behaviour rather than the guard.

## Release note text

> The registration duplicate-check advisory now compares cosine similarity of the text left after catalog boilerplate (`mcp`, `server`, `skill`, ...) is stripped, and `DEDUP_SCORE_THRESHOLD` drops from 0.7 to 0.6 to match that scale. Expect far fewer advisory matches: before 1.30.0 the check returned suggestions on nearly every registration regardless of what was typed. A registration whose name carries no signal beyond boilerplate gets no advisory at all and relies on the exact-URL check. On Docker, the three `DEDUP_*` variables now actually reach the container — check `.env` for stale values before upgrading.

Closes #1696

## Not in this PR

- The cosine is still computed for every search caller, not just the advisory (0.5-2.2 ms per `/api/search/semantic` request at `max_results=10`, and the MongoDB CE path already computes the same value at `search_repository.py:1812`). Separate PR, since it needs its own before/after numbers.
- `docs/registration-duplicate-check.md`, and the frontend label and borderline styling in `DuplicateCheckModal.tsx`.
- Making the boilerplate token list operator-tunable. Left as a constant until a deployment with org-prefixed names asks for it.


aarora79:
Executed the test plan (`.scratchpad/pr-1721/testing.md`) against a live registry on this branch. 25 of 27 checks passed as written; two failed against my own expectations and produced a threshold change, pushed as `438ec0fd`.

**Deployment under test:** `BUILD_VERSION=1.30.0-3-ga96d803d-fix/1696-dedup-precision`, `STORAGE_BACKEND=mongodb-ce` (so `_client_side_search`), `all-MiniLM-L6-v2`. The registry already held **149 servers / 404 indexed documents** of real catalog data, so probes ran against a realistic corpus, not just the 7 fixtures.

## The reported case, fixed

```
cosine('topic-mcp-server', 'category-mcp-server')  = 0.7621   <- what the pre-fix code embedded
cosine('topic',            'category')             = 0.4692   <- what this PR embeds
```

```
POST /api/servers/check-duplicates  {"name":"topic-mcp-server"}
-> {"advisory_matches":[], "threshold":0.45, "similarity_search_available":true}
   log: Duplicate-check similarity skipped: query carries 1 distinct token(s)
        after boilerplate stripping, minimum is 2.
```

No embedding call at all on that path.

## Degraded mode (the gap #1721 left)

With `EMBEDDINGS_API_BASE` pointed at a dead port, search falls back to keyword-only:

```
-> {"advisory_matches":[], "similarity_search_available":false}
   log: Duplicate check received 12 search hits with no similarity scores; treating
        similarity search as unavailable. First hits: ['/payroll-mcp-server',
        '/com-github-github-mcp-server', ...]
```

## End-to-end wiring, which the review flagged as untested

Every summary line reads `N candidates, N scored`, e.g. `28 candidates, 28 scored, top cosine 0.8024, threshold 0.60, 1 advised`. That proves `similarity_score` survives `search()` → `SemanticSearchService` → the advisory on the live MongoDB CE path.

Cross-entity works: an agent registration surfaced `/topic-extractor-mcp-server@0.835` (mcp_server) alongside `/topic-extraction-agent@0.824` (a2a_agent). Search integrity held: **0** occurrences of `similarity_score` in the `/api/search/semantic` response, top hit still normalizes to `1.0`, all six result groups present, ~130 ms per search.

Round-trip re-registration behaved as documented: re-registering a dumped config at the same path returns **409 Conflict**, while a clone with an identical name, description, and `proxy_pass_url` at a different path **registers successfully** — content duplication is advisory-only, and the CLI never consults the check.

## Why 0.6 became 0.45

P3 and P5 exposed a calibration error. The design measured 0.6 on symmetric stripped strings; the deployed comparison is asymmetric — the query is stripped, but candidates are compared on their **indexed** text, which keeps its boilerplate and also carries `Tags: ...` and flattened metadata. That costs about 0.2 of cosine.

| Pair (stripped query vs real indexed text) | Cosine | At 0.6 | At 0.45 |
|---|---|---|---|
| identical name and description | 0.6042 | passes by 0.004 | passes |
| genuine near-duplicate (topic extractor) | 0.8024 | passes | passes |
| genuine near-duplicate (topic miner) | 0.5371 | **missed** | passes |
| genuine near-duplicate (weather radar/forecast) | 0.5104 | **missed** | passes |
| unrelated (topic vs category) | 0.1652 | rejected | rejected |
| unrelated (topic vs payroll) | 0.0645 | rejected | rejected |

An exact content duplicate clearing the gate by four thousandths is not a working default — two more tags on that server and it stops being detected. Unrelated pairs top out at 0.25, so 0.45 sits mid-band. Verified at 0.45 against the real catalog: the reported case still advises nothing, an unrelated registration still advises nothing, and the extra hits are same-domain entries worth seeing (a docs-search registration surfaced `agentcore-aws-kb@0.569`, `aws-kb@0.551`, `context7@0.500`).

**Stripping the document side too was measured and rejected:** symmetric stripping moves these pairs by at most +0.06 (0.6042 → 0.6642), and the weather pair actually drops to 0.4999. The dilution is tags and metadata, not boilerplate, so a corpus re-index buys nothing.

## Three findings worth separate attention

1. **An all-boilerplate name logs nothing.** It returns on the `if not query` branch, above the guard that carries the log line. Behaviour is correct; the operator gets no trace. One line, happy to add it here.
2. **Pre-existing, not this PR:** newly registered entities are invisible to the advisory until promoted. All five seeded servers registered `is_enabled=false` (unreachable backends), and both agents registered `status="draft"` — search excludes both, so a genuine 0.8313 match stayed hidden until `toggle` and `PATCH {"status":"active"}`.
3. **`cli/examples/code_reviewer_agent.json` is stale** — `provider` must be `{organization, url}`, and `supportedProtocol` is required. That example fails validation as written.

`tests/unit`: 6840 passed, 8 skipped after recalibration. `ruff check` and `format --check` clean. All 9 fixtures removed from the registry and the search index verified empty of them.

Not covered: the UI spot-check (no browser session against this deployment) and the DocumentDB hybrid RRF branch, which needs an ECS environment.

Issue #1745, opened 2026-09-11
MCP proxy hop does not apply the server `auth_scheme` credential or `custom_headers` (FAQ says it does); no operator-side way to set User-Agent on egress


peelandsee:
## Summary

On the MCP proxy hop (`auth_server` `mcp_proxy`), a registered server's **Backend Authentication** credential (`auth_scheme` + `auth_credential`) and its **custom headers** (`custom_headers`, #1028) are **not** applied to the upstream request. They are only used by the registry for health checks and tool discovery (`registry/core/mcp_client.py`, `registry/health/service.py`).

Three consequences for operators fronting a third-party MCP server that expects a shared, operator-owned credential or a fixed header:

1. The FAQ `docs/faq/registering-auth-protected-servers.md` still states "**Request proxying** -- When clients connect through the gateway, the credential is injected into proxied requests". That is not what the code does, so operators configure `auth_scheme=bearer`, see the server turn healthy, and then every proxied call fails at the upstream.
2. `custom_headers` were proposed in #1028 with "Also injected outbound when the gateway proxies to the backend MCP server" and with "Required `User-Agent` … values some backends enforce" as a motivating case. PR #1041 implemented registration, encryption, health/discovery and the Connect JSON, but touched no `auth_server` file, so the outbound part never landed.
3. The 1.30.0 generic HTTP proxy (#1628) introduced a **shared operator key** credential model with operator-injected headers (`_vend_generic_upstream_headers` / `_merge_generic_upstream_headers`). The MCP hop has no equivalent, so the same upstream behaves differently depending on which hop fronts it.

I understand from #1265 that `auth_scheme` was redefined as "the gateway acting as itself" (health checks) and that per-user egress is the preferred model (`egress_auth_mode` pat / oauth_user / obo_exchange). This issue is not asking to weaken that: it asks for (a) the documented behaviour to match the code, and (b) an explicit, opt-in way to apply operator-owned headers on the MCP hop, as the generic hop already allows.

## Environment

- MCP Gateway Registry 1.29.x, Helm (EKS-style deployment), `AUTH_PROVIDER=entra`, `EGRESS_AUTH_ENABLED` unset (default off).
- Upstream: a vendor MCP server that requires two fixed headers (an API key and a protocol-signal header), hosted behind Cloudflare with Browser Integrity Check enabled.

## Observed

[...]

Issue #1266, opened 2026-06-19
Egress: ingress-token relay fix (relay-ingress + none)


aarora79:
Part of #1265 (Egress Token Credentials Broker). **Sub-issue 1 of 5. Implement first.**

## Problem

In the auth-server MCP-proxy egress hop, the inbound `Authorization` header is copied through to the upstream backend unchanged. For third-party backends (e.g. GitHub) this leaks the gateway **ingress** token (the token the IDE used to log into the gateway), which the backend rejects (GitHub returns `400 invalid_token`). In-trust backends rely on the relay, so the fix must be per-server.

## Proposed solution

Pure bugfix, no schema change, no config field. In the MCP-proxy egress hop, relay the inbound `Authorization` **only** for a hardcoded in-trust allowlist of server names; **strip** it for every other backend. Expressed as a module-level frozenset in the auth-server. Behaviorally `relay-ingress` for the allowlist, `none` for everyone else. The relay decision keys on the cryptographically verified, path-validated server name from the signed internal-hop token claims (fail-closed: a missing claim strips).

## Acceptance criteria

- [ ] Ingress `Authorization` is stripped on egress for any backend not in the in-trust allowlist.
- [ ] In-trust backends (the always-healthy MCP server) still receive the relayed token and stay reachable through the full proxy chain.
- [ ] No new config parameter, schema field, or dependency is introduced.
- [ ] A third-party backend no longer receives the gateway ingress token (verified via a mock upstream or observation).
- [ ] The egress path logs no token material.

## Out of scope

Making the relay decision a configurable per-server setting (that arrives with sub-issue 3, which introduces `egress_auth_mode`).

## Notes

The in-trust allowlist must be verified against the live registry, not assumed to contain only one server. Health checks and the REST API do not exercise the MCP-proxy path, so the proxy-chain reachability check must be a required gate on Docker, ECS, and EKS.

aarora79:
## Must-fix checklist (from design review)

Implementation-time requirements surfaced in review (no design change, just must-do):

- [ ] Confirm the complete in-trust inventory against the **live registry**, not just docs; seed the allowlist with every backend that currently relies on the relayed ingress token (expected: only `airegistry-tools`).
- [ ] No-leak test: forwarded headers to a non-allowlisted upstream contain no `Authorization`.
- [ ] Pin tests: `airegistry-tools` is in the allowlist; assert no allowlist entry contains a `/`.
[...]

Issue #1265, opened 2026-06-19
Umbrella: Egress Token Credentials Broker


aarora79:
## Problem

When an AI agent (Copilot CLI, Claude Code, Cursor) calls a third-party MCP server (GitHub, Slack, Atlassian, Outlook) **through the gateway**, that backend is a pure resource server: it validates a token its own IdP issued and expects it in `Authorization: Bearer`. Today the gateway relays the **ingress** token (the token the IDE used to log into the gateway) straight to the backend, which rejects it (e.g. GitHub returns `400 invalid_token`). The only workaround is pasting a long-lived provider token into `mcp.json` on every developer machine: no refresh, no rotation, plaintext secret in a client file, not auditable per user.

This umbrella builds a **managed egress credentials broker**: one per-server config setting selects how the gateway authenticates to each backend on the user's behalf, backed by a pluggable secret store. The agent authenticates **once** to the gateway; the gateway supplies the right backend credential per server. No hardcoded tokens in client config.

## The unifying design: `egress_auth_mode`

One optional per-server field on the registered server, defaulting to preserve today's behavior. Every server has a mode.

| Value | Gateway sends on egress | Vault used? | Sub-issue |
|-------|-------------------------|-------------|-----------|
| `relay-ingress` | the user's ingress Authorization, unchanged (in-trust backends only) | no | 1 |
| `none` | nothing (strip, inject nothing) | no | 1 |
| `vault-pat` | a per-user static PAT/API key from the vault | yes | 3 |
| `token-exchange` | the ingress token exchanged for a backend-audience token (OBO) | no | 4 |
| `vault-oauth` | a per-user OAuth token from the vault (3LO dance) | yes | 5 |

Key properties:

- **Injection point:** the auth-server MCP-proxy egress hop. The broker reads the already-verified user identity from the signed internal-hop token claims (header forgery on that hop is already closed by the signed inter-component tokens shipped in #1260 / #1262).
- **Vault key:** `(provider, sub, registered_server_name)`. `sub` is the stable identity half; the registered server name is the destination half (stable across URL changes).
- **Pluggable secret store:** OpenBao for self-hosted (EKS + Docker Compose) and local dev (dev mode), AWS Secrets Manager for ECS. **Token material never lives in the application database.**
[...]

PR #1747, opened 2026-09-11
docs(faq): backend credential is used for health checks and discovery, not injected on the MCP proxy hop

peelandsee:
## What

`docs/faq/registering-auth-protected-servers.md` said that the Backend Authentication credential is "injected into proxied requests". On the auth-server MCP proxy hop the client's ingress auth headers are stripped and no server-level credential is added unless an `egress_auth_mode` is configured (#1265, #1266). The credential is used by the registry for health checks and tool discovery only (`registry/core/mcp_client.py`, `registry/health/service.py`).

This PR rewrites bullet 3 of "How Tool Discovery Works with Auth" and adds a short pointer to the egress credential broker docs, so operators fronting a third-party MCP server do not expect the stored credential to reach the upstream at runtime.

## Why

Operators (us included) configure `auth_scheme=bearer`, see the server become healthy, then get 401 from the upstream on every proxied call and spend time debugging the wrong layer. See issue #1745 for the full analysis and the follow-up proposal (apply `custom_headers` and an opt-in operator credential on the MCP hop).

## Checklist

- [x] Docs only, no code change.
- [x] `ruff` not applicable.
- [x] Links checked against current `main`.

### DIFF
--- a/docs/faq/registering-auth-protected-servers.md
+++ b/docs/faq/registering-auth-protected-servers.md
@@ -51,9 +51,11 @@ Once registered with credentials, the registry automatically:
 
 1. **Health checks** -- Injects the decrypted credential when checking if the server is reachable
 2. **Tool discovery** -- Uses the credential to call the MCP `tools/list` method on the backend server
-3. **Request proxying** -- When clients connect through the gateway, the credential is injected into proxied requests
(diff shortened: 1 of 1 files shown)


omrishiv:
Hi @peelandsee, thank you for the added clarification. I think we'd like to fix this properly. We have #1781 to fix oauth 2.0, 2.1, and OBO backended servers and #1686 for a shared credential. If these do not cover your use case, can you please open an issue so we can address? I'll close if that's ok and we can reopen if you'd like to rescope based on those PRs. Thank you

PR #1748, opened 2026-09-11
fix(prm): resolve RFC 9728 path-aware per-server PRM under a path-prefixed registry_url

### LINKED ISSUE #1746: Per-server PRM: RFC 9728 path-aware URL returns 404 when registry_url has a path prefix (ROOT_PATH)
## Summary

When the gateway runs under a path prefix (`ROOT_PATH=/registry`, `REGISTRY_URL=https://gw.example.com/registry`), the per-server PRM route does not resolve the RFC 9728 §3.1 path-aware URL that a client derives from the advertised `resource`.

- Advertised per-server resource (Entra, #990): `https://gw.example.com/registry/my-server/mcp`
- RFC 9728 §3.1 well-known URL for that resource (well-known segment inserted between host and path): `https://gw.example.com/.well-known/oauth-protected-resource/registry/my-server/mcp`
- URL the gateway advertises in `WWW-Authenticate` and serves: `https://gw.example.com/registry/.well-known/oauth-protected-resource/my-server/mcp`

Clients that follow `resource_metadata` from the 401 header work (this is what the hosted coding agent we tested does first, so it was not blocked by this). Clients that derive the well-known URL from the resource, which RFC 9728 §3.1 allows and which the same client uses as its second attempt, hit the first URL and get `404 {"detail":"no per-server resource metadata"}`, because `_normalize_prm_server_path()` in `registry/api/wellknown_routes.py` only strips a trailing `/mcp` and turns `registry/my-server/mcp` into `/registry/my-server`, which is not a registered server. For such a client the path-aware form is a dead end and the discovery falls back to legacy behaviour.

## Reproduce (inside the registry container, path-prefixed deployment)

```
curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:7860/.well-known/oauth-protected-resource/my-server/mcp            # 200
curl -s -o /dev/null -w "%{http_code}\n" http://127.0.0.1:7860/.well-known/oauth-protected-resource/registry/my-server/mcp   # 404
```

## Proposal

1. In `_normalize_prm_server_path()`, strip the path component of `settings.registry_url` when the incoming `server_path` starts with it (`registry/my-server/mcp` → `/my-server`). No change for deployments whose `registry_url` has no path. PR attached.
2. Docs: note in `docs/oauth-discovery-endpoints.md` that in path-prefixed deployments the root `/.well-known/` location must be routed to the registry by the operator's ingress as well (the bundled nginx already has `location ^~ /.well-known/`, but an external ingress that only forwards `/<ROOT_PATH>` will 404 the §3.1 form before it reaches nginx).
3. Optional follow-up: advertise the §3.1 form in `WWW-Authenticate` when `registry_url` has a path, so both derivations agree byte-for-byte with the served document.

## Acceptance criteria

- [ ] `GET /.well-known/oauth-protected-resource/<root_path>/<server>/mcp` returns the same document as `GET /.well-known/oauth-protected-resource/<server>/mcp` when `registry_url` carries `<root_path>`.
- [ ] Behaviour unchanged when `registry_url` has no path.
- [ ] Unit test in `tests/unit/api/test_wellknown_routes.py`.

Related: #989, #990, #988 (path-aware discovery), #995.

peelandsee:
## What

`_normalize_prm_server_path()` now strips the path component of `settings.registry_url` (for example `/registry`) from the incoming `server_path` before the existing `/mcp` trimming. With `REGISTRY_URL=https://gw.example.com/registry`:

- `GET /.well-known/oauth-protected-resource/registry/my-server/mcp` → looks up `/my-server` (previously `/registry/my-server`, 404)
- `GET /.well-known/oauth-protected-resource/my-server/mcp` → unchanged

Deployments whose `registry_url` has no path are unaffected.

## Why

RFC 9728 §3.1 forms the well-known URL by inserting `/.well-known/oauth-protected-resource` between host and path of the resource. For the advertised per-server resource `https://gw.example.com/registry/my-server/mcp` that is `https://gw.example.com/.well-known/oauth-protected-resource/registry/my-server/mcp`. Clients that derive the URL this way (as a first attempt, or as a fallback after `resource_metadata` from the 401 header) get a 404 and end up in legacy discovery. Closes #1746.

## Testing

- New unit tests in `tests/unit/api/test_wellknown_routes.py`: `test_path_normalization_strips_registry_url_prefix` (path-prefixed `registry_url` → handler looks up `/obo-echo`, 200, resource unchanged) and `test_path_normalization_without_prefix_is_unchanged`.
- `uv run pytest tests/unit/api/test_wellknown_routes.py`: 37 passed locally (the coverage gate fails on a single-file run, as expected). Full `./tests/run_all_tests.sh` not run locally; relying on CI.

## Notes

The bundled nginx already forwards `location ^~ /.well-known/` to the registry app. Operators with an external ingress that only routes `/<ROOT_PATH>` still need to route `/.well-known/` too; a one-line note is added to `docs/oauth-discovery-endpoints.md`.

### DIFF
--- a/docs/oauth-discovery-endpoints.md
+++ b/docs/oauth-discovery-endpoints.md
@@ -156,6 +156,21 @@ the client expects (e.g. `https://claude.ai/api/mcp/auth_callback` for
 Claude.ai's Custom Connector UI) and paste the resulting `client_id`/
 `client_secret` into the connector's "Advanced settings" panel.
 
+## Path-prefixed deployments
+
+When the gateway is served under a path prefix (`ROOT_PATH=/registry`,
+`REGISTRY_URL=https://gw.example.com/registry`), the per-server resource is
+`https://gw.example.com/registry/<server>/mcp`. RFC 9728 section 3.1 forms the
+path-aware well-known URL by inserting the well-known segment between host and
+path, so a client that derives it from the resource requests
+`https://gw.example.com/.well-known/oauth-protected-resource/registry/<server>/mcp`.
+The registry resolves that form (the gateway's own prefix is stripped before the
+server lookup), and the bundled nginx already forwards `location ^~ /.well-known/`
+to the app. If an external ingress or load balancer sits in front and only routes
+`/<ROOT_PATH>/*`, route `/.well-known/*` to the registry as well, or clients that
+do not follow the `resource_metadata` hint from the 401 will get a 404 and fall
+back to legacy discovery.
+
 ## Common diagnostic checks
 
 | Symptom | Likely cause |
--- a/registry/api/wellknown_routes.py
+++ b/registry/api/wellknown_routes.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from urllib.parse import urlparse
 
 from fastapi import APIRouter, HTTPException, Request
 from fastapi.responses import JSONResponse
@@ -310,6 +311,20 @@ async def get_oauth_protected_resource() -> JSONResponse:
     return JSONResponse(content=document, headers=OAUTH_DISCOVERY_CACHE_HEADERS)
 
 
+def _registry_url_path_prefix() -> str:
+    """Path component of ``settings.registry_url`` (e.g. ``/registry``), or ``""``.
+
+    Path-prefixed deployments (``ROOT_PATH`` / ``REGISTRY_URL`` with a path)
+    advertise per-server resources of the form ``https://gw/<prefix>/<server>/mcp``.
+    """
+    try:
+        path = urlparse(settings.registry_url).path or ""
+    except (TypeError, ValueError, AttributeError):
+        return ""
+    path = path.rstrip("/")
+    return path if path and path != "/" else ""
+
+
 def _normalize_prm_server_path(server_path: str) -> str:
     """Reduce a path-aware PRM suffix to the registered server path.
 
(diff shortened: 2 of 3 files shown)


docs/oauth-discovery-endpoints.md
```diff
@@ -156,6 +156,21 @@ the client expects (e.g. `https://claude.ai/api/mcp/auth_callback` for
...
+When the gateway is served under a path prefix (`ROOT_PATH=/registry`,
+`REGISTRY_URL=https://gw.example.com/registry`), the per-server resource is
+`https://gw.example.com/registry/<server>/mcp`. RFC 9728 section 3.1 forms the
+path-aware well-known URL by inserting the well-known segment between host and
+path, so a client that derives it from the resource requests
+`https://gw.example.com/.well-known/oauth-protected-resource/registry/<server>/mcp`.
+The registry resolves that form (the gateway's own prefix is stripped before the
+server lookup), and the bundled nginx already forwards `location ^~ /.well-known/`
+to the app. If an external ingress or load balancer sits in front and only routes
+`/<ROOT_PATH>/*`, route `/.well-known/*` to the registry as well, or clients that
```
omrishiv, line 170: Thank you for calling this out. Rather than document this, can we update https://github.com/agentic-community/mcp-gateway-registry/blob/main/charts/registry/templates/ingress.yaml#L42-L81 to include this path?

omrishiv (changes requested):
Hi @peelandsee, thank you for adding this. One small request if we can add.

Issue #1574, opened 2026-07-31
Scope seed files have transposed group_mappings: unrestricted scopes are unreachable, so admins resolve to 0 scopes and cannot register servers


billtarr-aws:
## Summary

The DocumentDB scope seed files in `scripts/` carry `group_mappings` arrays that were never inverted when scopes moved out of `auth_server/scopes.yml` (removed in `1db1f082`, "Remove file storage backend, FAISS search, and scopes.yml").

The old YAML was keyed **group → [scopes]**. The seed documents are keyed **scope → [groups]**. The migration copied each group's scope list verbatim into that scope's group list, so the relation is transposed.

| scope | groups per old `scopes.yml` | groups in seed file today |
|---|---|---|
| `mcp-servers-unrestricted/read` | `registry-admins`, `mcp-registry-admin`, `4c46ec66…` | **`[]`** |
| `mcp-servers-unrestricted/execute` | `registry-admins`, `mcp-registry-admin`, `4c46ec66…` | **`[]`** |
| `mcp-registry-admin` | `mcp-registry-admin` | `mcp-registry-admin`, **`…unrestricted/read`**, **`…unrestricted/execute`** |
| `registry-users-lob1` | `registry-users-lob1` | *(no seed file)* |
| `registry-users-lob2` | `registry-users-lob2` | *(no seed file)* |
| `public-mcp-users` | `public-mcp-users`, `5f605d68…` | *(no seed file)* |

## Why this breaks a fresh install

`map_cognito_groups_to_scopes()` ([dependencies.py](https://github.com/agentic-community/mcp-gateway-registry/blob/main/registry/auth/dependencies.py)) inverts every scope's `group_mappings` into a group → scopes lookup. Two consequences follow:

1. **The unrestricted scopes have no groups at all**, so they are unreachable. Every administrator resolves to zero server access.
2. **`mcp-registry-admin` lists two *scope ids* where group names belong.** A Keycloak group cannot contain `/`, so those entries are inert — they match no token claim and grant nothing.

Net effect: an operator in `mcp-registry-admin` or `registry-admins` authenticates successfully, resolves **0 scopes**, and is denied server registration. The documented quickstart cannot complete:

```
User service-account-agent-test-agent-m2m with 1 groups mapped to 0 scopes
resolve_scope_access: scopes=0 servers=0 tool_restricted_servers=0
User '...' attempted API register without register_service permission
POST /api/servers/register HTTP/1.1 403 Forbidden
[...]

PR #1757, opened 2026-09-15
fix(mongodb): run the init script from the image

omrishiv:
The `setup-mongodb` Job installed a database schema once and then never again. Kubernetes Jobs are immutable, so `helm upgrade` on an existing install either patched nothing — leaving the Completed Job alone, so an index added by a release was never created — or failed outright with "field is immutable" once the pod template changed. Either way the database kept whatever schema it was first installed with.

Making the Job a `post-install,post-upgrade` hook fixes that, but on its own it would have activated two destructive paths that then ran on every upgrade, forever. Both trace to the same root cause: the chart carried its own 589-line copy of `scripts/init-mongodb-ce.py`, because a chart cannot `.Files.Get` outside its own directory. The copy had drifted from source in six functional ways. Two mattered:

- Its scope seeds had transposed `group_mappings` — the regression fixed for #1574, where the unrestricted scopes get no groups and every administrator resolves to zero server access. `_load_default_scopes` upserts with `$set`, so each upgrade would re-apply it over the database.
- The Job's only environment was the MongoDB credential secret, which carries no TTL key, and the chart had no knob for one. `AUDIT_LOG_MONGODB_TTL_DAYS` therefore always resolved to the script's 7-day default, and the `IndexOptionsConflict` handler dropped and recreated the TTL at 7 days. An operator who had raised retention for compliance would lose everything older than 7 days, within about a minute, on an upgrade that said nothing.

So rather than resync the copy, delete it. The Job now runs the registry image's own `/app/scripts/init-mongodb-ce.py`. The image already ships the script, the scope seeds it loads from its own directory, and a uv.lock-pinned venv with motor/pymongo — docker-compose already invoked it exactly this way. A copy cannot drift if there is no copy, and this also removes the `pip install` of unpinned packages that ran on every upgrade with the database credential in scope, the `wait.py` init container (which looped with no deadline and always built a credentialed URI, so it could never connect to a no-auth MongoDB CE), and the second, competing mechanism for injecting the Entra admin group.

Retention becomes a Helm value, and the script now refuses to SHORTEN a TTL unless `AUDIT_LOG_MONGODB_TTL_ALLOW_SHRINK` is set, leaving the longer window in place and saying so. Lengthening still applies immediately.

Other correctness fixes in the same files:

- The unique-index migration now creates the composite `(request_id, log_type)` index BEFORE dropping the single-field one, so uniqueness is never briefly unenforced on a live cluster. On a pre-existing true duplicate it fails loudly with the aggregation that lists the offending pairs and leaves the old index in place, rather than half-migrating and wedging every later upgrade. That single-field index is why the second audit record of a request was rejected and logged as CRITICAL "AUDIT RECORD DROPPED" on Helm-deployed MongoDB CE.
- The blind `time.sleep(10)` and the deleted `wait.py` are replaced by a bounded wait in the script: wait for `ping`, then for replica-set readiness, and only initialize the set here after a grace period, so it cannot race the MongoDB Kubernetes operator into a single-member set under the wrong name.
- Identity-claim indexes are pinned per stream with a `partialFilterExpression` and lead with `log_type`, which every audit query filters on. Measured on a representative 30k-record mix: 2.09 MB across 13 pinned indexes versus 3.46 MB across 9 unpinned, and an insert updates 4-5 of them instead of all 9. `sparse` is not used — on a compound index it keeps a document when any indexed field exists, and `log_type` always exists, so it measured byte-identical to passing no option. The filter is a single-stream equality rather than an `$in` because Amazon DocumentDB only uses a partial index when the query predicate matches its filter expression.
- Index names are explicit and identical on both backends. They were previously auto-generated, producing different names per engine for the same logical index — which is why an earlier migration had to guess at name variants to drop one.
- Hook weight is -10, not 5. openbao-init is also 5, so Helm broke the tie alphabetically and the OpenBao bootstrap gated this database migration.
- The Job gets `activeDeadlineSeconds`, `ttlSecondsAfterFinished` and resource requests: it now gates the release, so an unbounded one hung `helm upgrade --wait` until the release timeout with the cause in a pod that the next attempt deleted. Helm also does not delete hook resources on uninstall.

`tests/unit/test_chart_mongodb_init_wiring.py` asserts what a byte-parity test on a copy could not: that no chart embeds an init script, that the path the Job executes exists, that the image ships it, that every scope file the script loads is present, and that both backends name each claim index identically. The helm unittest suite covers the hook, the absence of any mounted script, the retention passthrough and the bounds; each case was falsified by reverting the behaviour it covers.

Verified: 8113 passed; helm lint/template/unittest clean; and the migration run against a real MongoDB from the exact pre-upgrade index state — 13 pinned indexes created, 9 legacy dropped, the old unique index retired, both records for one request storable, true duplicates still rejected, idempotent on a second run, and a 90-day retention preserved against a defaulted 7.

*Issue #, if available:*

*Description of changes:*


aarora79:
Rebased onto `main` (was 18 commits behind, no conflicts) and added one commit reconciling this with #1774, which merged after this PR was opened.

**What #1774 changed:** the audit API no longer searches identity claims on the `registry_api_access` stream. The registry receives a thin signed assertion from the auth server rather than raw IdP claims, so those fields are permanent nulls there and matching them could only cost a predicate per document for a branch that cannot hit.

**Consequence here:** four of the thirteen claim indexes served no query, on the largest stream in the collection and the one the audit UI opens by default. The new commit drops those targets (13 to 9), adds the four superseded names to `LEGACY_AUDIT_CLAIM_INDEXES` so a cluster that already built them retires them on the next run, and renames `AUDIT_NESTED_LOG_TYPES` to `AUDIT_CLAIM_NESTED_LOG_TYPES` since the set is no longer "streams that nest the claims" but "streams whose nested claims are worth indexing".

**One correction to the docstring.** It said each claim needs an index "or the whole `$or` degrades to a collection scan". Measured on MongoDB: adding the claim indexes leaves both the plan and the documents examined unchanged, because the identity filter is one `$or` mixing claim equality with case-insensitive regex on the readable fields, and an `$or` is served by index union only when every branch is indexable, which a case-insensitive regex never is. Removing the regex branches took documents examined from 5200 to 1. What `log_type` leading every key does buy is a bound on the stream, which is real. A seek straight to a claim value additionally needs the equality branches split out of that `$or`, which is a change to the query rather than the schema, and is not in scope here.

`test_registry_api_stream_gets_no_claim_index` pins the exclusion on both backends and fails if the stream is added back or the superseded names are not retired.

**Verification after the rebase:** 8392 passed / 57 skipped / 0 failed; `helm unittest` 220 passed across all 5 charts; `helm lint` and `helm template` clean; ruff clean; bandit 0 issues at every severity on both init scripts.

I also reviewed the rest of this PR and found nothing breaking and no security issue. The destructive paths are all guarded correctly: the TTL guard refuses to shorten retention and keeps the longer window, the unique-index migration builds before dropping and fails loud on a real duplicate, replica-set initiate only fires when genuinely uninitialized and after a grace period, and index drops tolerate absence. Credentials arrive via `envFrom: secretRef` rather than argv, the pod runs non-root with capabilities dropped, and `_redact_uri` strips credential-bearing URIs from error messages (tested against multi-URI messages). Removing the runtime `pip install` that ran with the DB credential in scope is a real improvement.

Issue #1642, opened 2026-08-17
Audit Log displays Entra `sub` instead of identifiable user for OBO-authenticated MCP calls


zoltan-fedor:
## Summary

When an Entra-authenticated client accesses an OBO-capable MCP server through AI Gateway, the Gateway Audit Log displays the JWT `sub` claim in the **USER** field rather than a human-readable or stable Entra user identity.

For example, an audit entry for a request made by `user@example.com` is displayed as:

```text
USER: <subject-id>
```

I confirmed that this value is exactly the `sub` claim from the Entra access token.

The same token contains both the user's UPN and Entra Object ID, so Gateway has sufficient information to record an identifiable user.

## Environment

* AI Gateway: QA
* Gateway version: `1.29.0`
* Authentication: Microsoft Entra ID
* Flow: delegated user authentication / OBO
* Target: MCP server supporting OBO
* MCP operation observed: `tools/call`

## Steps to reproduce

1. Configure an MCP server behind AI Gateway with Entra OBO authentication.
2. Authenticate to Gateway using an Entra delegated user token.
3. Invoke an MCP operation, e.g. `tools/call`.
4. Open **AI Gateway → Audit**.
5. Locate the corresponding request.
6. Inspect the value shown in the **USER** column.
7. Decode the Entra access token used for the request and compare its identity claims.

## Actual behavior

Gateway displays an opaque value in the Audit Log **USER** field:

```text
<subject-id>
```

Decoded token claims confirm that this is the JWT `sub`:

```text
sub  = <subject-id>
oid  = <user-object-id>
upn  = user@example.com
name = Example User
scp  = access_as_user
```

Therefore, the value shown as **USER** is currently sourced from `sub`.

## Expected behavior

The Audit Log should identify the originating Entra user in a way that is meaningful to an administrator.

For example:

```text
USER: user@example.com
```

For durable identity and audit correlation, Gateway should also retain the Entra identity using:

```text
oid + tid
```

The `sub` claim may still be useful for protocol-level correlation and could remain available in the event details, but it should not be the only identity exposed as the human-facing **USER** value.

Suggested mapping:

```text
Audit USER        -> upn / preferred_username
Canonical user ID -> oid + tid
Event details     -> sub, oid, tid, appid/azp, upn, name
```

## Why this is a problem

`sub` is an opaque, pairwise subject identifier. An administrator reviewing the Audit Log cannot reasonably determine which Entra user performed an action from this value alone.

[...]

PR #1758, opened 2026-09-15
fix(audit): close two exposures on the audit path, and make degradation visible

omrishiv:
Merge after #1642 

Follow-ups from a review of the Entra OBO identity work (#1642). None of these are
required to attribute a caller, which is why they are separate — but two of them
close exposures that the identity work widens, so this should land with it or
immediately after, not later.

Two data-exposure fixes:

- `emit_audit_event` serialized the WHOLE audit record and logged it on
  `registry.audit`, which propagates to root handlers: stdout, a rotating file,
  and the `application_logs` MongoDB collection. Application logs have a different
  retention (a day by default) and a different access boundary than the audit
  collection the compliance documentation reasons about. The `username` leak
  predates this; the durable IdP identifiers (UPN, `oid`, `tid`, `appid`) are new,
  and shipping them into three extra sinks contradicts the data-minimisation
  rationale used to justify dropping the `name` claim. The record stream now goes
  to a dedicated `registry.audit.records` logger with `propagate = False` and its
  own handler, and the operator-facing payload masks the durable identifiers.
  `app_log_excluded_loggers` was NOT usable as the fix: it is only consulted by the
  MongoDB handler, so the console and file copies would still have received the
  values, and an operator can unset it. `registry.audit` itself could not be made
  terminal either — it is the parent of `registry.audit.service`, which emits the
  CRITICAL "AUDIT RECORD DROPPED" line. Full-fidelity claims remain in the durable
  audit store, which is where the audit UI and the identity feature read them.
- `_forward_headers` filtered the hop-by-hop and proxy-context header sets, but
  `x-internal-token` was in neither: it was stripped only on the vault/OBO
  branches. So for a plain registered MCP server, the gateway's own SECRET_KEY-
  signed hop JWT reached the registrant's backend — a credential replayable
  against `/mcp-proxy`, and now also a carrier of the caller's identity. It is
  stripped unconditionally, including when `relay_authorization` is set, and the
  `_EGRESS_STRIP_HEADERS` comment that claimed otherwise is corrected. The
  non-egress path had no coverage at all; it does now.

Correctness and robustness:

- `_claim_str` guarded only 2 of the 7 `_emit_token_mint_audit` call sites; a
  duplicated resolver fed the other five. `email` arrives from
  `registry/auth/dependencies.py` as `claims.get("email") or ""`, so a multivalued
  Keycloak mapper put a list into the record, `ValidationError` fired inside the
  best-effort emit, and the whole audit record was silently DROPPED. Both identity
  values are now coerced once, inside `_emit_token_mint_audit`, so every site is
  covered by construction, and the duplicated resolver is gone.
- The OBO mint record used a fresh `request_id` and a null `correlation_id`, so it
  could never be joined to the `mcp_server_access` record for the same request —
  even though the unique `(request_id, log_type)` index exists precisely so records
  of different `log_type` can share one id. `/validate` now signs its `request_id`
  into the hop claim as `rid`, and the mint record records it as `correlation_id`.
  A token minted before this change still audits.
- `_audit_provider` fell back to the configured IdP even for a `self_signed`
  token — one the gateway minted itself, which no IdP ever saw. That fallback is
  skipped now; such a token still reports the provider its originating session
  recorded.
- `_csv_safe` missed formulas behind leading whitespace or a newline. Spreadsheets
  trim a cell before deciding whether it is a formula, so `" =HYPERLINK(...)"` was
  live content in an export. The `username` query parameter was also unbounded
  while being interpolated into a regex; it is now length-limited.
- An `if audit_identity:` branch documented as "omitted when empty" could never be
  false, because the claim builder always seeds a display value. Its test passed a
  call shape no caller produces, so it would have kept passing if the branch were
  deleted. Both are gone.

Observability:

- The degradation paths were silent — an absent hop claim falling back to the
  opaque `sub`, and a claim coerced away or truncated. They now increment
  `audit_integrity_degraded_total`, labelled by a bounded `reason`, so they are
  alertable as a rate instead of invisible; the only prior signal was a
  `logger.critical` line, which cannot be alerted on as a rate. Deliberately no
  unbounded label dimensions: no tenant, app, principal or request id as an
  attribute. An ABSENT claim counts nothing — normal for any IdP that omits it —
  which is what keeps the series meaningful.
- The Status Distribution card is hidden for the `token_mint` stream, which
  structurally has no `response.status_code` to bucket, rather than rendering a
  titled card containing "No data available".

Verified: 8221 passed. Both exposure fixes re-proved independently of their own
tests — a record's six claim values reach the durable store intact and no root
handler (re-enabling propagation leaks them again, so the guard is load-bearing),
and no `x-internal-token` variant survives `_forward_headers` on the non-egress
path.

Refs #1642


aarora79:
Follow-up landed in #1775 (merged as `d22270f1`), closing the one should-fix from the review of this PR.

**The readable identity is now reduced on the operator stream too.** `username` and `principal_name` resolve to the same UPN on an Entra v1.0 access token, so redacting the claim while emitting `username` verbatim left the address in the stream unchanged. An address now emits as `a***@contoso.com`; the full value stays in the durable audit store.

Two carve-outs, the second of which your own test docstring argued for: values with no `@` (opaque subjects, service identities, the `anonymous` literal) are left alone, and `ToolFilterAuditEvent` is never reduced because this stream is its only sink, so masking it would destroy the only record of who was affected. That distinction is explicit in `_NO_DURABLE_SINK_TYPES` with a regression test.

Also from the review: `docs/logging.md` now states what the operator stream carries and that audit records no longer surface in the admin log API, and the counter comment names `mcpgw_registry_token_mint_total` rather than saying "above" for a metric declared in the auth server's meters module.

Nothing else from this PR needed changing. Both exposure fixes verified load-bearing (flipping `propagate = False` fails three tests; the header strip has a dedicated test for the plain-MCP path that previously had none), and the `_claim_str` chokepoint move is the fix I'd have called out as the most valuable non-security item here.

PR #1764, opened 2026-09-16
feat(keycloak): make the admin token realm configurable

mrchi:
## Problem

`registry/utils/keycloak_manager.py` authenticates to the Keycloak Admin REST API with the password grant against a token endpoint that is hardcoded to the `master` realm:

```python
token_url = f"{KEYCLOAK_ADMIN_URL}/realms/master/protocol/openid-connect/token"
```

That realm is where the *admin user* lives, which is not the realm being administered (`KEYCLOAK_REALM`). The two coincide in the default deployment, where a `master` user holds `view-users` / `query-groups` on the `<realm>-realm` client.

An operator who would rather not distribute a `master` credential grants the same `realm-management` roles to a user inside the managed realm instead. That deployment cannot authenticate: Keycloak isolates users per realm, so the master realm's token endpoint answers `invalid_grant` for a user that lives elsewhere. What the operator sees is a generic admin-API authentication failure with nothing in it to suggest the realm is the cause.

## Change

Add `KEYCLOAK_ADMIN_REALM` (default `master`) and interpolate it into the token URL:

```python
KEYCLOAK_ADMIN_REALM: str = os.environ.get("KEYCLOAK_ADMIN_REALM") or "master"
...
token_url = f"{KEYCLOAK_ADMIN_URL}/realms/{KEYCLOAK_ADMIN_REALM}/protocol/openid-connect/token"
```

`or "master"` rather than a `get` default, so a bare `KEYCLOAK_ADMIN_REALM=` in `.env` still falls back instead of producing a `//realms//` URL.

`KEYCLOAK_REALM` is deliberately not reused. It names the realm being administered, so collapsing the two would immediately break every deployment whose admin lives in `master`. `admin-cli` stays hardcoded — every realm ships it.

This is the only runtime site. The other token endpoints that mention `realms/` (`auth_server/providers/keycloak.py`, `registry/api/*`) build from `KEYCLOAK_REALM` for the managed realm and are untouched. `api/get-m2m-token.sh` also hardcodes `/realms/master`, but it is an AWS-only operational script that depends on SSM plus AWS credentials and is not on the request path.

## Compatibility

The default stays `master`, so existing deployments are unaffected. Verified with `docker compose config`: the registry service renders `KEYCLOAK_ADMIN_REALM: master` when the variable is unset, and honours `KEYCLOAK_ADMIN_REALM=<realm>` when it is set, with `KEYCLOAK_REALM` unchanged in both cases.

## Deployment surfaces

Wired: `docker-compose.yml`, `docker-compose.podman.yml`, `docker-compose.prebuilt.yml` — the three files that list the registry service's `KEYCLOAK_ADMIN`. `docker-compose.dhi.yml` has no registry service env block, so it does not apply.

Not wired, intentionally: `KEYCLOAK_ADMIN_REALM` is not a reserved name in `charts/registry/reserved-env-names.txt`, nor in Terraform's `registry_extra_env` validation. Helm operators can therefore already supply it through the `charts/registry` `extraEnv` value, and ECS operators through `registry_extra_env`, which `ecs-services.tf` concatenates into the registry container's `environment`. The `docs/unified-parameter-reference.md` row records this in the Terraform and Helm cells rather than leaving them blank. There is no CDK equivalent.

## Tests

`tests/unit/test_keycloak_admin_realm.py` covers:

- default `master` when the variable is unset, and when it is set to the empty string
- a configured realm reaching the token URL
- `KEYCLOAK_REALM` not leaking into the admin token realm
- each compose file receiving the variable, interpolating from `.env`, defaulting to `master`, and keeping `KEYCLOAK_REALM` separate

The compose assertions follow the invariant style of `tests/unit/test_compose_keycloak_external_url.py`.

## Docs

`.env.example`, `docs/configuration.md`, `docs/unified-parameter-reference.md`.

## Checks

- `uv run pytest tests/unit` — 6969 passed, 31 skipped
- `uv run pytest tests/ --ignore=tests/e2e --ignore=tests/stress -n 8` — 8227 passed, 99 skipped. The 11 failures in `tests/integration/test_custom_entities_integration.py` reproduce on the unmodified base commit and are unrelated to this change.
- `ruff check` and `ruff format --check` clean on the touched files
- Type check reports only the 3 diagnostics already present on the base commit in `keycloak_manager.py`


omrishiv (dismissed):
@mrchi thank you for the PR. Can you please extend this change into the charts and terraform directories?

omrishiv (changes requested):
@mrchi, sorry for the back and forth on the review. I see you've called out not putting this as discrete envvars in the charts/terraform, but there are still scripts that have `master` realm hardcoded in them. We probably need to address that properly, otherwise the code will be inconsistent with what is deployed if it's set in helm

Issue #1752, opened 2026-09-14
`search_registry` returns `relevance_score: 1.0` for many/all keyword-matched tools instead of a graded score


zoltan-fedor:
## Summary
When calling `search_registry`, we frequently see large numbers of tools in the `matching_tools` / `tools` results all reporting `relevance_score: 1`, even though their actual match quality clearly varies (e.g. a query for "LeanIX fact sheets" returns a dozen+ LeanIX tools that are all scored `1`). This makes the score field useless for ranking/filtering tools within a server.

## Root cause

In `DocumentDBSearchRepository.search()` (`registry/repositories/documentdb/search_repository.py`), the keyword "text boost" aggregation stage builds `matching_tools` via a MongoDB `$filter` + `$map`:

    "matching_tools": {
        "$map": {
            "input": {
                "$filter": {
                    "input": {"$ifNull": ["$tools", []]},
                    "as": "tool",
                    "cond": {
                        "$or": [
                            {"$regexMatch": {"input": ..."$$tool.name"..., "regex": token_regex, ...}},
                            {"$regexMatch": {"input": ..."$$tool.description"..., "regex": token_regex, ...}},
                        ]
                    },
                }
            },
            "as": "tool",
            "in": {
                "tool_name": "$$tool.name",
                "description": ...,
                "relevance_score": 1.0,
                ...
            },
        }
    }

Any tool whose name or description matches the query's tokenized regex is kept by the `$filter`, and then **every** matched tool is unconditionally stamped with `relevance_score: 1.0`. This is a boolean "did this tool's text match" flag, not a graded relevance/similarity score — there's no distinction between a tool that matches one weak token versus one that matches every query token exactly.

This is inconsistent with the equivalent Python-side scoring function `_score_tool_relevance()` (same file), used in the lexical/legacy fallback search paths, which computes a graded score in `[0.0, 1.0]` based on match strength and token coverage:

    def _score_tool_relevance(tool_name, tool_description, query_tokens) -> float:
        ...
        score += 0.5  # token in name
        score += 0.3  # token in description
        ...
        score = min(1.0, score) * 0.7 + token_coverage * 0.3
        return round(min(1.0, score), 4)

## Secondary/compounding factor

`_normalize_scores()` (same file) performs min-max normalization that always maps the *top-ranked* result of a given search call to exactly `1.0`:

    max_score = scored_results[0][1]
    min_score = scored_results[-1][1]
    if max_score == min_score:
[...]

PR #1765, opened 2026-09-16
fix(mcpgw): return the custom records and asset metadata the registry already sends

go-faustino:
Two independent places where the MCP tools discard data the registry had already returned. Both are in `servers/mcpgw/`; neither needs a registry or schema change.

---

# 1. `search_registry` could not see custom entity records

## Problem

The registry indexes custom entity records for semantic search, gates them per type, and returns them in the `custom` array of `POST /api/search/semantic`. `search_registry` never saw them, for two reasons at once:

1. It pinned `entity_types` to the five built-in names, so the registry filtered custom records out of the candidate pool before scoring them.
2. It then read only `servers`, `tools`, `agents`, `skills` and `virtual_servers` off the response, so even a registry that returned `custom` would have had it dropped.

The effect is that custom entity records are undiscoverable through MCP. The catalog can hold them and the UI can find them, but no agent can, even though discovery is the surface every MCP client uses.

## Fix

Send no `entity_types` filter. A longer hard-coded list cannot solve this: custom type names are defined at runtime by an administrator, so the tool cannot know them when it builds the request. Omitting the filter is what `useSemanticSearch` in the frontend already does for exactly this reason, and it leaves the registry searching `_default_search_scope()`, which is the built-in types plus every custom type in the descriptor cache.

Then read `custom` off the response, count it in `total_results`, and return it.

### Built-in results are unchanged

Dropping the filter does not narrow or widen what built-ins come back:

- The hybrid path strips `"tool"` from the requested types (`search_types = [t for t in (entity_types or default_scope) if t != "tool"]`) and lifts tools out of their parent server documents instead, so passing `"tool"` was already a no-op there.
- The four types that remain (`mcp_server`, `a2a_agent`, `skill`, `virtual_server`) are exactly `_default_search_scope()`, so virtual servers stay in scope — the guarantee added in #1752 holds.
- The lexical-only fallback now runs unfiltered, so a standalone `mcp_tool` document can match there too.

### Discovery receipt

Custom records consume the `max_results` budget the same way virtual servers did before #1752, so they now join the receipt's candidate set. `asset_type` carries the custom type's own name rather than a flat `"custom"`, so an eval can tell which kind of record matched. `docs/dynamic-tool-discovery.md` now notes that `asset_type` is an open set rather than a fixed enum.

### Compatibility

A registry with `custom_entity_types_enabled` off, or one too old to know the key, sends no `custom` array and yields `[]`. Per-type visibility and scope gating is unchanged: it happens server-side in `search_routes.py`, so a caller without the `list_<type>_entity` grant still sees nothing.

---

# 2. The tools dropped the metadata attached to an asset

## Problem

Whoever registers a server, agent or skill can attach arbitrary metadata to it — ownership, provenance, configuration — and the registry returns it on the read and list APIs. The MCP tools threw it away, so an agent could see that an asset exists but never learn anything about it beyond name, description and tags. "Who owns this server?" was unanswerable through MCP even when the registry held the answer.

Two separate causes:

**The list tools reshaped it away.** `list_services`, `list_agents` and `list_skills` funnel every record through a local Pydantic model in `servers/mcpgw/models.py` before returning it (`ServerInfo(**s).model_dump()`). None of those models declared `metadata`, so Pydantic dropped it silently — the registry sent it, the tool discarded it.

**`search_registry` never asked.** The search endpoint returns the metadata subdocument only when the request names the keys to project (`metadata_fields`, #1277). The tool did not pass it, so results always came back with none.

## Fix

The three list models now carry `metadata`. No new argument is needed there, because the list endpoints already return metadata in full unless a caller narrows it — `project_metadata(md, None)` returns `md` unchanged.

`search_registry` takes an optional `metadata_fields` and forwards it. I deliberately did not hard-code a key set: which metadata an asset carries is up to whoever registered it, so any fixed list would only fit one deployment's conventions. The docstring instead tells the model *when* to reach for it — when the question is about an asset rather than about finding one. An empty string is treated as unasked rather than forwarded, so it cannot turn into a 422 from the projection validator.

`docs/metadata-field-projection.md` documented only the HTTP surface; it now also covers the MCP tools and notes the asymmetry, which is the search endpoint's rather than the tools': listing returns metadata unless you narrow it, search returns none unless you ask.

---

## Tests

`tests/unit/servers/mcpgw/test_intelligent_tool_finder.py` — custom records are returned; they count toward `total_results`; they appear in the discovery receipt with the custom type name as `asset_type`; a record missing its `entity_type` discriminator falls back to `"custom"`; a missing `custom` key yields `[]`. Plus `metadata_fields` is absent unless asked, is forwarded when given, arrives on results, and an empty string is treated as unasked.

`test_search_registry_still_requests_virtual_servers` asserted on the `entity_types` payload this change removes. It is now `test_search_registry_does_not_pin_entity_types`, keeping the same guarantee (virtual servers stay in scope) against the new mechanism.

New `tests/unit/servers/mcpgw/test_list_tools_metadata.py` covers metadata passthrough on all three list tools, including skills (whose free-form keys nest under `metadata.extra`) and records that carry no metadata at all.

It follows the import pattern `test_get_skill_content.py` documents — stub `fastmcp` only when genuinely absent and do not pop the server module from `sys.modules`, since that makes `patch()` targets resolve to a different module object than the tools under test came from, and silently breaks sibling test modules.

`tests/unit/servers/mcpgw/` passes (66 tests), as do `tests/unit/api/test_search_routes.py` and `tests/unit/utils/test_metadata_projection.py` (141 tests). `ruff check` and `ruff format --check` are clean on the changed files.

## Notes

I did not find an existing issue or open PR for either half, so there is no `Refs` line. Happy to split this into two PRs or file issues first if you would prefer.


servers/mcpgw/server.py
```diff
@@ -839,20 +860,39 @@ async def search_registry(
...
+                    "asset_type": record.get("entity_type") or "custom",
+                    "service_path": record.get("path") or "",
+                    "name": record.get("name") or "",
+                    "similarity_score": record.get("relevance_score") or record.get("score"),
+                }
+            )
         # Dedupe so a tool returned in both tools[] and a server's matching_tools
         # is counted once. Then split into what the caller saw vs what the limit
         # held back.
         candidate_results = _dedupe_candidates(candidate_results)
```
omrishiv:
```suggestion
        candidate_results = _dedupe_candidates(candidate_results)
        candidate_results.sort(key=lambda c: c.get("similarity_score") or 0, reverse=True)
```
We need to sort here otherwise the results will be returned in append order, not necessarily the reverse sorted order

PR #1766, opened 2026-09-16
fix(security): a scan that could not run is not an unsafe verdict

go-faustino:
## The problem

`SecurityScanResult` already carries two distinct booleans — `is_safe` (the verdict) and `scan_failed` (whether the scan completed at all). When a scan raises, the scanner sets **both**:

```python
# registry/services/security_scanner.py
is_safe=False,
critical_issues=0,
high_severity=0,
...
scan_failed=True,
```

But the three registration handlers branch on `is_safe` alone:

- `server_routes.py` — `if not scan_result.is_safe:` → `toggle_service(path, False)`
- `agent_routes.py` — `if not scan_result.is_safe:` → `toggle_agent(path, False)`
- `skill_routes.py` — `if not result.is_safe and config.block_unsafe_skills:` → `toggle_skill(enabled=False)`

So *"I couldn't open a session with this endpoint"* disables an asset exactly as *"I found a critical prompt injection"* does.

Fail-closed is the right instinct in general. What makes this different is that it fails closed on an **inconclusive** result, for an asset that is already registered and was serving traffic a moment ago, on the strength of **zero findings**.

And some assets can never produce a conclusive scan. An MCP server behind per-user egress OAuth has no credential to offer the scanner at registration time, so its scan raises *every* time. Because registration re-runs the scan, that makes any edit to such a server's definition disable it.

What convinced me this is a bug rather than a deliberate policy: `scan_failed` is already surfaced throughout the read path — the `/security-scan` endpoints return it, and `scan_summary.py` feeds it to the UI's shield icon. The enforcement path is the only place the distinction gets dropped.

## How we hit it

On a 128-server deployment, adding a metadata field to every definition re-registered them all. The 112 servers using per-user egress OAuth failed their scans with zero findings, were disabled, and were dropped from the generated nginx config — which took them off the gateway entirely until they were re-enabled by hand. Requests fell through to the registry app and returned `405`.

```
Security scan failed endpoint=https://redacted.internal/mcp type=RuntimeError
Server /redacted failed security scan. Critical: 0, High: 0
Added 'security-pending' tag to /redacted
Disabled server /redacted due to failed security scan
```

The servers that survived were precisely the ones reachable anonymously. The mass re-registration only set the blast radius; the underlying defect disables a single such server on any edit to its definition.

## The change

Gate the disable on the scan having actually produced a verdict, in all three families, and keep the `security-pending` tag on **both** paths so an inconclusive scan stays visible:

```python
if scan_config.block_unsafe_servers and (
    not scan_result.scan_failed or scan_config.block_on_scan_failure
):
```

`SECURITY_BLOCK_ON_SCAN_FAILURE` (default `false`) lets operators who genuinely want strictly fail-closed registration restore the old behaviour. I defaulted it off because disabling an asset the scanner never assessed is almost never what an operator means by "block unsafe servers" — but it's a one-line default change if you'd rather ship it the other way.

One shared setting covers all three families rather than three parallel ones, since the semantics are identical and it's a strictness escape hatch rather than per-type policy. Happy to split it if you'd prefer symmetry with the existing `block_unsafe_*` trio.

**Also fixed:** the skill path applied the `security-pending` tag only *inside* its `block_unsafe_skills` branch, so with blocking off an unsafe skill was left both enabled and unlabelled — unlike the server and agent paths, which tag independently. The tag now lands regardless, which is what makes turning blocking off a safe choice at all.

No schema change is needed (`scan_failed` already exists on all three result models), and `auto_disabled` already flows into `fire_scan_complete_event`, so webhook consumers get the corrected signal for free.

## Testing

12 new unit tests across the three existing test files. For each family: a scan that raised leaves the asset enabled but tagged; a real finding still disables it (the regression guard that keeps this from becoming "we turned the security control off"); and `block_on_scan_failure=true` restores fail-closed. Plus one for the skill tagging bug.

I verified the tests actually catch the bug by reverting each fix and watching the relevant tests fail, rather than trusting that they passed.

```
12 new tests                           passed
test_server_routes.py                  138 passed, 5 skipped
test_agent_routes.py                   80 passed, 2 skipped
test_skill_routes_security.py          29 passed
test_skill_security_schemas.py         4 passed
test_server_routes_put.py              23 passed
test_server_routes_patch.py            20 passed
ruff check / ruff format               clean on changed files
```

One note in the interest of not overclaiming: `TestUnsafeSkillDisablingAndTagging::test_unsafe_skill_disabled_and_tagged` fails when the skill suite runs after the large server suite, via a hypothesis `DeadlineExceeded` (5233 ms on the first call, 8.7 ms on retry). I confirmed it fails identically on unmodified `main`, so it's pre-existing timing flakiness rather than anything from this change, and I've left it alone. It would want `deadline=None` if you want it fixed.

## Relationship to #1719

[#1719](https://github.com/agentic-community/mcp-gateway-registry/pull/1719) (per-tool blocking) touches some of the same files but is orthogonal: it refines *what to do when findings exist*, whereas this refines *what to do when there is no result at all*. They compose cleanly, though whichever lands second will want a trivial rebase around the `block_unsafe_servers` branch in `server_routes.py`.


omrishiv:
hi @go-faustino, thanks for this. I think this makes sense, but I'd ask @aarora79 as well.

I would suggest that we set the default `SECURITY_BLOCK_ON_SCAN_FAILURE` to `True` to keep the current behavior unless Amit confirms this is a bug and not intentional

go-faustino:
Thanks @omrishiv, that's a fair call and I've made the change in 6367f42: `SECURITY_BLOCK_ON_SCAN_FAILURE` now defaults to `true`, so out of the box the behaviour is exactly what it is today and nothing shifts for existing deployments unless an operator opts in.

Worth being precise about what that does and doesn't alter. The substance of the patch is unchanged: the three handlers now branch on `scan_failed` rather than on `is_safe` alone, so "no verdict" and "bad verdict" become distinguishable at all — the default only decides which way the switch points. The skill tagging fix also stays unconditional, so an unsafe skill is tagged `security-pending` whether or not blocking is enabled, matching the server and agent paths. That's what makes turning the flag off a visible choice rather than a silent one.

@aarora79 — the question for you, as small as I can make it: **when a scan raises, should the asset be disabled?** The scanner sets `is_safe=False` alongside `scan_failed=True` with all four severity counts at zero, and the registration handlers have been reading only `is_safe`.

Two things suggest an oversight rather than policy, though you'd know better than I would:

- `scan_failed` is already surfaced throughout the read path — the `/security-scan` endpoints return it, and `scan_summary.py` feeds it to the UI's shield icon. Enforcement is the only place the distinction gets dropped.
- Some assets can never produce a conclusive scan. An MCP server behind per-user egress OAuth has no credential to offer the scanner at registration time, so its scan raises every time; because registration re-runs the scan, any edit to such a server's definition disables it.

If you confirm it's a defect, flipping the default back is a one-line change and I'll push it straight away. If it's deliberate, the flag stays opt-in exactly as it is now, and the separation plus the tagging fix still seem worth having. Happy either way — your call.

In the interest of declaring my own bias: our deployment will set this to `false` explicitly, so the default costs us nothing either way. I'd just rather the out-of-box behaviour not surprise the next operator who runs into it.

go-faustino:
Rebased onto current `main` (`5e08a3d8`) — conflicts resolved. #1719 has landed in the meantime, and it composed as expected: the diff is unchanged at 417 insertions across the same 13 files.

The one textual conflict was in `tests/unit/api/test_server_routes.py`, where #1719's `TestToggleTool` and this PR's scan-failure tests were appended at the same place. Both kept, purely additive.

`registry/api/server_routes.py` merged without a conflict — #1719's work there is the separate `toggle-tool` route, so the `block_unsafe_servers` branch this PR rewrites was untouched by it. The "trivial rebase" I flagged in the description turned out to be free.

Verification: `test_server_routes.py`, `test_agent_routes.py`, `test_skill_routes_security.py` and #1719's `test_server_service.py` are `357 passed, 7 skipped`. A full `tests/unit/` run matches clean `main` exactly — same failures before and after the rebase. `ruff format --check` is clean on all 13 files; `ruff check` reports only the three pre-existing `UP042` findings in `config.py` that are byte-identical on `main`.

The one skill-suite failure is the pre-existing hypothesis flake noted in the description, unchanged: `DeadlineExceeded: Test took 5213.35ms, which exceeds the deadline of 200.00ms`, and the file passes `29 passed` in isolation.

---

One thing worth flagging from the merge, not changed here. #1719 added `ServerService.reconcile_security_blocks()`, which rebuilds `tool_overrides` from a scan's `raw_output` and deliberately does not carry forward any `source="security_scan"` entry that the current scan did not re-flag — that's what clears a stale auto-block after a clean rescan.

It has no production call site yet, so nothing is affected today. But when it does get wired into the scan path, it will meet the same question this PR is about, from the other direction: a scan that could not run produces no `tool_results`, so reconciliation would read it as "nothing is flagged any more" and drop every auto-block — fail-open, where the handlers here were failing closed. Whatever you decide about `SECURITY_BLOCK_ON_SCAN_FAILURE`, that call site will want the same `scan_failed` check before it is allowed to clear anything. Happy to open a separate issue for it if that's useful.

@aarora79 — the original question is still the only thing blocking this one: when a scan raises, should the asset be disabled? The default is currently `true`, so merging changes nothing out of the box either way.

aarora79:
Hey @go-faustino, thanks for this, and the diagnosis is right. `scan_failed` is already on every read path, so enforcement was the only place throwing that distinction away.


The new setting isn't wired up anywhere. It only exists in `registry/core/config.py`. The compose files list every env var one by one, with no `env_file` catch-all, so anything they don't list never reaches the container. Its two siblings are right there at `docker-compose.yml:377-378`. So an operator on Docker, ECS or EKS can't turn the flag off without editing Python, which means the people who hit this still can't fix it. We did the same thing in 1.31.0 with the `DEDUP_*` variables and ended up writing an upgrade warning into the release notes.

Ten files need it: `.env.example`, the three compose files, `terraform/aws-ecs/variables.tf` plus the module wiring and `terraform.tfvars.example`, then `charts/registry/values.yaml`, `templates/deployment.yaml` and `reserved-env-names.txt`, and finally `docs/unified-parameter-reference.md` and `CONFIG_GROUPS` in `registry/api/config_routes.py`. The reserved-names one is easy to miss. If you skip it, someone's `extraEnv` can shadow the setting, and only the Helm unittest suite will catch that.

On your two questions. Keep the default at true. Blocking on an inconclusive scan costs you availability, which is the cheaper mistake to make. Go the other way and an operator who never finds the setting keeps unassessed assets serving without knowing it. And keep it as one setting. The behaviour is identical for servers, agents and skills, and there isn't much symmetry to match anyway: only the server flag is in `.env.example`, the agent one uses an `AGENT_SECURITY_ prefix`, and the skill one isn't declared at all.

PR #1772, opened 2026-09-17
fix(compose): forward OPENBAO_ROLE_ID and OPENBAO_SECRET_ID to the registry

### LINKED ISSUE #1771: `OPENBAO_AUTH_METHOD=approle` is unusable in Docker Compose: `OPENBAO_ROLE_ID` / `OPENBAO_SECRET_ID` are never passed to the registry container
## Summary

`registry/secrets/factory.py` supports `OPENBAO_AUTH_METHOD=approle` and reads `OPENBAO_ROLE_ID` and `OPENBAO_SECRET_ID` from the process environment. None of the Docker Compose files forward those two variables to the `registry` service, so the approle branch can never succeed there. Every other `OPENBAO_*` variable is forwarded; these two are the only ones missing.

The two variables also do not appear in `.env.example`, in the variable table of `docs/egress-credential-vault.md`, or in `docs/unified-parameter-reference.md`. Outside of `factory.py` they are not mentioned anywhere in the repository, so an operator has no way to discover them short of reading the source or hitting the error.

The practical consequence is that the only working OpenBao auth method on the Compose path is `token`, i.e. handing the registry a static token with access to every vaulted egress credential. An AppRole bound to a scoped policy is the least-privilege option the code already supports; it is just not reachable from Compose.

## Environment

- Version: 1.30.0, tag `1.30.0`
- Docker Compose deployment using `docker-compose.prebuilt.yml` (prebuilt images)
- `SECRET_STORE_BACKEND=openbao`, `OPENBAO_AUTH_METHOD=approle`
- Also present on `main` at the time of writing

## Steps to reproduce

1. In `.env`, set `SECRET_STORE_BACKEND=openbao`, `OPENBAO_ADDR` to a reachable OpenBao, `OPENBAO_AUTH_METHOD=approle`, and valid `OPENBAO_ROLE_ID` / `OPENBAO_SECRET_ID` for an AppRole with a policy on the egress KV path.
2. `docker compose -f docker-compose.prebuilt.yml up -d`
3. Confirm the variables are absent from the container: `docker compose exec registry env | grep OPENBAO_` lists `OPENBAO_ADDR`, `OPENBAO_NAMESPACE`, `OPENBAO_KV_MOUNT`, `OPENBAO_AUTH_METHOD`, `OPENBAO_ROLE`, `OPENBAO_TOKEN` and nothing else.
4. Log in and open any egress-credential view, or call any authenticated egress-vault endpoint.

## Expected

The registry authenticates to OpenBao with the AppRole and the request succeeds.

## Actual

The registry starts and reports healthy. Every authenticated request that touches the egress vault returns `500 Internal Server Error`, and the registry log shows:

```
ValueError: OPENBAO_AUTH_METHOD=approle requires OPENBAO_ROLE_ID and OPENBAO_SECRET_ID.
```

The message is accurate, and the variables *are* set in `.env`. They just never reach the container.

## Where the gap is (tag `1.30.0`)

**The code reads the two variables from the environment** — `registry/secrets/factory.py`:

```python
# L122-131
elif method == "approle":
    import os

    role_id = os.environ.get("OPENBAO_ROLE_ID")
    secret_id = os.environ.get("OPENBAO_SECRET_ID")
    if not role_id or not secret_id:
        raise ValueError(
            "OPENBAO_AUTH_METHOD=approle requires OPENBAO_ROLE_ID and OPENBAO_SECRET_ID."
        )
    client.auth.approle.login(role_id=role_id, secret_id=secret_id)
```

**Compose forwards every `OPENBAO_*` variable except those two** — the `registry` service's `environment:` block in all three files:

| File | `OPENBAO_*` block | `OPENBAO_ROLE_ID` / `OPENBAO_SECRET_ID` |
|---|---|---|
| `docker-compose.prebuilt.yml` | L124–131 | absent |
| `docker-compose.podman.yml` | L113–120 | absent |
| `docker-compose.yml` | L233–240 | absent |

The first two are the ones we reproduced on; `docker-compose.yml` has the identical block and the identical gap.

**Documentation and examples do not mention them either:**

| File | What is there | What is missing |
|---|---|---|
| `.env.example` L2042–2058 | `OPENBAO_ADDR`, `_NAMESPACE`, `_KV_MOUNT`, `_AUTH_METHOD`, `_ROLE`, `_TOKEN` | `OPENBAO_ROLE_ID`, `OPENBAO_SECRET_ID` |
| `docs/egress-credential-vault.md` L374–378 | same six rows | same two |
| `docs/unified-parameter-reference.md` L792–797 | same six rows | same two |

## Two related documentation defects, same root

While confirming the above I found the existing docs describe `approle` incorrectly, which would send an operator down the wrong path even after the variables are wired:

1. `docs/egress-credential-vault.md` L378 says `OPENBAO_ROLE` is *"required for `kubernetes`/`approle` auth"*. The code disagrees: the `kubernetes` branch reads `settings.openbao_role` (L113–116), but the `approle` branch never reads `OPENBAO_ROLE` at all — it reads `OPENBAO_ROLE_ID` and `OPENBAO_SECRET_ID` only. Setting `OPENBAO_ROLE` for approle does nothing.
2. `docs/unified-parameter-reference.md` L795 lists `OPENBAO_AUTH_METHOD` as `` `token` | `kubernetes` `` and omits `approle` entirely, while L377 of the vault doc and the code both accept it.

## Suggested fix

Forward the two variables in the `registry` service of all three compose files, next to the existing `OPENBAO_ROLE` line, with empty defaults so the `token` and `kubernetes` paths are unaffected:

```yaml
      - OPENBAO_ROLE=${OPENBAO_ROLE:-}
      - OPENBAO_ROLE_ID=${OPENBAO_ROLE_ID:-}
      - OPENBAO_SECRET_ID=${OPENBAO_SECRET_ID:-}
```

And document them alongside the other `OPENBAO_*` entries:

- `.env.example` — two lines after `OPENBAO_ROLE=`, with a comment that they are used only by `OPENBAO_AUTH_METHOD=approle` and that `OPENBAO_SECRET_ID` is a secret.
- `docs/egress-credential-vault.md` variable table — two rows, and correct the `OPENBAO_ROLE` row to say `kubernetes` only.
- `docs/unified-parameter-reference.md` — two rows, and add `approle` to the `OPENBAO_AUTH_METHOD` row.

Two smaller things worth considering in the same change, though neither is required to fix the bug:

- **Consistency with `Settings`.** `OPENBAO_ROLE` is a `Settings` field (`registry/core/config.py` L1704) and the `kubernetes` branch reads it as `settings.openbao_role`, while the `approle` branch bypasses `Settings` and reads `os.environ` directly. Adding `openbao_role_id` / `openbao_secret_id` fields and reading them the same way would make the approle branch behave like the other two and show up in the config surface (`/api/config`) like the rest of the `OPENBAO_*` family. If you prefer to keep the secret out of `Settings`, that is a reasonable choice too — just noting the asymmetry.
- **`OPENBAO_TOKEN` stays mandatory even when approle is used.** All three compose files reference it as `${OPENBAO_TOKEN:?...}`, both for the registry and as `BAO_DEV_ROOT_TOKEN_ID` for the bundled dev-mode OpenBao container. So an operator on approle against an external OpenBao must still set a value for a variable the registry will never use. That is defensible for the bundled dev instance, but a note in the docs that it can be any non-empty placeholder in that scenario would save a confused minute.

## Workaround

Until the compose files carry the variables, drop them into `extra_env/registry.env` (the `env_file` hook every Compose variant already loads for the registry, `required: false`), or add them to the `registry` service in a `docker-compose.override.yml`. We are running the former in production and approle works as intended once the variables reach the container.

Evolver-sweden:
Fixes #1771

## What was wrong

`registry/secrets/factory.py` supports `OPENBAO_AUTH_METHOD=approle` and reads `OPENBAO_ROLE_ID` and `OPENBAO_SECRET_ID` from the process environment (L122–131). None of the compose files forwarded those two variables to the `registry` service, while every other `OPENBAO_*` variable was forwarded. The approle branch therefore could never succeed on the Compose path.

The failure is quiet in the wrong place: the registry starts and reports healthy, and every authenticated request that touches the egress vault returns 500 with

```
OPENBAO_AUTH_METHOD=approle requires OPENBAO_ROLE_ID and OPENBAO_SECRET_ID.
```

The message is accurate and the variables *are* set in `.env`. They just never reach the container. The practical consequence is that the only working OpenBao auth method in Compose was `token`, i.e. a static token with access to every vaulted egress credential, when a scoped AppRole is the least-privilege option the code already supports.

Outside `factory.py` the two names appeared nowhere in the repository: not in any compose file, not in `.env.example`, not in either parameter table. An operator could not discover them short of reading the source or hitting the error.

## The fix

Forward both variables in the `registry` service of all three compose files, next to `OPENBAO_ROLE`, with empty defaults so the `token` and `kubernetes` paths are unaffected:

```yaml
      - OPENBAO_ROLE=${OPENBAO_ROLE:-}
      - OPENBAO_ROLE_ID=${OPENBAO_ROLE_ID:-}
      - OPENBAO_SECRET_ID=${OPENBAO_SECRET_ID:-}
```

`docker-compose.yml` had the identical gap, not only the prebuilt and podman variants named in the issue, so all three are changed.

Document them in `.env.example` (with a note that `OPENBAO_SECRET_ID` is a secret) and in both parameter tables. While confirming the gap I found two related inaccuracies in the existing docs and fixed them in the same change, because they would send an operator down the wrong path even after the variables are wired:

| File | Before | After |
|---|---|---|
| `docs/egress-credential-vault.md` `OPENBAO_ROLE` row | "required for `kubernetes`/`approle` auth" | `kubernetes` only. The approle branch never reads `OPENBAO_ROLE`; only the kubernetes branch does (L113–116). |
| `docs/unified-parameter-reference.md` `OPENBAO_AUTH_METHOD` row | `token` \| `kubernetes` | `token` \| `kubernetes` \| `approle` |

The Helm `reserved-env-names.txt` is deliberately **not** extended: these two are not wired by the chart (Helm uses `kubernetes` auth), and keeping them unreserved is what lets a Helm operator inject them via `extraEnv` if they ever need to.

## Verification

New `tests/unit/test_compose_openbao_approle_env.py`, 19 cases:

- both credentials present in the `registry` service of every compose file, and interpolated as `${VAR:-}` (a hard-coded value would ignore `.env`; a required `:?` default would break the other two auth methods);
- a generalisation that **derives the expected `OPENBAO_*` set from `factory.py`** instead of a hand-maintained list, so a future auth method that reads a new variable fails the parity check rather than repeating this silently;
- `.env.example` lists both;
- both parameter tables list both and mention `approle`;
- the vault doc no longer presents `OPENBAO_ROLE` as an approle requirement.

Against the pre-fix tree, with the test kept and only the source changes stashed: **all 19 fail.** With the fix: 19 pass.

`pre-commit run --files <changed files>` passes with `SKIP=pytest-fast`, matching `.github/workflows/pre-commit.yaml`.

## Deliberately not changed

The approle branch reads `os.environ` directly while the kubernetes branch reads `settings.openbao_role`. Adding `openbao_role_id` / `openbao_secret_id` fields to `Settings` would make the three branches consistent and surface the variables in `/api/config` like the rest of the `OPENBAO_*` family. It is a code change with a config-surface side effect, though, and keeping a secret out of `Settings` is a reasonable position too, so I left it for you to decide rather than bundling it into a compose fix. Happy to follow up either way.

Also unchanged: `OPENBAO_TOKEN` stays mandatory via `${OPENBAO_TOKEN:?}` in every compose file, because the same value seeds the bundled dev-mode OpenBao container. An operator on approle against an external OpenBao must therefore still set a value the registry never uses. That is defensible for the dev instance; a one-line note in the docs would save a confused minute, but it is out of scope here.


omrishiv (approved):
@Evolver-sweden thanks for this, LGTM

